PR #2, opened 2018-02-08
Do not allow a device to be enslaved to a bridge and bond simultaneously

daxtens:
It gets very difficult if we use a single function to handle both the
bridge and bond case, so for the sake of simplicity I duplicated the
function.

LP: #1748057

I've also switched to strduping the interface name rather than using
references which just makes parsing a little more robust against things
failing part way through.


cyphermox:
Looks good, thanks!

PR #3, opened 2018-02-19
Support migrating static interfaces from /e/n/i

daxtens:
This allows migration of most uses of static interfaces:
 - IPv4
 - IPv6
 - setting DNS nameservers and search domains (also for DHCP)
 - setting custom MTUs
 - setting custom MAC addresses (also for DHCP)

It doesn't support static bonds or bridges because currently there's no support for that at all in ifupdown-migrate.

### DIFF
--- a/netplan/cli/commands/__init__.py
+++ b/netplan/cli/commands/__init__.py
@@ -16,5 +16,6 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from netplan.cli.commands.ip import NetplanIp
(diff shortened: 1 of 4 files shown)


daxtens:
Hmm, this passes locally but not on Semaphore because my tests assume a particular ordering. I will fix that up.

daxtens:
Will do.

daxtens:
I don't think GH can be told to try merging to a different branch, so I have opened a new PR: https://github.com/CanonicalLtd/netplan/pull/6

That will merge into the command-refactor branch, which then merges cleanly to master.

PR #7, opened 2018-02-28
migrate: line up accept_ra behaviour with systemd

daxtens:
Per https://www.freedesktop.org/software/systemd/man/systemd.network.html:
"systemd's setting of 1 (i.e. true) corresponds to kernel's setting of 2."

Fix the migration code to recognise 0 and 2 but not 1, rather than 0 and 1
but not 2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ clean:
 
 check: default linting
 	tests/generate.py
-	tests/cli.py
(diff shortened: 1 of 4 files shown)


daxtens:
The test failure is because the tests don't set ENABLE_TEST_COMMANDS atm. Patch incoming.

PR #8, opened 2018-02-28
LP: #1746842 - No network devices configured if YAML error in any file

daxtens:
- Start a VM with a network interface created by cloud-init (say ens3), and assume you have no physical console (e.g. cloud VM)
- Add another network interface (say ens7)
- Create another YAML file to configure ens7, but make a small error. This file need not reference ens3 at all!
- Reboot machine

- Observe the following message on boot
systemd[332]: /lib/systemd/systemd-generators/netplan failed with error code 1.

- Observe that you can no longer SSH in to the machine because the error in the new yaml file prevents all generation, so ens3 is not configured.

This renders the system entirely unusable unless there is an alternative way to get in.

This fixes that - with 2 caveats:

1) It only helps if the error occurs in the parse phase rather than the generate phase.

2) If you have settings for a device spread out over multiple files, you can potentially end up in an unexpected state. This is probably still better than having no configuration at all, which is arguably the ultimate unexpected state.

Overall, I think it's a massive usability win, and it brings us more in line with how ifupdown works.

[Old behaviour can be retained by passing `--strict` to generate.]

### DIFF
--- a/src/generate.c
+++ b/src/generate.c
@@ -35,8 +35,10 @@ static gchar* rootdir;
 static gchar** files;
 static gboolean any_networkd;
 static gchar* mapping_iface;
+static gboolean strict;
 
 static GOptionEntry options[] = {
+    {"strict", 's', 0, G_OPTION_ARG_NONE, &strict, "Exit on first error rather than ignoring broken files"},
     {"root-dir", 'r', 0, G_OPTION_ARG_FILENAME, &rootdir, "Search for and generate configuration files in this root directory instead of /"},
     {G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &files, "Read configuration from this/these file(s) instead of /etc/netplan/*.yaml", "[config file ..]"},
     {"mapping", 0, 0, G_OPTION_ARG_STRING, &mapping_iface, "Only show the device to backend mapping for the specified interface."},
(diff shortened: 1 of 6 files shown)


daxtens:
So the tests run with Xenial's glib, and I was building with Artful's glib, and the pointer array type I rely on was added at some point between them (https://abi-laboratory.pro/tracker/headers_diff/glib/2.53.1/current/diff.html). I guess I'll rework this to build on Xenial.

daxtens:
(or more specifically, g_ptr_array_find was added in 2.54)

daxtens:
... and now working with a simple implementation of my own. Seemed simpler than moving data structures (especially since GArray doesn't have a find function either).

cyphermox:
This is a great idea. Please fix the coverage to retain 100%

schopin-pro:
Hello @daxtens,

Thank you for your contribution, it is much appreciated.

However, after a discussion with the rest of the current Netplan team, we've reached the conclusion that we prefer the current behaviour of failing completely rather than appying a partial configuration. Our decision is based on several factors:

* The predictability of the system is greater: either it produces a full backend configuration, or none at all
* Partial application might actually break systems when applied directly via `netplan apply`. For instance, if the operator's SSH connection is routed through a bridge, and said bridge configuration is faulty, `netplan apply` will break the system, whereas an overall failure would leave the current configuration untouched.
* Operators are expected to test their configurations using `netplan try` and `netplan apply`, which should at least catch syntax issues before the reboot scenario.

Thanks for raising this issue and provide us with an alternative, though!

daxtens:
Wow what a blast from the past. That's a perfectly reasonably position to take, especially after all this time.

I'm glad to see that netplan is seeing some active maintainership again. I'll have to get my [netplan config tester](https://github.com/daxtens/netplan-config-tester) going again now :)

PR #9, opened 2018-03-08
Rework the Makefile to be more distribution-agnostic and easier to override

Conan-Kudo:
Rework the Makefile to handle supporting different file system layouts across a variety of distributions, especially merged-usr vs split-usr filesystems. Now all file system paths can be trivially overridden from the `make` invocation.

In addition, path detection via `pkg-config` is used for `bash-completion`, with a fallback to `/etc/bash_completion.d/` in the event the Makefile cannot probe for the correct path.

The symlinks made are now relative to `DESTDIR`, and `DESTDIR` is now consistently used everywhere.

Finally, the Python code linting program detection has been extended slightly to support the executable file names used in Fedora/Mageia/openSUSE in addition to the Debian/Ubuntu ones.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,22 @@ BUILDFLAGS = \
 	$(NULL)
 
 SYSTEMD_GENERATOR_DIR=$(shell pkg-config --variable=systemdsystemgeneratordir systemd)
+SYSTEMD_UNIT_DIR=$(shell pkg-config --variable=systemdsystemunitdir systemd)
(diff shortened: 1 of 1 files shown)


cyphermox:
Looks great. I've already cut netplan 0.34 (was about to push), so I'll merge this in 0.34.1 or something after testing it.

PR #10, opened 2018-03-14
docs: make dhcp6 more comprehensive

daxtens:
dhcp6 is useful for stateful and stateless DHCP.

dhcp6 is also useful to cause an interface to be brought up in a
fully autoconfigured IPv6-only environment, where otherwise the
interface wouldn't be brought up.

Point out that rdnssd is needed for rdnss with networkd.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -159,7 +159,19 @@ Common properties for all device types
 
 ``dhcp6`` (bool)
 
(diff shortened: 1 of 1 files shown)


daxtens:
This affects https://bugs.launchpad.net/netplan/+bug/1722862 and https://bugs.launchpad.net/netplan/+bug/1752211

PR #11, opened 2018-03-19
Introduce initial RPM spec for packaging

Conan-Kudo:
This spec file builds Netplan for Fedora and EL7 (with EPEL repositories enabled).

Signed-off-by: Neal Gompa <ngompa13@gmail.com>


cyphermox:
Thanks!

PR #12, opened 2018-03-20
parser: print helpful message if tabs are used

daxtens:
libyaml's error message is unhelpful:
Invalid YAML at //etc/netplan/10-bridge.yaml line 5 column 0: found
character that cannot start any token

Replace it with something more helpful.


cyphermox:
The idea is great, but let's try to figure out a different way to do this, without first going through the entire file once to check for tab characters...

cyphermox:
Closing; I'm about to merge fixes for the error-context branch that includes checks for tabs.

PR #13, opened 2018-03-20
Move replug blacklisted devices into a config file

daxtens:
Currently, devices that are blacklisted for replug are hardcoded
in apply.py. This is going to make support harder as people will
have to patch a python file to add another blacklist entry.

Move it to a yaml file - this means people can patch a less scary
/usr/share/netplan/blacklist.yaml to test blacklisting of a device
or to apply a workaround while waiting for the package to go through
the SRU process.

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -22,6 +22,8 @@ import os
 import sys
 import glob
 import subprocess
+import yaml
+import fnmatch
 
 import netplan.cli.utils as utils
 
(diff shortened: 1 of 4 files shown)


netplan/cli/commands/apply.py
```diff
@@ -124,30 +135,27 @@ def replug(self, device):  # pragma: nocover (covered in autopkgtest)
...
-            if 'brcmfmac' in driver_name:
-                logging.debug('replug %s: brcmfmac drivers do not support rebinding, ignoring', device)
+
+            for entry in self.blacklist:
+                # if a device matches all the criteria in a blacklist entry,
+                # then do not replug it
+                # note that an empty blacklist entry will match everything!
+
+                if 'driver' in entry:
+                    if entry['driver'] != driver_name:
```
cyphermox: This is wrong. Driver may be a partial name, such as for brcmfmac.

daxtens:
OK, so v2 has both globbing and an /etc/ file. I don't have the makefile to install the config file yet; this is so you can give feedback on the design and basic implementation.

daxtens:
(I'll resolve the merge conflict with v3 as well.)

daxtens:
Closed in favour of #21.

PR #15, opened 2018-03-27
try: add new 'netplan try' command

cyphermox:
Allows users to "test" new network configuration, requiring input to confirm changes.

### DIFF
--- a/netplan/cli/commands/try_command.py
+++ b/netplan/cli/commands/try_command.py
@@ -17,17 +17,15 @@
 
 '''netplan try command line'''
 
-import fcntl
 import os
-import termios
 import time
-import select
 import signal
 import sys
 
 from netplan.configmanager import ConfigManager
 import netplan.cli.utils as utils
 from netplan.cli.commands.apply import NetplanApply
+import netplan.terminal
 
 # Keep a timeout long enough to allow the network to converge, 60 seconds may
 # be slightly short given some complex configs, i.e. if STP must reconverge.
@@ -56,15 +54,8 @@ class NetplanTry(utils.NetplanCommand):
     def command_try(self):  # pragma: nocover (requires user input)
         try:
             fd = sys.stdin.fileno()
-            old_flags = fcntl.fcntl(fd, fcntl.F_GETFL)
-            old_term = termios.tcgetattr(fd)
-
-            raw_term = termios.tcgetattr(fd)
-            raw_term[3] = raw_term[3] & ~termios.ICANON
-            raw_term[3] = raw_term[3] & ~termios.ECHO
-            termios.tcsetattr(fd, termios.TCSANOW, raw_term)
-
-            fcntl.fcntl(fd, fcntl.F_SETFL, old_flags | os.O_NONBLOCK)
+            t = netplan.terminal.Terminal(fd)
+            t.enable_nonblocking_io()
 
             # we really don't want to be interrupted while doing backup/revert operations
             signal.signal(signal.SIGINT, self._signal_handler)
@@ -74,39 +65,25 @@ class NetplanTry(utils.NetplanCommand):
 
             NetplanApply.command_apply()
 
-            print("Do you want to keep these settings?\n\n")
-            print("Press ENTER before timeout to accept the new configuration\n\n")
-            timeout_now = INPUT_TIMEOUT
-            while (timeout_now > 0):
-                print("Network changes will revert in {:>2} seconds".format(timeout_now), end='\r')
-                i, o, err = select.select([sys.stdin], [], [], 1)
-                try:
-                    sys.stdin.read()
-                    raise ConfigurationAccepted()
-                except ConfigurationAccepted:
-                    raise
-                except ConfigurationRejected:
-                    raise
-                except TypeError:
-                    pass
-                timeout_now -= 1
-            raise ConfigurationRejected()
-        except ConfigurationRejected:
+            t.get_confirmation_input(timeout=INPUT_TIMEOUT)
+        except netplan.terminal.InputRejected:
             print("\nReverting.")
             self.revert()
-        except ConfigurationAccepted:
+        except netplan.terminal.InputAccepted:
             print("\nConfiguration accepted.")
         except Exception as e:
             print("\nAn error occured: %s" % e)
             print("\nReverting.")
             self.revert()
         finally:
+            t.reset()
             self.cleanup()
-            termios.tcsetattr(fd, termios.TCSAFLUSH, old_term)
-            fcntl.fcntl(fd, fcntl.F_SETFL, old_flags)
 
     def backup(self):  # pragma: nocover (requires user input)
-        self.config_manager.backup(with_config_file=self.config_file)
+        backup_config_dir = False
+        if self.config_file:
+            backup_config_dir = True
+        self.config_manager.backup(backup_config_dir=backup_config_dir)
 
     def setup(self):  # pragma: nocover (requires user input)
         if self.config_file:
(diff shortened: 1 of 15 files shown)


netplan/cli/commands/try_command.py
```diff
@@ -0,0 +1,137 @@
...
+            while (timeout_now > 0):
+                print("Network changes will revert in {:>2} seconds".format(timeout_now), end='\r')
+                i, o, err = select.select([sys.stdin], [], [], 1)
+                try:
+                    sys.stdin.read()
+                    raise ConfigurationAccepted()
+                except ConfigurationAccepted:
+                    raise
+                except ConfigurationRejected:
+                    raise
```
mwhudson: I don't think these lines achieve anything

netplan/cli/commands/try_command.py
```diff
@@ -0,0 +1,137 @@
...
+                except TypeError:
+                    pass
```
mwhudson: TypeError for me usually means a bug. What is this line for?

netplan/cli/commands/try_command.py
```diff
@@ -0,0 +1,137 @@
...
+            NetplanApply.command_apply()
+
+            print("Do you want to keep these settings?\n\n")
+            print("Press ENTER before timeout to accept the new configuration\n\n")
+            timeout_now = INPUT_TIMEOUT
+            while (timeout_now > 0):
+                print("Network changes will revert in {:>2} seconds".format(timeout_now), end='\r')
+                i, o, err = select.select([sys.stdin], [], [], 1)
+                try:
+                    sys.stdin.read()
```
mwhudson: I'm confused by this code. Shouldn't you inspect i before calling this? Shouldn't you check that the user has indeed pressed enter and not some other key?

netplan/cli/commands/try_command.py
```diff
@@ -0,0 +1,137 @@
...
+                                 help='Apply the config file in argument on top of current configuration.')
+
+        self.func = self.command_try
+
+        self.parse_args()
+        self.run_command()
+
+    def command_try(self):  # pragma: nocover (requires user input)
+        try:
+            fd = sys.stdin.fileno()
```
mwhudson: could you put the terminal mangling in another method to make the overall structure here clearer to read?

netplan/configmanager.py
```diff
@@ -0,0 +1,90 @@
...
+            self._copy_tree(os.path.join(self.prefix, "etc/netplan"),
+                            os.path.join(self.temp_etc, "netplan"))
+        self._copy_tree(os.path.join(self.prefix, "run/NetworkManager/system-connections"),
+                        os.path.join(self.temp_run, "NetworkManager", "system-connections"),
+                        missing_ok=True)
+        self._copy_tree(os.path.join(self.prefix, "run/systemd/network"),
+                        os.path.join(self.temp_run, "systemd", "network"),
+                        missing_ok=True)
+
+    def revert(self, with_config_file=False):
```
mwhudson: This method doesn't use the with_config_file argument?

netplan/terminal.py
```diff
@@ -0,0 +1,104 @@
...
+            print("Changes will revert in {:>{}} seconds".format(timeout_now, len(str(timeout))), end='\r')
+
+            # wait for usable input from stdin; but we don't /really/ care what it is just yet
+            select.select([sys.stdin], [], [], 1)
+            try:
+                # retrieve the input from the terminal
+                c = sys.stdin.read()
+                if (c == '\n'):
+                    raise InputAccepted()
+            except TypeError:
```
vorlonofportland: why would you have a TypeError here?

netplan/terminal.py
```diff
@@ -0,0 +1,104 @@
...
+                pass
+            timeout_now -= 1
+        raise InputRejected()
```
vorlonofportland: why use exceptions rather than returned values?

tests/test_terminal.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertTrue(attrs[3] & termios.ECHO)
+        self.assertEquals(orig_attrs, attrs)
+
+    def test_nonblocking_io(self):
+        orig_flags = flags = fcntl.fcntl(self.terminal.fd, fcntl.F_GETFL)
+        self.assertFalse(flags & os.O_NONBLOCK)
+        self.terminal.enable_nonblocking_io()
+        flags = fcntl.fcntl(self.terminal.fd, fcntl.F_GETFL)
+        self.assertTrue(flags & os.O_NONBLOCK)
+        self.assertNotEquals(flags, orig_flags)
```
vorlonofportland: It seems redundant to me to have both of the above asserts, as opposed to just the second.

netplan/configmanager.py
```diff
@@ -0,0 +1,90 @@
...
+            temp_nm_path = "{}/NetworkManager/system-connections".format(self.temp_run)
+            temp_networkd_path = "{}/systemd/network".format(self.temp_run)
+            if os.path.exists(temp_nm_path):
+                shutil.rmtree(os.path.join(self.prefix, "run/NetworkManager/system-connections"))
+                self._copy_tree(temp_nm_path,
+                                os.path.join(self.prefix, "run/NetworkManager/system-connections"))
+            if os.path.exists(temp_networkd_path):
+                shutil.rmtree(os.path.join(self.prefix, "run/systemd/network"))
+                self._copy_tree(temp_networkd_path,
+                                os.path.join(self.prefix, "run/systemd/network"))
```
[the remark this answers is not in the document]
cyphermox: The virtual interface may be left behind, but the previous config will be re-applied -- there's no harm in leaving the interface behind, aside from it being unnecessary.

As for potentially un-revertable changes (bridge params), I agree we should exit non-zero instead of trying at all.

I'll write something to skip testing at all in the case of those parameters, but there are still corner cases that can't be covered until we can also diff config (which will be a separate feature on its own)

netplan/terminal.py
cyphermox: read() on stdin in this case is non-blocking; if after its timeout of 1 sec there was no input, it will TypeError.

tests/test_terminal.py
cyphermox: I did notice cases were flags appeared to get changed along the way. It's also making sure that it's not just we changed O_NONBLOCK, we also really did revert things to exactly what they were before, there are no side-effects to the code.

netplan/cli/commands/try_command.py
```diff
@@ -0,0 +1,110 @@
...
+            t.disable_echo()
+            t.enable_nonblocking_io()
+
+            # we really don't want to be interrupted while doing backup/revert operations
+            signal.signal(signal.SIGINT, self._signal_handler)
+
+            self.backup()
+            self.setup()
+
+            NetplanApply.command_apply()
```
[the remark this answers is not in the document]
cyphermox: It's not. Right now it's very much asynchronous (by that  I mean, the daemons are queue to be restarted, but we don't wait for them to signal that they have completed). That's certainly something that can be changed, but in the grand scheme of things, where you can't guarantee that the network will even converge during the timeout, it doesn't seem like that big a deal. Even if we wait for the daemon to say it has restarted, they will only have started to apply the changes, which might still easily take up to a minute to complete, depending on the configuration.

I'll change this to make this particular call synchronous, but we'll still be left with the convergence time.

netplan/terminal.py
cyphermox: I need to catch for exceptions anyway in the code to catch for legitimate errors, and cleanup is required at the end. This is an elegant way to achieve all of it in a consistent way (see try_command.py where get_confirmation_input() is used). This has the benefit of allowing to signal a failure on Control-C as well.

cyphermox:
OK; review has been done and corrected, merging.

PR #16, opened 2018-03-27
yaml: add dhcp-identifier: field for compatibility with non-RFC4361 things

cyphermox:
Make sure our users who use DHCP server software that doesn't understand
RFC-4361 unique client identifiers can still get the right reservation
behavior. When 'dhcp-identifier: mac' is set, set ClientIdentifier: mac
for systemd-networkd so it does the right thing.

'dhcp-identifier: duid' is valid too, but is the default and thus a noop.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -159,6 +159,12 @@ Virtual devices
     Note that **``rdnssd``**(8) is required to use RDNSS with networkd. No extra
     software is required for NetworkManager.
 
+``dhcp-identifier`` (scalar)
+
+:   When set to 'mac'; pass that setting over to systemd-networkd to use the
+    device's MAC address as a unique identifier rather than a RFC4361-compliant
+    Client ID. This has no effect when NetworkManager is used as a renderer.
+
 ``accept-ra`` (bool)
 
 :   Accept Router Advertisement that would have the kernel configure IPv6 by itself.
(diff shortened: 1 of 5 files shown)


vorlonofportland:
Just to be sure, is ClientIdentifier=mac a setting that is supported to set for networkd globally, rather than per-link?  Is there a reason that we believe ClientIdentifier=duid is the correct default behavior for networkd, despite differing from prior practice?

Is the need for configurability a consequence of bugs in dhcp server implementations, or is this a workaround for the fact that a single Ubuntu system may have multiple dhcp clients talking to the same server and using different client ids?  (e.g. isc-dhcp-client in initramfs, vs. networkd in booted system)

cyphermox:
ClientIdentifier= in systemd-networkd is a per-link attribute, not configurable elsewhere (ie. if we wanted to change the default, we'd need to patch systemd-networkd). While it's different from prior behavior, it is consistent with what is proper behavior for systems: given a stable identifier across network adapters and across changes in hardware, you'd be able to get the same reservation (ie. my IP might be the same if I connect over wireless or over wired).

The net effect AFAICT is that this is properly handled in most environments, with the possible exceptions of some Windows environments where the DHCP server is Windows Server's implementation, with reservations configured by some other tool (c.f. https://askubuntu.com/questions/987673/how-to-get-netplan-on-17-10-server-to-work-with-a-windows-server-dhcp-server)

I do think the behavior change introduced by systemd-networkd is correct, but this is precisely the kind of improvement that can't happen without breaking _some_ setups.

I'm open to tabling this for now until more testing can be done with a Windows Server for DHCP server.

PR #18, opened 2018-03-28
yaml: extract error context from parser

cyphermox:
When hitting low-level errors; extract all the information we can from the
parser object from libyaml to provide meaningful error messages to the user,
including the exact position of the error.

Signed-off-by: Mathieu Trudel-Lapierre <mathieu.trudel-lapierre@canonical.com>

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -76,14 +76,14 @@ get_error_context(yaml_parser_t parser, GError **error)
     while (current > parser.buffer.start) {
         current--;
         if (*current == '\n') {
-            line = current + 2;
+            line = current + 1;
             break;
         }
     }
     if (current <= parser.buffer.start)
         line = parser.buffer.start;
     current = line + 1;
-    while (current <= parser.buffer.end) {
+    while (current <= parser.buffer.last) {
         if (*current == '\n') {
             *current = '\0';
             break;


cyphermox:
@daxtens that one's for you. What do you think?

The pointer magic in get_error_context() is horrible, there's certainly a better way to do this, but given we're at a failure case anyway we shouldn't be afraid to walk all around libyaml's memory. I only lightly tested this, but it seems to work for the most part.

daxtens:
I will do more comprehensive code review in a sec, but on a more fundamental level, what is the issue with going through the file twice?

My first thought was performance, but there is so much caching thoughout the stack that we should only hit the disk once: the file should still be in the page cache. My second thought was elegance - it's a bit inelegant to re-read the file, but it's simpler and less scary than introspecting private libyaml state. Or is there another reason I haven't thought of?

src/parse.c
```diff
@@ -63,6 +63,45 @@ int missing_ids_found;
...
+        if (*current == '\n') {
+            line = current + 2;
+            break;
+        }
+    }
+    if (current <= parser.buffer.start)
+        line = parser.buffer.start;
+    current = line + 1;
+    while (current <= parser.buffer.end) {
+        if (*current == '\n') {
```
daxtens: This line is causing an ASAN splat in fuzz testing:

```
==16737==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62f00000c400 at pc 0x55f193ce7689 bp 0x7ffd2cf54c30 sp 0x7ffd2cf54c20                                                        
READ of size 1 at 0x62f00000c400 thread T0                                                                                                                                                       
    #0 0x55f193ce7688 in get_error_context src/parse.c:87                                                                                                                                        
    #1 0x55f193ce7a18 in load_yaml src/parse.c:128                                                                                                                                               
    #2 0x55f193cefa88 in parse_yaml src/parse.c:1578                                                                                                                                             
    #3 0x55f193ce62b2 in process_input_file src/generate.c:144                                                                                                                                   
    #4 0x55f193ce6fac in main src/generate.c:237                                                                                                                                                 
    #5 0x7fc59bbb2b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)                                                                                                             
    #6 0x55f193ce55e9 in _start (/home/ubuntu/netplan/generate+0xd5e9)                                                                                                                           

0x62f00000c400 is located 0 bytes to the right of 49152-byte region [0x62f000000400,0x62f00000c400)
allocated by thread T0 here:
    #0 0x7fc59cfacb50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x7fc59c3acfa9 in yaml_parser_initialize (/usr/lib/x86_64-linux-gnu/libyaml-0.so.2+0x4fa9)
    #2 0x55f193cefa88 in parse_yaml src/parse.c:1578
    #3 0x55f193ce62b2 in process_input_file src/generate.c:144
    #4 0x55f193ce6fac in main src/generate.c:237
    #5 0x7fc59bbb2b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

SUMMARY: AddressSanitizer: heap-buffer-overflow src/parse.c:87 in get_error_context
Shadow bytes around the buggy address:
  0x0c5e7fff9830: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5e7fff9840: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5e7fff9850: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5e7fff9860: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x0c5e7fff9870: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x0c5e7fff9880:[fa]fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5e7fff9890: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5e7fff98a0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5e7fff98b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5e7fff98c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c5e7fff98d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==16737==ABORTING
```

This can be reproduced by compiling with `make CFLAGS+="-g -fsanitize=address"`, and testing with a specifically chosen input file. It seems to happen when there is an invalid yaml file with no trailing newline.
A sample input file which causes issues is:
```
}k
```
(with no trailing newline)

If a trailing newline is added, the ASAN splat goes away.

daxtens: I think the issue may be that `buffer.end` might point to just beyond the end of the buffer?

src/parse.c
```diff
@@ -63,6 +63,45 @@ int missing_ids_found;
...
+    }
+
+    g_string_append_printf(message, "%s\n", line);
+
+    for (i = 0;
+         (parser.problem_mark.column > 0 && i < (parser.problem_mark.column - 1));
+         i++)
+        g_string_append_printf(message, " ");
+    g_string_append_printf(message, "^");
+
```
daxtens: Irritatingly, this isn't printing the invalid tab character for me.

Input:
```
networks:
        hi: {}
```
(where the second line is indented with a tab)

Output:
```
Invalid YAML at testroot//etc/netplan/a.yaml line 1 column 0: tabs are not allowed for indent:
hi: {}
^
```
The caret is pointing at the wrong thing - it's somehow stripped/missed the actual tab character. This makes it look like 'hi' is the problem, rather than the tab.

tests/generate.py
```diff
@@ -4153,6 +4153,24 @@ def test_invalid_dhcp_identifier(self):
...
+    *engreen:
+      dhcp4: yes''')
+
+    def test_invalid_yaml_undefined_alias_at_eof(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      dhcp4: *yes
+''')
```
daxtens: As semaphoreci points out, these all need expect_fail=True.

src/parse.c
jsoref: You want to use `buffer.last` instead, see: https://github.com/snoyberg/yaml/blob/50a95e00e49fcfad31c460bf0776575667382571/libyaml/yaml.h#L1147-L1148

src/parse.c
```diff
@@ -63,6 +63,45 @@ int missing_ids_found;
...
+    unsigned char* line = parser.buffer.pointer;
+    unsigned char* current = line;
+    int i;
+
+    message = g_string_sized_new(200);
+
+    while (current > parser.buffer.start) {
+        current--;
+        if (*current == '\n') {
+            line = current + 2;
```
jsoref: Afaict, this line is buggy, it should just be `current + 1`.

I've tried tracing this through with sample input:
```
    end:                         v
   last:           v
  start: v
 buffer: ABC␤↹EF␤GH∅
current:     ^ (#70)
   line:     ^ (#71)
current:    ^ (#77)
   line:      ^ (#79) -- this appears buggy
current:      ^ (#85)
current:       ^ (#91)
current:        ^ (#91)
 buffer: ABC␤↹EF∅GH∅ (#88)
   line:      ^ (unchanged from above)
current:        ^ (unchanged from above)
message:…EF∅ (#94)
```

src/parse.c
```diff
@@ -63,6 +63,45 @@ int missing_ids_found;
  * Loading and error handling
  ****************************************************/
 
+static char *
+get_error_context(yaml_parser_t parser, GError **error)
+{
```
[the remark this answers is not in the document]
jsoref: yaml_parser_t is a nontrivial object: https://github.com/snoyberg/yaml/blob/50a95e00e49fcfad31c460bf0776575667382571/libyaml/yaml.h#L1083-L1305 you definitely want to pass a pointer to it.

cyphermox:
Replacing with a new pull request to clean up / unbreak automerge.

PR #19, opened 2018-04-11
accept-ra: do not enable by default, default to unset

raharper:
Do not always emit an IPv6AcceptRA value into network configurations.
Systemd-networkd defaults to kernel value as long as the value stays
unset in the configuration.  When IPv6AcceptRA is enabled in systemd
this makes networkd wait 10 seconds for potential RAs to arrive.  This
delays boot by 10 seconds always whether or not the network has an
IPV6 Router running.  Leaving things unsets keeps existing behavior
of Ubuntu systems where RAs are accepted and processed at the time
they are received and boot is not delayed by waiting for an RA to arrive.

If users want to forcibly enable or disable, they may do so by including
a value for accept-ra in their yaml configuraion.

LP: #1732002

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -168,7 +168,8 @@ Virtual devices
 ``accept-ra`` (bool)
 
 :   Accept Router Advertisement that would have the kernel configure IPv6 by itself.
-    On by default.
+    When enabled, accept Router Advertisements. When disabled, do not respond to
+    Router Advertisements.  If unset use the host kernel default setting.
 
 ``addresses`` (sequence of scalars)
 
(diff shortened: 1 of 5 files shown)


cyphermox:
The bond and bridge accept-RA handling is unnecessary, the earlier cases already write it.

I've started to merge this already, applying the fixes.

raharper:
> The bond and bridge accept-RA handling is unnecessary, the earlier cases already write it.

Yes, I should have just dropped it from emitting the IPv6AcceptRA=no by default.

thanks

cyphermox:
Closing; I put together https://github.com/CanonicalLtd/netplan/pull/20 with the changes; what do you think?

raharper:
Thanks, that looks great.

PR #21, opened 2018-04-13
Support blacklisting devices for replug in config files

daxtens:
Currently, devices that are blacklisted for replug are hardcoded
in `apply.py`. This is going to make support harder as people will
have to patch a python file to add another blacklist entry.

Introduce `replug:` as a mapping at the top level that contains:

 - `disable_all_replug`: a kill switch for replugging

 - `blacklist`: a sequence of mappings that describes devices to be
              blacklisted

The contents of the current list is moved to `blacklist.yaml` which
is installed in `/lib/netplan/00-netplan-replug-blacklist.yaml`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ clean:
 check: default linting
 	tests/generate.py
 	tests/cli.py
+	python3-coverage run --append --omit=tests/\* -- tests/internals.py
 
 linting:
 	$(PYFLAKES3) $(PYCODE)
@@ -70,6 +71,7 @@ install: default
 	install -m 755 generate $(DESTDIR)/$(ROOTLIBEXECDIR)/netplan/
 	find netplan/ -name '*.py' -exec install -Dm 644 "{}" "$(DESTDIR)/$(DATADIR)/netplan/{}" \;
 	install -m 755 src/netplan.script $(DESTDIR)/$(DATADIR)/netplan/
+	install -m 644 blacklist.yaml $(DESTDIR)/$(ROOTLIBEXECDIR)/netplan/00-netplan-replug-blacklist.yaml
 	ln -sr $(DESTDIR)/$(DATADIR)/netplan/netplan.script $(DESTDIR)/$(SBINDIR)/netplan
 	ln -sr $(DESTDIR)/$(ROOTLIBEXECDIR)/netplan/generate $(DESTDIR)/$(SYSTEMD_GENERATOR_DIR)/netplan
 	install -m 644 doc/*.html $(DESTDIR)/$(DOCDIR)/netplan/
--- a/netplan/cli/utils.py
+++ b/netplan/cli/utils.py
@@ -19,6 +19,9 @@ import sys
 import os
 import argparse
 import subprocess
+import glob
+import yaml
+import logging
 
 NM_SERVICE_NAME = 'NetworkManager.service'
 NM_SNAP_SERVICE_NAME = 'snap.network-manager.networkmanager.service'
@@ -60,6 +63,40 @@ def systemctl_network_manager(action):  # pragma: nocover (covered in autopkgtes
         service_name = NM_SNAP_SERVICE_NAME
 
     subprocess.check_call(['systemctl', action, '--no-block', service_name])
+
+
+def gather_replug_yaml(root_dir):
+    # create the file list
+    # per generate.c, /run/netplan shadows /etc/netplan/, which shadows /lib/netplan
+
+    names_to_paths = {}
+    for yaml_dir in ['lib', 'etc', 'run']:
+        for yaml_file in glob.glob(os.path.join(root_dir, yaml_dir, 'netplan', '*.yaml')):
+            names_to_paths[os.path.basename(yaml_file)] = yaml_file
+
+    files = [names_to_paths[name] for name in sorted(names_to_paths.keys())]
+
+    # now build the result from each file
+    result = {'disable_all_replug': False, 'blacklist': []}
+    for yaml_file in files:
+        try:
+            with open(yaml_file) as f:
+                yaml_data = yaml.load(f)
+        except (IOError, yaml.YAMLError):
+            logging.error('Error while loading %s, aborting.' % yaml_file)
+            sys.exit(1)
+
+        if 'replug' not in yaml_data:
+            continue
+
+        yaml_data = yaml_data['replug']
+        if 'disable_all_replug' in yaml_data:
+            result['disable_all_replug'] = yaml_data['disable_all_replug']
+
+        if 'blacklist' in yaml_data:
+            result['blacklist'] += yaml_data['blacklist']
+
+    return result
 
 
 class NetplanCommand(argparse.Namespace):
--- a/src/parse.c
+++ b/src/parse.c
@@ -1463,8 +1463,15 @@ const mapping_entry_handler network_handlers[] = {
  * Grammar and handlers for root node
  ****************************************************/
 
+/* we don't stress about validating this as it only affects apply */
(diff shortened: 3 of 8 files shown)


netplan/cli/utils.py
```diff
@@ -62,6 +65,40 @@ def systemctl_network_manager(action):  # pragma: nocover (covered in autopkgtes
...
+            names_to_paths[os.path.basename(yaml_file)] = yaml_file
+
+    files = [names_to_paths[name] for name in sorted(names_to_paths.keys())]
+
+    # now build the result from each file
+    result = {'disable_all_replug': False, 'blacklist': []}
+    for yaml_file in files:
+        try:
+            with open(yaml_file) as f:
+                yaml_data = yaml.load(f)
```
cyphermox: yaml.load() alone is potentially unsafe. You'd want safe_load().

netplan/cli/utils.py
```diff
@@ -62,6 +65,40 @@ def systemctl_network_manager(action):  # pragma: nocover (covered in autopkgtes
...
+        except (IOError, yaml.YAMLError):
+            logging.error('Error while loading %s, aborting.' % yaml_file)
+            sys.exit(1)
+
+        if 'replug' not in yaml_data:
+            continue
+
+        yaml_data = yaml_data['replug']
+        if 'disable_all_replug' in yaml_data:
+            result['disable_all_replug'] = yaml_data['disable_all_replug']
```
cyphermox: I'm not convinced yet that disable_all_replug is a good idea even if it might be convenient to users. That was also the point of having the whole replug story elsewhere.

If we're going to change something, I feel like we should instead have a single keyword that allows skipping replug on a per-interface basis.

The essential idea I have is that you'd want to not recommend users disable replug -- in general, it's a good idea to have the replugging happen, and not doing it can break renaming interfaces; let's not make it hard for people to disable it, but let's also make sure it's absolutely a conscious decision.

src/parse.c
```diff
@@ -1463,8 +1463,15 @@ const mapping_entry_handler network_handlers[] = {
...
 
+/* we don't stress about validating this as it only affects apply */
+static gboolean noop_handler(yaml_document_t* doc, yaml_node_t* node, const void* _, GError** error)
+{
+    return TRUE;
+}
+
 const mapping_entry_handler root_handlers[] = {
     {"network", YAML_MAPPING_NODE, NULL, network_handlers},
+    {"replug", YAML_MAPPING_NODE, noop_handler},
```
cyphermox: As above; let's make sure all the parsing happens in a single place, preferrably in C -- you want this to be as fast as possible and with as few dependencies as possible, since it happens at early boot.

Makefile
```diff
@@ -36,6 +36,7 @@ clean:
 check: default linting
 	tests/generate.py
 	tests/cli.py
+	python3-coverage run --append --omit=tests/\* -- tests/internals.py
```
cyphermox: Please make tests/internals.py  --> tests/test_replug.py or something similar that is more descriptive of the use of that file. The test_ prefix is important, let's use 'nosetests3 -v --with-coverage' if anything, so that additional tests that do not require running the binaries directly (e.g. real unit tests of code functions) can be detected automatically, that way we won't have to make more changes to Makefile when new unittests are added (see the try-command branch).

netplan/cli/utils.py
```diff
@@ -62,6 +65,40 @@ def systemctl_network_manager(action):  # pragma: nocover (covered in autopkgtes
     subprocess.check_call(['systemctl', action, '--no-block', service_name])
 
 
+def gather_replug_yaml(root_dir):
```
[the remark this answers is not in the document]
cyphermox: @vorlonofportland  Could we have your take as architect on this particular schema change, and the PR in general?

src/parse.c
daxtens: How would doing this in C work? The C layer doesn't do replugging, and none of the files it spits out control that process. AFAICT it's only done when you call `netplan apply`, which is pure python and also - unless I'm missing something - not run on boot.

netplan/cli/utils.py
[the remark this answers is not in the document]
daxtens: From a support point of view, I really want a kill switch. It will mean that I can ask users/customers who are experiencing issues to disable replug everywhere, and if that helps, to then narrow down a more specific rule. It means they have a work-around while we work on a fix.
From my point of view, trying to get a per-interface matching rule right - especially in the presence of renaming - is likely to require a few round-trips. If we don't have a workaround, it could well lead to lots of frustration. That's just where I'm coming from.

I'm happy for it to be something ominously named, like `disable_all_replug: yes_I_know_this_may_break_things`, if that helps.

netplan/cli/utils.py
daxtens: My concern about making it per-interface in this way how that would interact with the matching rules.

I think it's a bit chicken-and-egg. Looking at `apply.py`:
 - the backends get to try to bring up interfaces (apply.py, line 60)
 - then all interfaces which are down are replugged
 - if anything was replugged, udev is settled and the backends try again.

How does step 2 match an interface? Clearly it wasn't matched in step 1, otherwise the interface would be up - so it isn't matching on MAC address. Matching by current name won't work as the point of replugging is that you cannot rely on the original name, and if you're matching by the new name then that won't take effect until after replug, by which time the device will have failed. So I think this would only work if you're matching by driver - I can't think of anything else that would serve this purpose.

This means that if we have a user who has a device that is going away on replug, we first have to tell them to match by driver, and then add a stanza to stop that being replugged. (Also, if they have multiple devices like this - e.g. multiple Xen vifs or multiple qeths, which is very possible - they have to add the driver match and the replug prohibition to each one.)

This will also make it impossible to support the existing hardcoded blacklist in the same way as we support user-configured blacklists, which I think is a cute feature.

netplan/cli/utils.py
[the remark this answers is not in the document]
cyphermox: I don't think you understand my point; having:

network:
    ethernets:
        blah:
            replug: no

Would appropriately allow you to avoid replugging one interface -- there's no need to blanked disable everything.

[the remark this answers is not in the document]
daxtens: I'm still not clear on how that would work.

Say I have the following:
```
network:
  version: 2
  ethernets:
   ens8:
     match:
       name: ens8
     addresses: [10.10.10.2/24]
```

Now this will work without a replug if I don't rename ens8. But, if I do the following:
```
ubuntu@netplan:~$ sudo ip a del dev ens8 10.10.10.2/24
ubuntu@netplan:~$ sudo ip link set dev ens8 down
ubuntu@netplan:~$ sudo ip link set dev ens8 name renamedif0
```

The device now has a different name. Currently, it won't be matched based on the name in the first pass, so it will be replugged and then matched - per the following from `netplan --debug apply`:

```
DEBUG:replug renamedif0: unbinding virtio4 from /sys/bus/virtio/drivers/virtio_net
DEBUG:replug renamedif0: rebinding virtio4 to /sys/bus/virtio/drivers/virtio_net
```
(`ip -a` shows the match has worked - the interface has the IP address.)

So if I put `replug: no` in to the device stanza, it won't work: the match rule won't kick in until after the device has been replugged.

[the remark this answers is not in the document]
cyphermox: That's expected behavior; the name of the interface doesn't match the match: clause. If you want renaming to work correctly, you need to positively, uniquely identify a device; the same applies to disabling replug.

cyphermox:
We're quickly moving towards dropping replug instead, so we probably won't need this.

Leaving open for now, just in case.

daxtens:
There is no more replug so this can be closed.

PR #22, opened 2018-04-20
bonds: default to milliseconds for unspecified values in intervals

raharper:
Linux kernel bond parameter settings are exclusively in millisecond
units, save lp_interval which is in seconds.  Existing configurations and
documentation indicate values for mii-monitor-interval, arp-interval,
up-delay and down-delay are expressed in milliseconds.  Netplan will render
these values as milliseconds in the networkd backend unless a user has
indicated a different time unit by appending a suffix, like 's' for seconds.

LP: #1765833


cyphermox:
Looks great. Can you please make sure this passes coverage?

PR #25, opened 2018-05-04
ip leases: Add error handling when interface is not configured in net…

slashdd:
…plan or not set to DHCP.

For the command "netplan ip leases <interface>" to work, each interface needs to
be directly specified in a file under /etc/netplan and set to DHCP.

If the interface doesn't meet the above requirement, the command will
fail and generate a Traceback as follow :

    netplan.main()
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 130, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/ip.py", line 56, in run
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 130, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/ip.py", line 75, in run
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 130, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/ip.py", line 142, in command_ip_leases
    out = subprocess.check_output(argv, universal_newlines=True)
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/lib/netplan/generate', '--mapping', 'eth1']' returned non-zero exit status 1.

### DIFF
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -21,6 +21,7 @@ import logging
 import os
 import sys
 import subprocess
+from subprocess import CalledProcessError
 
 import netplan.cli.utils as utils
 
@@ -139,7 +140,10 @@ class NetplanIpLeases(utils.NetplanCommand):
 
         # Extract out of the generator our mapping in a dict.
         logging.debug('command ip leases: running %s', argv)
-        out = subprocess.check_output(argv, universal_newlines=True)
+        try:
+            out = subprocess.check_output(argv, universal_newlines=True)
+        except CalledProcessError as e:  # pragma: nocover (better be covered in autopkgtest)
+            sys.exit(1)
         mapping = {}
         mapping_s = out.split(',')
         for keyvalue in mapping_s:


slashdd:
LP bug: https://bugs.launchpad.net/bugs/1768823

slashdd:
Example with the proposal patch:

```
$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
45: eth0@if46: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 00:16:3e:91:0d:24 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.104.180.139/24 brd 10.104.180.255 scope global dynamic eth0
       valid_lft 3422sec preferred_lft 3422sec
    inet6 fd42:ad0:706e:4129:216:3eff:fe91:d24/64 scope global dynamic mngtmpaddr noprefixroute
       valid_lft 3497sec preferred_lft 3497sec
    inet6 fe80::216:3eff:fe91:d24/64 scope link
       valid_lft forever preferred_lft forever
```

```
$ netplan ip leases eth1
Interface eth1 is not managed by netplan or not configured for DHCP
```

```
$ netplan ip leases eth2
Interface eth2 is not managed by netplan or not configured for DHCP
```

```
$ netplan ip leases eth0
# This is private data. Do not parse.
ADDRESS=10.104.180.139
NETMASK=255.255.255.0
ROUTER=10.104.180.1
SERVER_ADDRESS=10.104.180.1
NEXT_SERVER=10.104.180.1
BROADCAST=10.104.180.255
T1=1800
T2=3150
LIFETIME=3600
DNS=10.104.180.1
DOMAINNAME=lxd
HOSTNAME=nplan8bionic
CLIENTID=ffb1e17c0e00020000ab11ab1476591f0e77eb
```

netplan/cli/commands/ip.py
```diff
@@ -139,7 +140,10 @@ def lease_method_nm_connection():  # pragma: nocover (covered in autopkgtest)
 
         # Extract out of the generator our mapping in a dict.
         logging.debug('command ip leases: running %s', argv)
-        out = subprocess.check_output(argv, universal_newlines=True)
+        try:
+            out = subprocess.check_output(argv, universal_newlines=True)
+        except CalledProcessError as e:  # pragma: nocover (better be covered in autopkgtest)
+            sys.exit("Interface %s is not managed by netplan or not configured for DHCP" % (self.interface))
```
vorlonofportland: Why print this message instead of returning an empty string?  The latter is far preferable for parsability; and certainly, we expect 'netplan ip leases' to be machine-consumable.

(It's ironic that currently we pass through the 'This is private data, do not parse' comment from netif/leases; we should really be filtering that, because regardless of whether systemd wants these treated as internal and not to be parsed, we are deliberately exposing them and need to deliver an interface guarantee here.)

PR #26, opened 2018-05-07
docs: document mtu option

daxtens:
Also explain that setting MTU while matching by name is unlikely to
work reliably and people should match by MAC!


cyphermox:
Closing, superseded by #27

PR #29, opened 2018-05-09
Implement support for advanced authentication options

RavuAlHemio:
Greetings,

we are currently preparing to deploy 802.1X authentication on our wired network. Since Ubuntu Bionic uses netplan for network configuration, support for additional authentication options is interesting for us.

The changes allow authentication options to be set using an `auth` block on an `ethernet` interface, `wifi` interface, or `wifi` `access-point`:

    network:
      version: 2
      ethernets:
        eth0:
          dhcp4: yes
          auth:
            key-mgmt: 8021x
            eap-method: ttls
            anonymous-identity: "@internal.example.com"
            identity: "joe@internal.example.com"
            password: "v3ryS3kr1t"

An `auth` block on a `wifi` interface is taken as a default for any descendant `access-points` that don't have their own `auth` block.

At this stage, this pull request is a proof of concept and request for comments. (Does the structure make sense? Should some of the option names be changed?) The set of supported authentication options is not yet as complete as that of NetworkManager (feature parity should be a realistic goal) and the functionality might have to be expanded to additional interface types.

Thanks in advance!

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -494,9 +494,9 @@ interfaces, as well as individual wifi networks, by means of the ``auth`` block.
      The following properties can be used if ``key-management`` is ``wpa-eap``
      or ``802.1x``:
 
-     ``eap-method`` (scalar)
-     :    The supported EAP methods are ``tls`` (TLS), ``peap`` (Protected EAP),
-          and ``ttls`` (Tunneled TLS).
+     ``method`` (scalar)
+     :    The EAP method to use. The supported EAP methods are ``tls`` (TLS),
+          ``peap`` (Protected EAP), and ``ttls`` (Tunneled TLS).
 
      ``identity`` (scalar)
      :    The identity to use for EAP.
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -482,9 +482,9 @@ interfaces, as well as individual wifi networks, by means of the ``auth`` block.
 
      ``key-management`` (scalar)
      :    The supported key management modes are ``none`` (no key management);
-          ``wpa-psk`` (WPA with pre-shared key, common for home wifi);
-          ``wpa-eap`` (WPA with EAP, common for enterprise wifi); and ``802.1x``
-          (used primarily for wired Ethernet connections).
+          ``psk`` (WPA with pre-shared key, common for home wifi); ``eap`` (WPA
+          with EAP, common for enterprise wifi); and ``802.1x`` (used primarily
+          for wired Ethernet connections).
 
      The following properties can be used if ``key-management`` is ``wpa-psk``:
 
@@ -550,7 +550,7 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           is equivalent to
 
               auth:
-                key-management: wpa-psk
+                key-management: psk
                 psk: "S3kr1t"
 
      ``mode`` (scalar)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -551,7 +551,7 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
 
               auth:
                 key-management: psk
-                psk: "S3kr1t"
+                password: "S3kr1t"
 
      ``mode`` (scalar)
      :    Possible access point modes are ``infrastructure`` (the default),
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -475,82 +475,54 @@ interfaces, as well as individual wifi networks, by means of the ``auth`` block.
 
 ``auth`` (mapping)
 
-:    Specifies authentication settings for a device of type ``ethernets:`` or
-     ``wifis:``, or a single wifi ``access-points:`` entry. If ``auth`` blocks
-     are specified on both a wifi device and one of its access points, only the
-     settings in the access point's ``auth`` block are taken into account;
-     however, the ``auth`` block of the wifi device, if present, is used for any
-     ``access-points:`` without their own ``auth`` block.
+:    Specifies authentication settings for a device of type ``ethernets:``, or
+     an ``access-points:`` entry on a ``wifis:`` device.
 
-     Example:
+     The ``auth`` block supports the following properties:
 
-          wifis:
-            wlp1s0:
-              auth:
-                key-mgmt: wpa-psk
-                psk: "d3f4ul7"
-              access-points:
-                office:
-                  auth:
-                    # the authentication settings above are ignored for the
-                    # wifi named "office"
-                    key-mgmt: wpa-eap
-                    eap-method: ttls
-                    anonymous-identity: "@internal.example.com"
-                    identity: "joe@internal.example.com"
-                    password: "v4ryS3kr1t"
-                home:
-                  mode: infrastructure
-                  # the authentication settings above are applied to the wifi
-                  # named "home"
+     ``key-management`` (scalar)
+     :    The supported key management modes are ``none`` (no key management);
+          ``wpa-psk`` (WPA with pre-shared key, common for home wifi);
+          ``wpa-eap`` (WPA with EAP, common for enterprise wifi); and ``802.1x``
+          (used primarily for wired Ethernet connections).
 
-      The ``auth`` block supports the following properties:
+     The following properties can be used if ``key-management`` is ``wpa-psk``:
 
-      ``key-mgmt`` (scalar)
-      :    The supported key management modes are ``none`` (no key management),
-           ``wpa-psk`` (WPA with pre-shared key, common for home wifi),
-           ``wpa-eap`` (WPA with EAP, common for enterprise wifi) and ``8021x``
-           (used primarily for wired Ethernet connections).
+     ``psk`` (scalar)
+     :    The pre-shared key for this connection.
 
-      The following properties can be used if ``key-mgmt`` is ``wpa-psk``:
+     The following properties can be used if ``key-management`` is ``wpa-eap``
+     or ``802.1x``:
 
-      ``psk`` (scalar)
-      :    The pre-shared key for this connection. Useful only if ``key-mgmt``
-           is ``wpa-psk``.
+     ``eap-method`` (scalar)
+     :    The supported EAP methods are ``tls`` (TLS), ``peap`` (Protected EAP),
+          and ``ttls`` (Tunneled TLS).
 
-      ``eap-method`` (scalar)
-      :    The supported EAP methods are ``tls`` (TLS), ``peap`` (Protected EAP)
-           and ``ttls`` (Tunneled TLS). Useful only if ``key-mgmt`` is
-           ``wpa-eap`` or ``8021x``.
+     ``identity`` (scalar)
+     :    The identity to use for EAP.
 
-      The following properties can be used if ``key-mgmt`` is ``wpa-eap`` or
-      ``8021x``:
+     ``anonymous-identity`` (scalar)
+     :    The identity to pass over the unencrypted channel if the chosen EAP
+          method supports passing a different tunnelled identity.
 
-      ``identity`` (scalar)
-      :    The identity to use for EAP.
+     ``password`` (scalar)
+     :    The password string for EAP.
 
-      ``anonymous-identity`` (scalar)
-      :    The identity to pass over the unencrypted channel if the chosen EAP
-           method supports passing a different tunnelled identity.
+     ``ca-certificate`` (scalar)
+     :    Path to a file with one or more trusted certificate authority (CA)
+          certificates.
 
-      ``password`` (scalar)
-      :    The password string for EAP.
+     ``client-certificate`` (scalar)
+     :    Path to a file containing the certificate to be used by the client
+          during authentication.
 
-      ``ca-certificate`` (scalar)
-      :    Path to a file with one or more trusted certificate authority (CA)
-           certificates.
+     ``client-key`` (scalar)
+     :    Path to a file containing the private key corresponding to
+          ``client-certificate``.
 
-      ``client-certificate`` (scalar)
-      :    Path to a file containing the certificate to be used by the client
-           during authentication.
-
-      ``client-key`` (scalar)
-      :    Path to a file containing the private key corresponding to
-           ``client-certificate``.
-
-      ``client-key-password`` (scalar)
-      :    Password to use to decrypt the private key specified in
-           ``client-key`` if it is encrypted.
+     ``client-key-password`` (scalar)
+     :    Password to use to decrypt the private key specified in
+          ``client-key`` if it is encrypted.
 
 
 ## Properties for device type ``ethernets:``
@@ -578,7 +550,7 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           is equivalent to
 
               auth:
-                key-mgmt: wpa-psk
+                key-management: wpa-psk
                 psk: "S3kr1t"
 
      ``mode`` (scalar)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -469,6 +469,90 @@ These options are available for all types of interfaces.
      :    Match this policy rule based on the type of service number applied to
           the traffic.
 
+## Authentication
+Netplan supports advanced authentication settings for ethernet and wifi
+interfaces, as well as individual wifi networks, by means of the ``auth`` block.
+
+``auth`` (mapping)
+
+:    Specifies authentication settings for a device of type ``ethernets:`` or
+     ``wifis:``, or a single wifi ``access-points:`` entry. If ``auth`` blocks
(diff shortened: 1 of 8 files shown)


cyphermox:
Looks great! Mind fixing the conflicts so we can cleanly merge this?

RavuAlHemio:
Sure, no problem; I've fixed the conflict and made a first pass at `doc/netplan.md` .

Would you prefer to merge the current changes or should I implement additional options first (to gain feature parity with [NetworkManager's 802.1x settings](https://developer.gnome.org/NetworkManager/stable/nm-settings.html#id-1.2.7.4.4))?

doc/netplan.md
```diff
@@ -356,9 +440,17 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+          this nor an ``auth`` block are given, the network is assumed to be
+          open. The setting
+
+              password: "S3kr1t"
+
+          is equivalent to
+
+              auth:
+                key-mgmt: wpa-psk
+                psk: "S3kr1t"
```
vorlonofportland: @cyphermox do you think we should be marking 'password' as deprecated in this context?

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+      :    The supported key management modes are ``none`` (no key management),
+           ``wpa-psk`` (WPA with pre-shared key, common for home wifi),
+           ``wpa-eap`` (WPA with EAP, common for enterprise wifi) and ``8021x``
+           (used primarily for wired Ethernet connections).
+
+      The following properties can be used if ``key-mgmt`` is ``wpa-psk``:
+
+      ``psk`` (scalar)
+      :    The pre-shared key for this connection. Useful only if ``key-mgmt``
+           is ``wpa-psk``.
```
vorlonofportland: what is the rationale for using a different key name for wpa-psk vs other auth types? "password" and "pre-shared key" are synonymous, and used interchangeably in the wild; I don't think it matters that wpa-psk doesn't have an a username associated with the password, and that we should still just call it 'password'.

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+
+      ``eap-method`` (scalar)
+      :    The supported EAP methods are ``tls`` (TLS), ``peap`` (Protected EAP)
+           and ``ttls`` (Tunneled TLS). Useful only if ``key-mgmt`` is
+           ``wpa-eap`` or ``8021x``.
```
vorlonofportland: This doesn't match the preceding header.  If it's only valid fow wpa-eap and 8021x, then it should be in the section below, and the last sentence can be dropped.

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+
+      ``identity`` (scalar)
+      :    The identity to use for EAP.
+
+      ``anonymous-identity`` (scalar)
+      :    The identity to pass over the unencrypted channel if the chosen EAP
+           method supports passing a different tunnelled identity.
+
+      ``password`` (scalar)
+      :    The password string for EAP.
```
vorlonofportland: Inconsistent nomenclature.  My understanding is that this is valid for 802.1x as well as for EAP, but the key documentation only mentions EAP.

Which raises another question, my understanding is that 802.1x and EAP are in fact the same protocol, and the only difference is 802.1x is standardized for ethernet.  Why do we support 'wpa-eap' and '8021x' as two separate values for 'key-mgmt' if it's the same protocol?  If these are synonymous, or if the intent is that one should only be used in an ethernet context and the other should only be used in a wifi context, then I think we should collapse them into a single name instead of contributing confusion by having different names for the same thing.

If you consider existing naming so entrenched that saying 'wpa-eap' on an ethernet, or '8021x' on a wifi connection, then I think it should be clear in the documentation that wpa-eap is ONLY supported on wifi (along with wpa-psk), and 8021x is ONLY supported on ethernet.

doc/netplan.md
cyphermox: IMHO password: shouldn't be deprecated, it's a very convenient shorthand for
"""
key-mgmt: wpa-psk
psk: "password"
"""

It certainly makes it a lot easier to remember how to quickly write up a netplan for the most common wireless (WPA personal) without having to resort to doc. At least, that's a clear improvement on my personal workflow for writing wpasupplicant config.

doc/netplan.md
cyphermox: +1, for consistency and because "password" is more human-parseable than "psk", but I do understand the logic of being able to map 1:1 between configuration parameters for wpasupplicant/NM.

WPA personal has "password" being actually mapped to "psk" in supplicant config. "password" here is convenient to remember / easier for a human to map to what it's used for.

WPA enterprise / EAP has "password" being an actual password for the username for the connection.

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+                  mode: infrastructure
+                  # the authentication settings above are applied to the wifi
+                  # named "home"
+
+      The ``auth`` block supports the following properties:
+
+      ``key-mgmt`` (scalar)
+      :    The supported key management modes are ``none`` (no key management),
+           ``wpa-psk`` (WPA with pre-shared key, common for home wifi),
+           ``wpa-eap`` (WPA with EAP, common for enterprise wifi) and ``8021x``
```
cyphermox: Let's use "802.1x" as the value. For a scalar value, it's perfectly fine to use and should be easier to figure out than whether one should write "8021x" or "802-1x" -- at least, it's exactly the name of the standard.

Similarly, what about dropping "wpa-" from the other names? Would that work and still avoid clashing with other key-management methods?

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+                    identity: "joe@internal.example.com"
+                    password: "v4ryS3kr1t"
+                home:
+                  mode: infrastructure
+                  # the authentication settings above are applied to the wifi
+                  # named "home"
+
+      The ``auth`` block supports the following properties:
+
+      ``key-mgmt`` (scalar)
```
cyphermox: Is there anything more verbose  / meaningful to use than key-mgmt? Can we longform this to "key-management"? What about "protocol"?

doc/netplan.md
cyphermox: +1, let's use "password" everywhere unless it conflicts with something else that is more obviously a password.

doc/netplan.md
cyphermox: Let's name this just "method", please.

doc/netplan.md
```diff
@@ -340,6 +340,90 @@ These options are available for all types of interfaces.
...
+interfaces, as well as individual wifi networks, by means of the ``auth`` block.
+
+``auth`` (mapping)
+
+:    Specifies authentication settings for a device of type ``ethernets:`` or
+     ``wifis:``, or a single wifi ``access-points:`` entry. If ``auth`` blocks
+     are specified on both a wifi device and one of its access points, only the
+     settings in the access point's ``auth`` block are taken into account;
+     however, the ``auth`` block of the wifi device, if present, is used for any
+     ``access-points:`` without their own ``auth`` block.
```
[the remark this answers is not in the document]
RavuAlHemio: I agree. I wasn't sure whether the inheritance semantics make all that much sense; I was probably taking consistency ("an interface should have an `auth` block") too seriously. I'll remove it (and modify the documentation accordingly).

doc/netplan.md
RavuAlHemio: Oops, that was a relic from the initial structure of the documentation before I decided to insert those headers.

doc/netplan.md
RavuAlHemio: I guess it boils down to whether netplan should provide a straightforward mapping between its own options and wpa_supplicant/NM options (although we can certainly change some of the option names), or if there should be some reinterpretation (i.e. knowing whether `password` should be passed as a PSK or as an EAP password depending on `key-mgmt`).

For what it's worth, the [current sample wpa_supplicant.conf](https://w1.fi/cgit/hostap/tree/wpa_supplicant/wpa_supplicant.conf?id=0fe3ede0a4f6fe984bbaad90f1a9d9e8decc155d#n1708) allows both to be set simultaneously, supplying one or the other depending on which authentication mechanism is announced by the other end. However, wpa_supplicant also allows setting multiple values for key_mgmt, picking whichever one the other side claims to support; I don't think making this functionality available in netplan makes much sense.

doc/netplan.md
RavuAlHemio: It appears the options dictate the behavior of wpa_supplicant: with `key_mgmt=IEEE8021X`, it performs 802.1X authentication; with `key_mgmt=IEEE8021X` and nonzero `eapol_flags`, it performs 802.1X authentication and requests WEP keys; and with `key_mgmt=WPA-EAP`, it performs 802.1X authentication and then the WPA four-way handshake.

If WEP support is categorically not within the scope of netplan, I can fold both into `eap` and let the value depend on the device type.

doc/netplan.md
RavuAlHemio: Dropping "wpa-" might become a problem if NM starts supporting additional key management schemes that wpa_supplicant [already supports](https://w1.fi/cgit/hostap/tree/wpa_supplicant/wpa_supplicant.conf?id=0fe3ede0a4f6fe984bbaad90f1a9d9e8decc155d#n870). Not sure how likely that is, though.

doc/netplan.md
RavuAlHemio: *protocol* is apparently used to mean the choice between [WPA and WPA2](https://w1.fi/cgit/hostap/tree/wpa_supplicant/wpa_supplicant.conf?id=0fe3ede0a4f6fe984bbaad90f1a9d9e8decc155d#n865), so overloading that might be risky. The long form (*management*) is probably okay, although the longer the word, the greater the potential for typos.

doc/netplan.md
vorlonofportland: Thanks, it's helpful to understand the nuance that 802.1x means bare 802.1x without the wpa setup.

I don't want us to paint ourselves into a corner with the key names as a result of ruling out possible use cases.  I also don't want us to make the schema more cumbersome only in the interest of supporting obsolete and insecure protocols that no one would ever reasonably do a new deployment of via netplan.

Are there any valid reasons to use 802.1x either with WEP or with no encryption at all on wifi?

Thinking forward to a possible eventual replacement for wpa2, how would we expect that to be represented?  Would we expect to see an additional set of 'key-mgmt' values for 'wpa3-psk' and 'wpa3-eap'?  If so, I believe it makes sense to keep the current separation between '802.1x' and 'wpa-eap' instead of treating them as synonyms or having context-dependent interpretation.

doc/netplan.md
vorlonofportland: I agree with cyphermox that 'key-management' is better than 'key-mgmt'.  Yes there's more chance for typos, but it has the advantage of being spell-checkable in wetware by english speakers, and it makes the meaning clear.

See also: creat(2).

doc/netplan.md
vorlonofportland: The guiding vision here is that we should learn from and be informed by prior art in naming, but not be constrained by it when it's not user-friendly.  In the short term, there will always be people familiar with the old thing who are dissatisfied with a name change.  In the long term, these people will be outnumbered by others who are learning it new.

netplan.yaml is intended to model network configuration - not any particular backend implementation.

doc/netplan.md
vorlonofportland: @cyphermox how confident can we be that we will never need to support other types of "methods" on 802.1x authentication besides EAP methods (possibly at the same time)?  How clear is it to the user that "method" refers to EAP without the prefix?

doc/netplan.md
RavuAlHemio: > Are there any valid reasons to use 802.1x either with WEP or with no encryption at all on wifi?

I cannot think of one. It might exist, but it's probably so esoteric that netplan need not support it until someone announces their need on the bug tracker.

> Thinking forward to a possible eventual replacement for wpa2, how would we expect that to be represented? Would we expect to see an additional set of 'key-mgmt' values for 'wpa3-psk' and 'wpa3-eap'?

WPA3 should be published this year; according to [this Aruba blog post](http://community.arubanetworks.com/t5/Technology-Blog/WPA3-The-Next-Generation-in-Secure-Mobility/ba-p/410832), unsecured networks can be supplanted by OWE (RFC8110) and PSK authentication is being superseded by something named *Simultaneous Authentication of Equals*, but all other changes appear to only have an effect on the more specific WPA parameters (baseline rules for EAP; CNSA as a more secure alternative to WPA's TKIP and WPA2's CCMP).

The exhaustive list would thereby probably become `802.1x`, `wpa-psk`, `wpa-eap`, `wpa-owe` and `wpa-sae`. (I wouldn't explicitly mention a WPA version in the values, since most of the methods span multiple versions; e.g. WPA-EAP spans "WPA1" to WPA3.) I bet differentiating between `wpa-psk` and `wpa-sae` will be pretty challenging to users; while wpa_supplicant can probably make use of its "multiple key_mgmt values" feature, NetworkManager only supports one value. I wonder how other Wi-Fi client software will deal with this once WPA3 is released.

There is always the "threat" of a new and completely different security architecture succeeding WPA3, but that could full well get rid of the concept of *key management*, necessitating new netplan configuration keys anyway.

It's pretty obvious that UX was out of scope for the teams designing network authentication, and I don't know how well netplan can counteract that...

doc/netplan.md
cyphermox: It's a convenient way to refer to these (commonly called EAP methods) and being more obvious than "authentication"; you could have similar deeper-level config options for other modes aside from EAP -- but right now you only have a choice of using "eap" or "psk".

https://en.wikipedia.org/wiki/Extensible_Authentication_Protocol

doc/netplan.md
cyphermox: That's true of FT-*, but I think we'll be fine calling those whatever value they are exactly, whereas any option starting with "WPA-" we'd be better off not having "WPA-" in the name -- it's shorter, and we don't need that bit -- it's already obvious that it's wpa. In my mind, people are more likely to know, without looking at doc, to write "psk" or "eap", and wouldn't know that they'd need to name those "wpa-psk", "wpa-eap" without looking at the doc.

I'm not strongly attached to either way, but I think anything we can do to make "looking at the doc" less necessary to configure some of the "simplest" complex network configurations (like WPA) is a win.

doc/netplan.md
jsoref: I think unencrypted wifi isn't an unreasonable thing to support. I can imagine wanting users to know about a network name + sid, wanting to set up custom ip addresses / dns servers, and relying on HTTPS or a VPN.

The reality is that most networks *are* insecure and probably *sniffable*. The WPAx level is just an attempt to prevent someone from stealing the credentials for the network, it isn't a guarantee that no one on the network will sniff. For that, one needs End-to-end encryption (TLS/VPN).

cyphermox:
Hi @RavuAlHemio , are you still planning on applying the few changes that were requested? I would very much like to start merging this ASAP, at least in a separate branch while we finish working out any details about the schema.

RavuAlHemio:
Hi @cyphermox, sorry, I've been busy with other projects on the one hand and unsure how to proceed given the running schema discussions on the other hand. I have now implemented those changes where you, @vorlonofportland and @jsoref were in agreement.

cyphermox:
Great, design-wise this looks good (modulo the small doc issue); so I'll start to rebase and merge this manually and do more review of the code in the process. AFAICT it all looks good to me already.

Thanks for spending the time to make the tests pass a rebase on the changes I made to the generator tests. :)

cyphermox:
Closing; this was merged.

doc/netplan.md
mckenzm: Ultimately any network is sniffable, security is relative. There is no absolute security, and we should not rely upon wifi encryption alone, it would be one of several layers. Not supporting WEP means having to have dedicated open SSID's with other protections applied as necessary. Firmware limited to WEP cannot simply be replaced in all cases. Generally it is disappointing wherever a standard is not supported, it renders the whole of product unusable. It is not the job of the framework to make these decisions.

q2dg:
There's no option for OWE yet...

jsoref:
For the curious, this was merged in at least:
- a73ed0c7f10079dae76f058667e350c364f84f4c
- a07594a1580055e11d2c096e83524462a4541bbc
- f4808f5bf16586fe39652cefa885e69137d5406c

Each commit was included in [0.95](https://github.com/canonical/netplan/releases/tag/0.95).

Anyone interested in anything else should read https://github.com/canonical/netplan/blob/main/CONTRIBUTING

PR #33, opened 2018-05-29
Fix support for link-scope routes

cyphermox:
Signed-off-by: Mathieu Trudel-Lapierre <mathieu.trudel-lapierre@canonical.com>


daxtens:
This looks good; I think it will solve a bug I identified with my config tester, and [LP: #1769550](https://bugs.launchpad.net/netplan/+bug/1769550) which was reported independently.

PR #35, opened 2018-07-20
Support IPv6 Privacy Extensions

daxtens:
[LP: #1750392](https://bugs.launchpad.net/ubuntu/+source/nplan/+bug/1750392)

Add 'ip6-privacy' as a boolean key, add docs, and wire it up for
systemd-networkd and NetworkManager.

Test procedure and results are in the LP bug.

## Checklist

- [👍] Runs 'make check' successfully.
- [💯] Retains 100% code coverage (make check-coverage).
- [👍] New/changed keys in YAML format are documented.
- [👍] Closes an open bug in Launchpad.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -159,6 +159,12 @@ Virtual devices
     Note that **``rdnssd``**(8) is required to use RDNSS with networkd. No extra
     software is required for NetworkManager.
 
+``ipv6-privacy`` (bool)
+
+:   Enable IPv6 Privacy Extensions (RFC 4941) for the specified interface, and
+    prefer temporary addresses. Defaults to false - no privacy extensions. There
+    is currently no way to have a private address but prefer the public address.
+
 ``link-local`` (sequence of scalars)
 
 :   Configure the link-local addresses to bring up. Valid options are 'ipv4'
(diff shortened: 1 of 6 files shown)


topia:
This is very nice changes to merge; I could enable IPv6 privacy extensions with this patch, and I believe this is backward-compatible (no harm to merge).

cyphermox:
@vorlonofportland Opinion on this schema change?

vorlonofportland:
On Thu, Aug 30, 2018 at 09:09:02AM -0700, Mathieu Trudel-Lapierre wrote:
> @vorlonofportland Opinion on this schema change?

Let's look at internal consistency of the schema (implemented or specified)
wrt ipv6 option naming.  We currently have the following:

 - link-local: [ ipv6 ]
 - optional-addresses: [ dhcp6, ipv6-ra ]
 - ipv6-mtu
 - dhcp6
 - gateway6 (https://bugs.launchpad.net/netplan/+bug/1756590)
 - addresses
 - nameservers
 - routes

etc.

The fields that deal purely with IPv4 and IPv6 addresses are unambiguous due
to the difference in the format of the addresses so don't need annotations
on the field names.

For the others, we have a mix of both '6' and 'v6' as suffix: dhcp6 vs.
ipv6.

I think that calling dhcp6 "dhcp6" instead of "dhcpv6" is justifiable
because this is a common way to refer to this protocol.  (NB: Google search
results strongly prefer 'dhcpv6'.  But I don't propose changing that in
netplan at this time.)

I think that ipv6 (vs "ip6") is justifiable and strongly preferable as a
name for this protocol in the schema.

And I think new ipv6 options should be named consistently.  Therefore,
please make this 'ipv6-privacy' rather than 'ip6-privacy'.

daxtens:
I have rebased this on master.

cyphermox:
Merged at https://github.com/CanonicalLtd/netplan/commit/9a7c589139d98731b692f7fd91d4982336fbbf12

PR #36, opened 2018-07-20
Explicitly set permissive umask on networkd .network files

daxtens:
[LP: #1736965](https://bugs.launchpad.net/ubuntu/+source/nplan/+bug/1736965)

Without this, a restrictive umask causes systemd-networkd to fail
with permission denied errors.

## Checklist

[👍] Runs 'make check' successfully.
[💯] Retains 100% code coverage (make check-coverage).
[n/a] New/changed keys in YAML format are documented.
[🎉] Closes an open bug in Launchpad.

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -209,6 +209,7 @@ static void
 write_netdev_file(net_definition* def, const char* rootdir, const char* path)
 {
     GString* s = NULL;
+    mode_t orig_umask;
 
     g_assert(def->type >= ND_VIRTUAL);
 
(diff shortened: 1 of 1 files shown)


daxtens:
v2: cover netdev as well as network, and add an extra LP bug to the commit description

cyphermox:
Should also apply to the .link files if we write them

cyphermox:
I merged this by hand, and added the umask to .link and .rules files. wpasupplicant config obviously stays 077, given that it might contain secrets.

daxtens:
Thanks!

PR #37, opened 2018-07-20
Add support for WPA2 Enterprise via wpa_supplicant

jameyhicks:
## Description

Adds support for access point key_mgmt and identity so that netplan can generate wpa_supplicant configuration files that work with WPA2 Enterprise.

Fixes bug 1739578

## Checklist

- [x ] Runs 'make check' successfully.
- [x ] Retains 100% code coverage (make check-coverage).
- [x ] New/changed keys in YAML format are documented.
- [x ] \(Optional\) Closes an open bug in Launchpad.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -379,15 +379,25 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
      supported properties:
 
      ``password`` (scalar)
-     :    Enable WPA2 authentication and set the passphrase for it. If not
-          given, the network is assumed to be open. Other authentication modes
-          are not currently supported.
+     :    Sets the passphrase for WPA2 Personal and Enterprise networks. If not
+          given, the network is assumed to be open.
 
      ``mode`` (scalar)
      :    Possible access point modes are ``infrastructure`` (the default),
           ``ap`` (create an access point to which other devices can connect),
           and ``adhoc`` (peer to peer networks without a central access point).
           ``ap`` is only supported with NetworkManager.
+
+     ``key_mgmt`` (scalar)
+     :    Set this to WPA-EAP to connect to WPA2 Enterprise WiFi
+          networks. Optional for WPA2 Personal (WPA-PSK).  Possible key_mgmt
+          values are WPA-PSK and WPA-EAP. Defaults to WPA-PSK if password is
+          provided.
+
+     ``identity`` (scalar)
+     :    Provides the identity for authenticating with WPA2 Enterprise
+          (WPA-EAP). Not needed for WPA2 Personal.
+
 
 ## Properties for device type ``bridges:``
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -420,6 +420,22 @@ handle_access_point_password(yaml_document_t* doc, yaml_node_t* node, const void
 }
 
 static gboolean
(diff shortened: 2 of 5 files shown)


tests/generate.py
```diff
@@ -64,7 +65,7 @@ def generate(self, yaml, expect_fail=False, extra_args=[], confs=None):
                 with open(os.path.join(self.confdir, f + '.yaml'), 'w') as f:
                     f.write(contents)
 
-        argv = [exe_generate, '--root-dir', self.workdir.name] + extra_args
+        argv = ['catchsegv', exe_generate, '--root-dir', self.workdir.name] + extra_args
```
daxtens: What's this change for? It looks from the man page to be a good idea, but it's not documented in the commit message.

daxtens: (Likewise for some of your other changes in this file.)

doc/netplan.md
```diff
@@ -380,15 +380,20 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
           ``ap`` (create an access point to which other devices can connect),
           and ``adhoc`` (peer to peer networks without a central access point).
           ``ap`` is only supported with NetworkManager.
 
+     ``key_mgmt`` (scalar)
+     :    Possible key_mgmt values are WPA-PSK and WPA-EAP. Optional.
+
+     ``identity`` (scalar)
+     :    Provides the identity for authenticating with WPA-EAP.
+
```
daxtens: This should be squashed in to the previous comment so that the build succeeds at each point (which is important for bisectability).

Also those descriptions are really terse, and should at the very least mention that they are for WPA2 Enterprise and not required for WPA2 Personal.

tests/generate.py
jameyhicks: I did not mean to leave the catchsegv in there -- I inserted it to see why generate was faulting.

src/parse.c
```diff
@@ -579,6 +597,7 @@ handle_wifi_access_points(yaml_document_t* doc, yaml_node_t* node, const void* d
 
         g_assert(cur_access_point == NULL);
         cur_access_point = g_new0(wifi_access_point, 1);
+	memset(cur_access_point, 0, sizeof(wifi_access_point));
```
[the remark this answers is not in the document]
jameyhicks: This call to memset seemed to cure a segv that was happening in the test app. However, I've removed it now and there is no segv. Must have been an inconsistent build somehow.

jameyhicks:
The build seems to have a DNS problem.

jsoref:
Not a dns problem, the builder didn't do `apt-get update` first which meant that the system's sense of the apt repository was stale, so when it asked the apt repository for a no longer current package, the repository replied `404`. (This hit me, and upon reviewing PRs, quite a few others.)

jsoref:
This feels similar to #29

jameyhicks:
does look like it's addressing the same problem as #29. If that one is going to be merged, I can try it out and close this one.

cyphermox:
We're looking at both; if there's anything that isn't covered in #29, then I'll merge them too.

jameyhicks:
Thanks!

cyphermox:
I think everything is already well covered by the code I landed for WPA2 Enterprise in master (and which is currently in -proposed repositories for Ubuntu); closing since there isn't anything else to merge there.

Thanks again for your work; we truly appreciate your interest in netplan :)

PR #39, opened 2018-07-23
Add ability to set interface route metrics

daxtens:
[LP: #1771834](https://bugs.launchpad.net/netplan/+bug/1771834)

Currently, we set a route metric of 600 for routes on a wifi device
and 100 on other devices. We allow static routes to set their own
metric, but have no way to set the metric for routes that come from
DHCP, or the route to the subnet. Add this ability through a 'metric'
key on an interface.

This is particularly useful for systems with multiple DHCP interfaces
where one should be preferred.

Add a 'metric' key to interfaces, plumb it through, document it,
add tests.

## Checklist

- [Y] New/changed keys in YAML format are documented.
- [Y]  Closes an open bug in Launchpad.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -254,6 +254,15 @@ similar to ``gateway*``, and ``search:`` is a list of search domains.
      only, due to interactions with device renaming in udev. Match
      devices by MAC when setting MTU.
 
+``metric`` (scalar)
+
+:    Set the relative priority of routes on this interface. Must be a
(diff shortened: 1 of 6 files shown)


cyphermox:
No; this should be folded under the "dhcp" stanza; we've started to define what this should look like in https://trello.com/c/8lTVQoPA/9-multi-nic-dhcp-default-route-control-extended-control-over-dhcp-response-config

schopin-pro:
Closing this PR as I believe it to be superseded by #64

PR #40, opened 2018-07-26
Disable 10-globally-managed-devices.conf if any device uses NM

daxtens:
LP: #1753860

I found that specifying NM as the renderer for a single ethernet
device didn't work - 10-globally-managed-devices.conf still stops
NM from managing it.

Disable that file if any interface uses NM.

I am a little concerned that this might lead to different behaviour
if you hotplug a device when any device is using NM - it will be
managed by NM. This already happens if NM is the global renderer, so
I don't think it's a big issue, but something to be aware of.

## Checklist

- [Y] Closes an open bug in Launchpad.

### DIFF
--- a/src/generate.c
+++ b/src/generate.c
@@ -34,6 +34,7 @@
 static gchar* rootdir;
 static gchar** files;
 static gboolean any_networkd;
+static gboolean any_nm;
 static gchar* mapping_iface;
 
 static GOptionEntry options[] = {
@@ -55,7 +56,8 @@ nd_iterator(gpointer key, gpointer value, gpointer user_data)
 {
     if (write_networkd_conf((net_definition*) value, (const char*) user_data))
         any_networkd = TRUE;
-    write_nm_conf((net_definition*) value, (const char*) user_data);
+    if (write_nm_conf((net_definition*) value, (const char*) user_data))
+        any_nm = TRUE;
 }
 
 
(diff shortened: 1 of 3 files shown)


jsoref:
@daxtens: your concern seems to warrant a release note...

cyphermox:
I think in that case we should also explicitly blacklist devices, to make sure any ethernet managed by networkd is still blacklisted in NM -- otherwise you might have weird timeouts due to the two fighting over the same thing.

slyon:
Thank you, this makes lots of sense wrt to LP#1951653

I took inspiration from this PR and implemented a similar solution (including much refiened NetworkManager allow- and denylists). I'm closing this PR in favour of #276

setharnold:
Oooh, would this provide a fix for the problems discussed at https://askubuntu.com/questions/71159/network-manager-says-device-not-managed and similar? It'd sure be nice to get this fixed! :) Thanks @slyon!

slyon:
@setharnold Yes, I think the final fix for such issues would be to drop the `/usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf` file from the network-manager package, as suggested in [LP#1951653](https://pad.lv/1951653)

netplan's PR #276 would be a first step into that direction.

setharnold:
Thanks!! This has been a pain point for ages.

PR #42, opened 2018-08-13
Assorted fixes from new test tool

daxtens:
## Description

I've been writing a [property-based testing tool](https://github.com/daxtens/netplan-config-tester) for netplan and it's got to the stage where it's found some interesting bugs and also seems to be pretty stable, so I figure now is a good time to push it to github and send the first tranche of fixes. It only covers some of the [bugs I've found](https://github.com/daxtens/netplan-config-tester/blob/master/bugs.txt) so far.

## Checklist

- [Y] New/changed keys in YAML format are documented.
- [Y] Closes an open bug in Launchpad.

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -402,8 +402,12 @@ write_nm_conf_access_point(net_definition* def, const char* rootdir, const wifi_
             g_string_append_printf(s, "%s;", g_array_index(def->ip4_nameservers, char*, i));
         g_string_append(s, "\n");
     }
-    write_search_domains(def, s);
-    write_routes(def, s, AF_INET);
+
+    /* We can only write search domains and routes if we have an address */
+    if (def->ip4_addresses || def->dhcp4) {
+        write_search_domains(def, s);
+        write_routes(def, s, AF_INET);
+    }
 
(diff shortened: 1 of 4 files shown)


daxtens:
Apparently in the final patch I wanted PreferredSource=, not Source=; I'll spin v2 next week.

src/nm.c
```diff
@@ -134,6 +134,31 @@ write_routes(const net_definition* def, GString *s, int family)
...
+                exit(1);
+            }
+
+            if (cur_route->from) {
+                g_fprintf(stderr, "ERROR: %s: NetworkManager does not support routes with 'from'\n", def->id);
+                exit(1);
+            }
+
+            if (cur_route->onlink) {
+                g_fprintf(stderr, "ERROR: %s: NetworkManager does not support onlink routes\n", def->id);
```
jsoref: I think this should be `on-link` or possibly `'on-link'`, at least, based on https://github.com/daxtens/netplan-config-tester/blob/e652e0f0ab203f2ed201990a1e9b11feb7a702fa/apply.py#L148

cyphermox:
Merged.

PR #43, opened 2018-08-27
7-dns-overrides-for-dhcp-configuration

jsoref:
This PR enables netplan users to suppress dhcp provided dns servers. Note that systemd-networkd doesn't support suppressing only one at a time, this pr forces one to opt into suppressing both...
```
network:
  version: 2
  ethernets:
    engreen:
      dhcp4: yes
      dhcp4-overrides:
        use-dns: no
      dhcp6-overrides:
        use-dns: no
```

## Description
Part of the work for:
https://trello.com/c/Hbz5Jfol/7-dns-overrides-for-dhcp-configuration

## Checklist

- [x] Runs 'make check' successfully.
- [x] Retains 100% code coverage (make check-coverage).
- [x] Closes an open bug in Launchpad.
https://bugs.launchpad.net/netplan/+bug/1759014

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -159,6 +159,27 @@ Virtual devices
     Note that **``rdnssd``**(8) is required to use RDNSS with networkd. No extra
     software is required for NetworkManager.
 
+``dhcp4-overrides`` (mapping)
+
+:   Control additional DHCP properties.
+    These are the supported fields:
+    * `use-dns`
+
+``dhcp6-overrides`` (mapping)
+
+:   Control additional DHCP properties.
+    These are the supported fields:
+    * `use-dns`
+
+``use-dns`` (bool)
+
+:   Enable DHCP4/DHCP6 to provide dns servers. On by default.
+
+    If you are planning to override your DHCP server provided DNS server,
+    set this to `no`.
+    
+    **Note:** networkd does not support disabling dhcp for only one of dhcp4/dhcp6, you will need to disable it for both.
+
 ``link-local`` (sequence of scalars)
 
 :   Configure the link-local addresses to bring up. Valid options are 'ipv4'
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -410,9 +410,17 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
     }
 
     if (def->dhcp4 || def->dhcp6) {
+        g_string_append_printf(s, "\n[DHCP]\n");
+        if (def->use_dns4 != def->use_dns6) {
+            /* networkd does not support disabling dhcp dns for only one of dhcp4/dhcp6 */
+            g_fprintf(stderr, "ERROR: %s: networkd backend does not support disabling dhcp for only one of dhcp4/dhcp6\n", def->id);
+            exit(1);
+        }
+        if (!def->use_dns4 || !def->use_dns6)
+            g_string_append(s, "UseDNS=false\n");
         /* isc-dhcp dhclient compatible UseMTU, networkd default is to
          * not accept MTU, which breaks clouds */
-        g_string_append_printf(s, "\n[DHCP]\nUseMTU=true\n");
+        g_string_append_printf(s, "UseMTU=true\n");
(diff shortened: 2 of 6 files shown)


doc/netplan.md
```diff
@@ -159,6 +159,13 @@ Virtual devices
     Note that **``rdnssd``**(8) is required to use RDNSS with networkd. No extra
     software is required for NetworkManager.
 
+``use-dns`` (bool)
+
+:   Enable DHCP to provide dns servers. On by default.
+
+    If you are planning to override your DHCP server provided DNS server,
+    set this to `no`.
+
```
daxtens: This description should note that this property is only used by networkd.

It also doesn't make it clear that (per your current schema design) it needs to be `dhcp: {use-dns: false}` - reading that I thought `use-dns` was at the same level as `dhcp4`/`dhcp6` etc.

tests/generate.py
```diff
@@ -816,6 +816,25 @@ def test_eth_manual_addresses_dhcp(self):
 [DHCP]
 UseMTU=true
 RouteMetric=100
+'''})
+
+   def test_eth_use_dns(self):
```
daxtens: This doesn't seem to be properly indented:

```
$ make check-coverage 
...
/usr/bin/pyflakes3 netplan/  tests/test_terminal.py tests/generate.py tests/cli.py tests/test_configmanager.py tests/integration.py
tests/generate.py:821:31: unindent does not match any outer indentation level
   def test_eth_use_dns(self):
                              ^
Makefile:43: recipe for target 'linting' failed
make[1]: *** [linting] Error 1
make[1]: Leaving directory '/home/dja/dev/nplan/netplan'
Makefile:49: recipe for target 'pre-coverage' failed
make: *** [pre-coverage] Error 2
```

If I fix that, I get an enormous amount of test failures because as I noted above you are not setting this to true by default.

src/networkd.c
```diff
@@ -314,6 +314,8 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
         g_string_append(s, "DHCP=ipv4\n");
     else if (def->dhcp6)
         g_string_append(s, "DHCP=ipv6\n");
+    if (!def->use_dns)
+        g_string_append(s, "UseDNS=false\n");
```
daxtens: Here you check for `!def->use_dns`, but I can't see anywhere where you'd set the default to be true. So I think this would change behaviour?

tests/generate.py
daxtens: I can't add a comment pointing to it directly, but you want to set this to true in `handle_network_type` at around line 1482.

src/networkd.c
daxtens: Oh, and please could you add code to nm.c that throws an error or at least a warning if the value is false for an interface rendered by NM?

jsoref: I'll use `ignore-auto-dns` per https://developer.gnome.org/NetworkManager/stable/nm-settings.html

daxtens:
Ah brilliant! Sorry, I didn't check to see if there was a NM equivalent, I
just assumed it didn't exist. Thanks!

jsoref:
Probably worth considering whether `use-dns` belongs under a `dhcp` or under `dhcp4` and `dhcp6`. My implementation is based on the trello doc.

jsoref:
It appears to be a limitation of networkd: https://www.freedesktop.org/software/systemd/man/systemd.network.html

The code is more or less distinct for 4/6, so I'm happy to split it. But, why would you want "dhcp4-overrides" instead of just putting it into "dhcp4" and similarly for 6?

vorlonofportland:
On Fri, Aug 31, 2018 at 04:20:51AM -0700, Josh Soref wrote:
> Ah.
> Ok, shall I change this PR to us use overrides?

Yes, please.

cyphermox:
Closing without merge, as PR #48 was just merged and includes the changes here. Still, thanks for the contribution, and sorry for the delay in looking at the PRs. I look forward to more contributions from you :)

PR #48, opened 2018-10-10
Add DHCP behavior override options for networkd.

alanwj:
## Description

Adds dhcp4-overrides and dhcp6-overrides. These mappings override
specific DHCP behaviors.

Example usage:

```
network:
  version: 2
  ethernets:
    eth0:
      dhcp4: yes
      dhcp4-overrides:
        use-dns: no
```
The available options added in this commit are:

-   `use-dns`
-   `use-ntp`
-   `send-hostname`
-   `use-hostname`
-   `hostname`

This fixes launchpad bug [#1759014](https://bugs.launchpad.net/netplan/+bug/1759014).

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -190,10 +190,15 @@ Virtual devices
     device's MAC address as a unique identifier rather than a RFC4361-compliant
     Client ID. This has no effect when NetworkManager is used as a renderer.
 
- ``dhcp-options`` (mapping)
+ ``dhcp4-overrides`` (mapping)
 
- :  (networkd backend only) Additional DHCP options; see the
-    ``DHCP Options`` section below.
+ :  (networkd backend only) Overrides default DHCP behavior; see the
+    ``DHCP Overrides`` section below.
+
+ ``dhcp6-overrides`` (mapping)
+
+ :  (networkd backend only) Overrides default DHCP behavior; see the
+    ``DHCP Overrides`` section below.
 
 ``accept-ra`` (bool)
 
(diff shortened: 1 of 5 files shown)


dave93cab:
Can this be extended to add other networkd DHCP options (e.g. UserClass)

doc/netplan.md
```diff
@@ -190,6 +190,11 @@ Virtual devices
     device's MAC address as a unique identifier rather than a RFC4361-compliant
     Client ID. This has no effect when NetworkManager is used as a renderer.
 
+ ``dhcp-options`` (mapping)
```
vorlonofportland: I really think this needs to be called 'dhcp-overrides', not 'dhcp-options'.  'dhcp-options' is bound to cause confusion with "options" as used in RFC2131.

vorlonofportland: I really think this needs to be called 'dhcp-overrides', not 'dhcp-options'. 'dhcp-options' is bound to cause confusion with "options" as used in RFC2131.

dave93cab: how about dhcp-config

vorlonofportland: I'd be ok with dhcp-config. @cyphermox, thoughts?

alanwj: I just noticed that pull request #43 does something similar, and there it was decided to go with dhcp4-overrides and dhcp6-overrides  (even though networkd doesn't support separate configuration).

Should the decision from that discussion still stand?

alanwj:
> Can this be extended to add other networkd DHCP options (e.g. UserClass)

Yes.  I just did the easy ones here (bools and string).  The ones I didn't include either require parsing that is less trivial, or are networkd options that I don't fully understand and don't feel comfortable putting my name on without more research first.

I plan to eventually follow up with a few other options.  If you look in launchpad bug #1759014 there are already a few requests for additional dhcp options.

doc/netplan.md
vorlonofportland: Indeed it should.

cyphermox:
I'm happy to merge the two PRs provided they all are naming the options "dhcpX-overrides". I don't think we need to block on adding userclass or the other features, those can be added later.

doc/netplan.md
alanwj: I've split dhcp-options into dhcpX-overrides.  I am still working on the docs and tests (probably finish Monday or so), but I figured getting some feedback on this overall approach could be useful.

Also, do you prefer everything to be squashed into a single commit?

alanwj:
I believe I have addressed all the feedback. Please let me know if there are additional changes you would like.

Also, should I squash the commits or do you prefer the commit history?

cyphermox:
I'm happy with a rich history, as long as the commits are sensical and not just "do stuff, fix things".

doc/netplan.md
alanwj: Marking this resolved since I think I have addressed the concerns here.

PR #50, opened 2018-10-15
Fix bug where interface name is wrongly set to ID

markkrj:
The by-mac array should contain the new name (newname variable) instead of the interface ID (phy), so when called in
`new_name = matches['by-mac'][macaddress]` the new name is returned.

Bugs:
https://bugs.launchpad.net/netplan/+bug/1770082
https://bugs.launchpad.net/netplan/+bug/1768827

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -140,9 +140,9 @@ class NetplanApply(utils.NetplanCommand):
             driver = match.get('driver')
             mac = match.get('macaddress')
             if driver:
-                matches['by-driver'][driver] = phy
+                matches['by-driver'][driver] = newname
             if mac:
-                matches['by-mac'][mac] = phy
+                matches['by-mac'][mac] = newname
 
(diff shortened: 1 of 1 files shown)


cyphermox:
Quite correct. I had already "fixed" that locally, but I'll merge your copy.

In my testing though, it didn't seem to necessarily fix the renaming when running 'netplan apply', Was I wrong in my testing?

markkrj:
I ran some tests and it is working for me... I tried with 2 interfaces in the yaml file, renamed multiple times, with different ID's and set-name's and it did honor the "set-name"...  
I posted a patch in the bug thread and it was working, but only because I had only one interface, when I tried adding one more, it tried to rename both to the same name (the last one from the previous loop) because that was the one kept in the 'newname' variable... 

Didn't you forget to set the interface to down before running 'netplan apply'? I ever run it with --debug, so it tell me when I forget to put the interface down before applying...

PR #55, opened 2018-11-09
Fix unknown local variable 'phy' which should be variable 'phys'

raharper:
Used the wrong local variable name here.


markkrj:
I think this is already fixed in https://github.com/CanonicalLtd/netplan/commit/9e5e5a020f945556dc0553046cd72e23da834505

cyphermox:
Indeed; I noticed this and fixed it already. Thanks!

PR #57, opened 2018-11-20
[snap] Add 'snap/snapcraft.yaml'

slashdd:
A snap is a bundle of your app and its dependencies that works without modification across Linux.
Snaps are discoverable and installable from the Snap store.e

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,43 @@
+name: netplan
+version: '0.90'
+summary: Backend-agnostic network configuration in YAML
+description: |
+  Netplan is a utility for easily configuring networking on a linux system.
+  You simply create a YAML description of the required network interfaces and
+  what each should be configured to do. From this description Netplan will
+  generate all the necessary configuration for your chosen renderer tool.
+grade: devel
+confinement: classic
+
+apps:
+  netplan:
+    command: usr/sbin/netplan
+    plugs: [network]
+    environment:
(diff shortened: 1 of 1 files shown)


slashdd:
Push as agreed with cyphermox this morning.

cyphermox:
Looks good to me (but I had already reviewed some parts of it). I have small nitpicks however:

The encoding of python version in PYTHONPATH isn't ideal; can we have that point to usr/lib/python3 instead, or wherever it's not specifically python3.5?

Furthermore, there's a potential issue there building because of pep8/pycodestyle, different releases of Ubuntu have different libraries available for that. This will require some work on the build side to make sure the right one to use can be detected.

slashdd:
Since pycodestyle has been first introduces starting with Bionic, I had to use pep8 in Xenial for the snap testing. I think it might be safe to rely on pycodestyle only since netplan is normally install on release w/ Bionic and late ?

slashdd:
IIRC pep8 is the old name. Now its pycodestyle (formerly called pep8)

slashdd:
- Changed the `PYTHONPATH` to `$SNAP/usr/lib/python3/dist-packages`
- Changed snap name from `netplan-io` to `netplan`. Otherwise it creates a binary in the form of `/snap/bin/netplan-io.netplan` instead of `/snap/bin/netplan`

Note
In the deb world `netplan` is a totally different project `network server for `plan', but chances this package will be snap one day are low IMHO.

cyphermox:
WFM. merging.

PR #59, opened 2018-11-28
Tunnel support in networkd/NM backends

cyphermox:
## Description

Add support for bringing up various tunnel types:

- gre
- sit
- ipip / ipip6 / ip6ip6
- isatap (NetworkManager only)
- ip6gre
- gretap / ip6gretap (networkd only)
- vti / vti6

## Checklist

- [WIP] Runs `make check` successfully.
- [WIP] Retains 100% code coverage (`make check-coverage`).
- [WIP] New/changed keys in YAML format are documented.
- [WIP] \(Optional\) Closes an open bug in Launchpad.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -680,10 +680,6 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
 
 ## Properties for device type ``tunnels:``
 
-``parent`` (scalar)
-
-:    The parent device for the tunnel (optional).
-
 ``mode`` (scalar)
 
 :   Defines the tunnel mode. Valid options are ``sit``, ``gre``, ``ip6gre``,
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -678,6 +678,67 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           ``active-backup``, ``balance-alb``, and ``balance-tlb`` modes.
 
 
+## Properties for device type ``tunnels:``
+
+``parent`` (scalar)
+
+:    The parent device for the tunnel (optional).
+
+``mode`` (scalar)
+
+:   Defines the tunnel mode. Valid options are ``sit``, ``gre``, ``ip6gre``,
+    ``ipip``, ``ipip6``, ``ip6ip6``, ``vti``, and ``vti6``. Additionally,
+    the ``networkd`` backend also supports ``gretap`` and ``ip6gretap`` modes.
+    In addition, the ``NetworkManager`` supports ``isatap`` tunnels.
+
+``local`` (scalar)
+
+:   Defines the address of the local endpoint of the tunnel.
+
+``remote`` (scalar)
+
+:   Defines the address of the remote endpoint of the tunnel.
+
+``key``  (scalar) or (mapping)
+:   See ``keys`` below.
+
+``keys`` (scalar) or (mapping)
+
+:   Define keys to use for the tunnel. The key can be a number or a dotted
+    quad (an IPv4 address). It is used for identification of IP transforms.
+    This is only required for ``vti`` and ``vti6`` when using the networkd
+    backend, and for ``gre`` or ``ip6gre`` tunnels when using the
+    NetworkManager backend.
+
+    This field may be used as a scalar (meaning that a single key is
+    specified and to be used for both input and output key), or as a mapping,
+    where you can then further specify ``input`` and ``output``.
+
+    ``input`` (scalar)
+    :    The input key for the tunnel
+
+    ``output`` (scalar)
+    :    The output key for the tunnel
+
+Examples:
+
+    tunnels:
+      tun0:
+        mode: gre
+        local: ...
+        remote: ...
+        keys:
+          input: 1234
+          output: 5678
+
+    tunnels:
+      tun0:
+        mode: vti6
+        local: ...
+        remote: ...
+        key: 59568549
+
+
 ## Properties for device type ``vlans:``
 
 ``id`` (scalar)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -680,12 +680,19 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
 
 ## Properties for device type ``tunnels:``
 
+Tunnels allow traffic to pass as if it was between systems on the same local
+network, although systems may be far from each other but reachable via the
+Internet. They may be used to support IPv6 traffic on a network where the ISP
+does not provide the service, or to extend and "connect" separate local
+networks. Please see https://en.wikipedia.org/wiki/Tunneling_protocol for
+more general information about tunnels.
+
 ``mode`` (scalar)
 
 :   Defines the tunnel mode. Valid options are ``sit``, ``gre``, ``ip6gre``,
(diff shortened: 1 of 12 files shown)


src/parse.c
```diff
@@ -1488,6 +1571,17 @@ const mapping_entry_handler vlan_def_handlers[] = {
...
 };
 
+const mapping_entry_handler tunnel_def_handlers[] = {
+    COMMON_LINK_HANDLERS,
+    {"mode", YAML_SCALAR_NODE, handle_tunnel_mode},
+    {"parent", YAML_SCALAR_NODE, handle_netdef_id_ref, NULL, netdef_offset(tunnel.parent)},
+    {"local", YAML_SCALAR_NODE, handle_netdef_ip4, NULL, netdef_offset(tunnel.local_ip)},
+    {"remote", YAML_SCALAR_NODE, handle_netdef_ip4, NULL, netdef_offset(tunnel.remote_ip)},
+    {"input-key", YAML_SCALAR_NODE, handle_tunnel_key, NULL, netdef_offset(tunnel.input_key)},
+    {"output-key", YAML_SCALAR_NODE, handle_tunnel_key, NULL, netdef_offset(tunnel.output_key)},
```
vorlonofportland: networkd supports 'key' as shorthand for setting input-key and output-key to the same value; and the draft suggested we would support all three names.  Did you decide against this?

One possibility that comes to mind is to allow 'key' to take either a scalar (key: $mykey) or a dict (key: input: $inputkey output: $outputkey).

src/networkd.c
```diff
@@ -448,6 +509,16 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
...
     }
 
+    if (def->has_tunnels) {
+        /* iterate over all netdefs to find child tunnels */
+        GHashTableIter i;
+        net_definition* nd;
+        g_hash_table_iter_init(&i, netdefs);
+        while (g_hash_table_iter_next (&i, NULL, (gpointer*) &nd))
+            if (nd->tunnel.parent == def)
+                g_string_append_printf(s, "Tunnel=%s\n", nd->id);
```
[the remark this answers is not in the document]
cyphermox: There is still a concept of parentage in tunnels; however it *does not* have to be there. I'm still working out the specifics, some additional options may be needed when a tunnel doesn't have a parent, and maybe it's only some specific tunnel types that insist on it.

src/parse.c
cyphermox: WIP. Still does need to add key. I like the idea of it being a scalar or dict, if we can convince libyaml to figure it out (I think it's possible).

doc/netplan.md
```diff
@@ -678,6 +678,67 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+
+:   Defines the address of the remote endpoint of the tunnel.
+
+``key``  (scalar) or (mapping)
+:   See ``keys`` below.
+
+``keys`` (scalar) or (mapping)
+
+:   Define keys to use for the tunnel. The key can be a number or a dotted
+    quad (an IPv4 address). It is used for identification of IP transforms.
```
vorlonofportland: I think we should do one of two things here - either make this documentation much more extensive regarding the meaning of keys, which tunnel types accept them, and what they're used for; or (my preference) delegate the explanation of this to the relevant lower-level documentation (perhaps ip-tunnel(8)).  I think it's just confusing to say that a key can be "a number or an IP address" without explaining what this means.

As a point of interest, systemd.netdev(5) claims that key is only used for VTI/VTI6 tunnels, and ip-tunnel claims it's only used for GRE tunnels... you touch on this below, but maybe this means there are some upstream documentation bugs that need sorting.  (Regardless, we should certainly make sure the netplan documentation is correct for the netplan implementation.)

cyphermox: I don't understand what you're missing there, aside perhaps from "refer to 'ip-tunnel(8)' for more information". We do explain that they are used to identify IP transforms, and there are not many more ways to put it. Similarly, the expected values are just a number (could specify range, if it was clearer) or a dotted quad.

We're not dealing with documentation issues here: networkd and NetworkManager have different levels of implementation of handling of transform keys, and netplan needs to correctly capture this (I think it does).

doc/netplan.md
```diff
@@ -678,6 +678,67 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+``local`` (scalar)
+
+:   Defines the address of the local endpoint of the tunnel.
+
+``remote`` (scalar)
+
+:   Defines the address of the remote endpoint of the tunnel.
+
+``key``  (scalar) or (mapping)
+:   See ``keys`` below.
```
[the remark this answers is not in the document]
cyphermox: Either way around works. We do already support another typo due to its introduction in syntax previously.

I can rearrange doc, this makes no difference in the actual implementation; given that the same field can be both a scalar or a mapping, for ease of use and comprehension by the user, "key" and "keys" have been made equivalent. I want to avoid people being tripped because they used "keys", when they should have used "key" despite the use of a mapping, and "keys" when they only specify a single string.

src/parse.c
```diff
@@ -1596,9 +1875,12 @@ handle_network_type(yaml_document_t* doc, yaml_node_t* node, const void* data, G
             /* create new network definition */
             cur_netdef = g_new0(net_definition, 1);
             cur_netdef->type = GPOINTER_TO_UINT(data);
-            cur_netdef->backend = backend_cur_type ?: BACKEND_NONE;
+            cur_netdef->backend = backend_cur_type ?: get_default_backend_for_type(cur_netdef->type);
```
mwhudson: Does this mean the other invocation of get_default_backend_for_type is unnecessary now?

ssahani:
@cyphermox I have added support for ISATAP. https://github.com/systemd/systemd/commit/918049ad53fb5136403b0a03e454d29edbcb0363 . Just need to turn on the flag in sit tunnel.

doc/netplan.md
```diff
@@ -678,6 +678,64 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
 
 
+## Properties for device type ``tunnels:``
+
+``mode`` (scalar)
+
+:   Defines the tunnel mode. Valid options are ``sit``, ``gre``, ``ip6gre``,
+    ``ipip``, ``ipip6``, ``ip6ip6``, ``vti``, and ``vti6``. Additionally,
+    the ``networkd`` backend also supports ``gretap`` and ``ip6gretap`` modes.
+    In addition, the ``NetworkManager`` supports ``isatap`` tunnels.
```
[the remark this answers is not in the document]
cyphermox: Thanks!

src/parse.h
```diff
@@ -63,6 +64,46 @@ typedef enum {
...
+    TUNNEL_MODE_VTI6        = 9,
+
+    /* systemd-only, apparently? */
+    TUNNEL_MODE_GRETAP      = 101,
+    TUNNEL_MODE_IP6GRETAP   = 102,
+
+    _TUNNEL_MODE_MAX,
+} tunnel_mode;
+
+static const char* const
```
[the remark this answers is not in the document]
cyphermox: Keeping it as-is for convenience / since netdef_backend_to_name is similarly defined.

cyphermox:
Let's add isatap in networkd separately. It's just been added to systemd, so we won't have it in Ubuntu just yet anyway.

PR #60, opened 2018-11-28
tests: refactor tests to avoid a single huge generate.py

cyphermox:
## Description

Refactor generate.py tests to split it up per device type; to stop dealing with just a single huge unwieldy file.

## Checklist

- [n/a] \(Optional\) Closes an open bug in Launchpad.


Makefile
```diff
@@ -35,7 +35,6 @@ clean:
 	rm -rf test-coverage .coverage
 
 check: default linting
-	tests/generate.py
```
mwhudson: How are the tests run now? By the nose invocation below?

(PS: can we stop using nose? it's dead upstream afaik)

tests/generator/generate.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/python3
```
mwhudson: I don't think this file needs a shebang or is particularly well named. base.py maybe?

tests/generator/test_args.py
```diff
@@ -0,0 +1,174 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import sys
+import subprocess
+
+sys.path.append(os.path.dirname(__file__))
```
mwhudson: Ick. Doesn't "from .generate import .." work here?

tests/generator/test_args.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/python3
```
mwhudson: There is a case for making this file executable but as it is now it won't do anything. So i'd rather you removed the shebang (running just these tests can be done with the runner)

Makefile
cyphermox: Yup, using nose below. I wasn't aware it was dead, anything to discover the test suite will do. I'll adress that.

tests/generator/generate.py
cyphermox: Copy-pasta fallout. Thanks!

PR #61, opened 2018-12-04
Add support to override networkd UseMTU setting

shashankram:
Currently, there is no way to override networkd's
UseMTU flag via Netplan. This is useful to have
as a dhcp-override option similar to the other
dhcp-override options.
If the 'use-mtu' key is set to false in the netplan
config file, then set 'UseMTU' as false in networkd's
config.

Signed-off-by: Shashank Ram <shashank08@gmail.com>

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -329,6 +329,10 @@ combine_dhcp_overrides(net_definition* def, dhcp_overrides* combined_dhcp_overri
             g_fprintf(stderr, DHCP_OVERRIDES_ERROR, def->id, "use-hostname");
             exit(1);
         }
+        if (def->dhcp4_overrides.use_mtu != def->dhcp6_overrides.use_mtu) {
+            g_fprintf(stderr, DHCP_OVERRIDES_ERROR, def->id, "use-mtu");
+            exit(1);
+        }
         if (g_strcmp0(def->dhcp4_overrides.hostname, def->dhcp6_overrides.hostname) != 0) {
             g_fprintf(stderr, DHCP_OVERRIDES_ERROR, def->id, "hostname");
             exit(1);
@@ -462,11 +466,8 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
     }
 
     if (def->dhcp4 || def->dhcp6) {
-        /* isc-dhcp dhclient compatible UseMTU, networkd default is to
-         * not accept MTU, which breaks clouds */
-        g_string_append_printf(s, "\n[DHCP]\nUseMTU=true\n");
         /* NetworkManager compatible route metrics */
-        g_string_append_printf(s, "RouteMetric=%i\n", (def->type == ND_WIFI ? 600 : 100));
+        g_string_append_printf(s, "\n[DHCP]\nRouteMetric=%i\n", (def->type == ND_WIFI ? 600 : 100));
         if (g_strcmp0(def->dhcp_identifier, "duid") != 0)
             g_string_append_printf(s, "ClientIdentifier=%s\n", def->dhcp_identifier);
         if (def->critical)
@@ -474,6 +475,15 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
 
         dhcp_overrides combined_dhcp_overrides;
         combine_dhcp_overrides(def, &combined_dhcp_overrides);
+
+        /* Only set MTU from DHCP if use-mtu dhcp-override is not false. */
+        if (!combined_dhcp_overrides.use_mtu) {
+            /* isc-dhcp dhclient compatible UseMTU, networkd default is to
+             * not accept MTU, which breaks clouds */
+            g_string_append_printf(s, "UseMTU=false\n");
+        } else {
+            g_string_append_printf(s, "UseMTU=true\n");
+        }
 
         /* Only write DHCP options that differ from the networkd default. */
         if (!combined_dhcp_overrides.use_dns)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -347,6 +347,11 @@ that ``dhcp4-overrides`` and ``dhcp6-overrides`` contain the same keys and value
      :    Default: ``true``. When ``true``, the hostname received from the DHCP
           server will be set as the transient hostname of the system.
 
+     ``use-mtu`` (bool)
+     :    Default: ``true``. When ``true``, the MTU received from the DHCP
+          server will be set as the MTU of the network interface. When ``false``,
+          the MTU advertised by the DHCP server will be ignored.
+
      ``hostname`` (scalar)
      :    Use this value for the hostname which is sent to the DHCP server,
           instead of machine's hostname.
@@ -713,9 +718,9 @@ DHCP:
       ethernets:
         eno1:
           dhcp4: true
-          
+
 This is an example of a static-configured interface with multiple IPv4 addresses
-and multiple gateways with networkd, with equal route metric levels, and static 
+and multiple gateways with networkd, with equal route metric levels, and static
 DNS nameservers (Google DNS for this example):
 
     network:
@@ -736,7 +741,7 @@ DNS nameservers (Google DNS for this example):
             metric: 100
           - to: 0.0.0.0/0
             via: 11.0.0.1
-            metric: 100 
+            metric: 100
 
 This is a complex example which shows most available features:
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -1417,6 +1417,7 @@ const mapping_entry_handler nameservers_handlers[] = {
     {"use-ntp", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_ntp)},              \
     {"send-hostname", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.send_hostname)},  \
     {"use-hostname", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_hostname)},    \
+    {"use-mtu", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_mtu)},              \
     {"hostname", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(overrides.hostname)}
 
 const mapping_entry_handler dhcp4_overrides_handlers[] = {
@@ -1549,6 +1550,7 @@ initialize_dhcp_overrides(dhcp_overrides* overrides)
     overrides->use_ntp = TRUE;
     overrides->send_hostname = TRUE;
     overrides->use_hostname = TRUE;
+    overrides->use_mtu = TRUE;
     overrides->hostname = NULL;
 }
 
--- a/src/parse.h
+++ b/src/parse.h
@@ -81,6 +81,7 @@ typedef struct dhcp_overrides {
     gboolean use_ntp;
     gboolean send_hostname;
     gboolean use_hostname;
+    gboolean use_mtu;
     char* hostname;
 } dhcp_overrides;
 
--- a/tests/generator/base.py
+++ b/tests/generator/base.py
@@ -35,10 +35,13 @@ exe_generate = os.path.join(os.path.dirname(os.path.dirname(
 os.environ['G_DEBUG'] = 'fatal-criticals'
 
 # common patterns for expected output
-ND_DHCP4 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv4\nLinkLocalAddressing=ipv6\n\n[DHCP]\nUseMTU=true\nRouteMetric=100\n'
-ND_WIFI_DHCP4 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv4\nLinkLocalAddressing=ipv6\n\n[DHCP]\nUseMTU=true\nRouteMetric=600\n'
-ND_DHCP6 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv6\nLinkLocalAddressing=ipv6\n\n[DHCP]\nUseMTU=true\nRouteMetric=100\n'
-ND_DHCPYES = '[Match]\nName=%s\n\n[Network]\nDHCP=yes\nLinkLocalAddressing=ipv6\n\n[DHCP]\nUseMTU=true\nRouteMetric=100\n'
+ND_DHCP4 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv4\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=true\n'
+ND_DHCP4_NOMTU = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv4\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=false\n'
+ND_WIFI_DHCP4 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv4\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=600\nUseMTU=true\n'
+ND_DHCP6 = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv6\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=true\n'
+ND_DHCP6_NOMTU = '[Match]\nName=%s\n\n[Network]\nDHCP=ipv6\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=false\n'
+ND_DHCPYES = '[Match]\nName=%s\n\n[Network]\nDHCP=yes\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=true\n'
+ND_DHCPYES_NOMTU = '[Match]\nName=%s\n\n[Network]\nDHCP=yes\nLinkLocalAddressing=ipv6\n\n[DHCP]\nRouteMetric=100\nUseMTU=false\n'
 UDEV_MAC_RULE = 'SUBSYSTEM=="net", ACTION=="add", DRIVERS=="%s", ATTR{address}=="%s", NAME="%s"\n'
 UDEV_NO_MAC_RULE = 'SUBSYSTEM=="net", ACTION=="add", DRIVERS=="%s", NAME="%s"\n'
 
--- a/tests/generator/test_bonds.py
+++ b/tests/generator/test_bonds.py
@@ -46,8 +46,8 @@ IPv6AcceptRA=yes
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'bond0.netdev': '''[NetDev]
 Name=bond0
@@ -79,8 +79,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 '''})
         self.assert_nm(None, '''[keyfile]
 # devices managed by networkd
@@ -110,8 +110,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\n',
@@ -142,8 +142,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\n',
@@ -215,8 +215,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\n',
@@ -258,8 +258,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\n',
@@ -294,8 +294,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\nPrimarySlave=true\n',
@@ -332,8 +332,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
 RouteMetric=100
+UseMTU=true
 ''',
                               'eno1.network': '[Match]\nName=eno1\n\n'
                                               '[Network]\nLinkLocalAddressing=no\nBond=bn0\n',
--- a/tests/generator/test_bridges.py
+++ b/tests/generator/test_bridges.py
@@ -41,8 +41,8 @@ LinkLocalAddressing=ipv6
 ConfigureWithoutCarrier=yes
 
 [DHCP]
-UseMTU=true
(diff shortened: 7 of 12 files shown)


shashankram:
@cyphermox Could you take a look at this commit, thanks!

shashankram:
@ssahani Thanks for the reference. This commit is more about being able to control networkd's UseMTU setting from Netplan, similar to other dhcp-overrides such as DNS, NTP etc.
For example, a user could manually set the mtu using the "ip link set" command, and apply the netplan configuration. If dhcp happens to be used, then there needs to be a way for the dhcp client to ignore the MTU advertised by the server.

shashankram:
I guess your concern is around not using UseMTU and MTUBytes together? That is handled by networkd itself in https://github.com/systemd/systemd/pull/9832/commits/7169cdc8b971485751e4d92a1cae2d706293ecf9 . However, netplan needs a way to configure UseMTU setting as a dhcp-override in case a user manually sets the MTU and then requests a dhcp address using "netplan apply" while using networkd as the backend.

ssahani:
@shashankram thanks for the explanation . Not sure I follow you . 

>For example, a user could manually set the mtu using the "ip link set" command, and apply the netplan configuration

 ```ip link set``` could be done anytime. 

rather than this patch enables to to use DHCP MTU . I just gave a reference which is pretty imp to know that the U

shashankram:
@ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.

Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.

Now, if the user manually sets the MTU using "ip link set dev eth0 mtu <val>",  then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.

This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.

ssahani:
>@ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.

>Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.

hmm what if  ```ip link set mtu done afterwards ``` ?

shashankram:
> > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> 
> > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> 
> hmm what if `ip link set mtu done afterwards ` ?

Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied,  the MTU will be overwritten again with the one provided by the DHCP server.

With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.

ssahani:
> > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > 
> > 
> > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > 
> > 
> > hmm what if `ip link set mtu done afterwards ` ?
> 
> Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> 
> With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
isn't UseMTU= is false know ?

shashankram:
> > > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > > 
> > > 
> > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > 
> > > 
> > > hmm what if `ip link set mtu done afterwards ` ?
> > 
> > 
> > Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> > With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
> 
> isn't UseMTU= is false by default know ?

UseMTU is false by default in networkd backend, not in netplan. Netplan always hardcodes UseMTU to True if dhcp is used: https://github.com/CanonicalLtd/netplan/blob/master/src/networkd.c#L467
So whenever a netplan configuration is generated, it generates a backend networkd config with UseMTU=true.
Which is why I am proposing a dhcp-override option.

ssahani:
> > > > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > > > 
> > > > 
> > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > 
> > > > 
> > > > hmm what if `ip link set mtu done afterwards ` ?
> > > 
> > > 
> > > Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> > > With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
> > 
> > 
> > isn't UseMTU= is false by default know ?
> 
> UseMTU is false by default in networkd backend, not in netplan. Netplan always hardcodes UseMTU to True if dhcp is used: https://github.com/CanonicalLtd/netplan/blob/master/src/networkd.c#L467
> So whenever a netplan configuration is generated, it generates a backend networkd config with UseMTU=true.
> Which is why I am proposing a dhcp-override option.

ahh I didn't know that. No clue why that is done.  But we should also mention  if MTU is specified in .network then what is the behavior  that was my original reference.

shashankram:
> > > > > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > > > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > > > > 
> > > > > 
> > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > 
> > > > > 
> > > > > hmm what if `ip link set mtu done afterwards ` ?
> > > > 
> > > > 
> > > > Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> > > > With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
> > > 
> > > 
> > > isn't UseMTU= is false by default know ?
> > 
> > 
> > UseMTU is false by default in networkd backend, not in netplan. Netplan always hardcodes UseMTU to True if dhcp is used: https://github.com/CanonicalLtd/netplan/blob/master/src/networkd.c#L467
> > So whenever a netplan configuration is generated, it generates a backend networkd config with UseMTU=true.
> > Which is why I am proposing a dhcp-override option.
> 
> ahh I didn't know that. No clue why that is done. But we should also mention if MTU is specified in .network then what is the behavior that was my original reference.

In my exising Ubuntu 18.04 test envorinment, I don't see the UseMTU setting ignored when MTUBytes is set. 

```
# systemctl --version
systemd 237
```

That piece of information is already documented in networkd man.

ssahani:
> > > > > > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > > > > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > > > > > 
> > > > > > 
> > > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > > 
> > > > > > 
> > > > > > hmm what if `ip link set mtu done afterwards ` ?
> > > > > 
> > > > > 
> > > > > Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> > > > > With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
> > > > 
> > > > 
> > > > isn't UseMTU= is false by default know ?
> > > 
> > > 
> > > UseMTU is false by default in networkd backend, not in netplan. Netplan always hardcodes UseMTU to True if dhcp is used: https://github.com/CanonicalLtd/netplan/blob/master/src/networkd.c#L467
> > > So whenever a netplan configuration is generated, it generates a backend networkd config with UseMTU=true.
> > > Which is why I am proposing a dhcp-override option.
> > 
> > 
> > ahh I didn't know that. No clue why that is done. But we should also mention if MTU is specified in .network then what is the behavior that was my original reference.
> 
> In my exising Ubuntu 18.04 test envorinment, I don't see the UseMTU setting ignored when MTUBytes is set.
> 
> ```
> # systemctl --version
> systemd 237
> ```
> That piece of information is already documented in networkd man.

Well this will be the behavior in 240 onwards. I belive this is a very critical piece of information which we outsource documented. https://github.com/systemd/systemd/pull/9832/commits/7169cdc8b971485751e4d92a1cae2d706293ecf9

shashankram:
> > > > > > > > @ssahani Thanks for the reference, it seems like if MTUBytes is set in the networkd .link config file and if UseMTU is set in networkd .network file, then the MTUBytes setting takes precedence.
> > > > > > > > Now, consider a case where the MTUBytes is not set in networkd's .link file. This will happen if the netplan config file corresponding to an interface does not have the 'mtu' key. Running "netplan apply" will generate a networkd .network file with the "UseMTU" setting as True.
> > > > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > > > > This patch provides a mechanism to avoid that by exposing 'use-mtu' as a dhcp-override option in netplan.
> > > > > > > 
> > > > > > > 
> > > > > > > > Now, if the user manually sets the MTU using "ip link set dev eth0 mtu ", then the next time the netplan config is applied, if dhcp is used then the UseMTU setting which defaults to True will override the MTU manually configured by the user with the MTU value advertised by the server.
> > > > > > > 
> > > > > > > 
> > > > > > > hmm what if `ip link set mtu done afterwards ` ?
> > > > > > 
> > > > > > 
> > > > > > Based on the existing behavior upstream, the command will set the MTU to the value provided. But the next time netplan config is applied, the MTU will be overwritten again with the one provided by the DHCP server.
> > > > > > With this change, we are explicitly instructing networkd to not overwrite the the set MTU on the interface when a DHCP server advertises an MTU.
> > > > > 
> > > > > 
> > > > > isn't UseMTU= is false by default know ?
> > > > 
> > > > 
> > > > UseMTU is false by default in networkd backend, not in netplan. Netplan always hardcodes UseMTU to True if dhcp is used: https://github.com/CanonicalLtd/netplan/blob/master/src/networkd.c#L467
> > > > So whenever a netplan configuration is generated, it generates a backend networkd config with UseMTU=true.
> > > > Which is why I am proposing a dhcp-override option.
> > > 
> > > 
> > > ahh I didn't know that. No clue why that is done. But we should also mention if MTU is specified in .network then what is the behavior that was my original reference.
> > 
> > 
> > In my exising Ubuntu 18.04 test envorinment, I don't see the UseMTU setting ignored when MTUBytes is set.
> > ```
> > # systemctl --version
> > systemd 237
> > ```
> > That piece of information is already documented in networkd man.
> 
> Well this will be the behavior in 240 onwards. I belive this is a very critical piece of information which we outsource documented. [systemd/systemd@7169cdc](https://github.com/systemd/systemd/commit/7169cdc8b971485751e4d92a1cae2d706293ecf9)

I don't follow your comment. Are you suggesting we document the behavior of networkd's MTU setting again in Netplan? That seems to me independent of this change, which is more about providing a dhcp-override option for MTU.

ssahani:
> I don't follow your comment. Are you suggesting we document the behavior of networkd's MTU setting again in Netplan? That seems to me independent of this change, which is more about providing a dhcp-override option for MTU.

 it's a information we should all know what is the underlying behavior I gave a reference. what you think ?

shashankram:
> > I don't follow your comment. Are you suggesting we document the behavior of networkd's MTU setting again in Netplan? That seems to me independent of this change, which is more about providing a dhcp-override option for MTU.
> 
> it's a information we should all know what is the underlying behavior I gave a reference. what you think ?

Like I said already, that documentation applies even without my change. It can be a separate patch.

ssahani:
> > > I don't follow your comment. Are you suggesting we document the behavior of networkd's MTU setting again in Netplan? That seems to me independent of this change, which is more about providing a dhcp-override option for MTU.
> > 
> > 
> > it's a information we should all know what is the underlying behavior I gave a reference. what you think ?
> 
> Like I said already, that documentation applies even without my change. It can be a separate patch.

even I already said  what is behavior .

ssahani:
how does you  ```ip link set`` holds good

ssahani:
what is .link logic has does anything to do with this ?

shashankram:
Please read the netplan code and understand what this patch is trying to do. I already explained the intent of this patch multipe times.

shashankram:
> ahh I didn't know that. No clue why that is done. But we should also mention if MTU is specified in .network then what is the behavior that was my original reference.

This is already documented in networkd man. I don't think it needs to be documented again in netplan.

ssahani:
> Please read the netplan code and understand what this patch is trying to do. I already explained the intent of this patch multipe times.

Also you take the reference what I am trying to say .

shashankram:
> > Please read the netplan code and understand what this patch is trying to do. I already explained the intent of this patch multipe times.
> 
> Also you take the reference what I am trying to say .

I already looked at your reference and noted it in my previous comments. Let's wait for a review from the maintainers.

ssahani:
> > ahh I didn't know that. No clue why that is done. But we should also mention if MTU is specified in .network then what is the behavior that was my original reference.
> 
> This is already documented in networkd man. I don't think it needs to be documented again in netplan.

Read again what I am trying to say.

ssahani:
> Please read the netplan code and understand what this patch is trying to do. I already explained the intent of this patch multipe times.

what do you mean  ?  you don't understand what I am saying.

ssahani:
> > > Please read the netplan code and understand what this patch is trying to do. I already explained the intent of this patch multipe times.
> > 
> > 
> > Also you take the reference what I am trying to say .
> 
> I already looked at your reference and noted it in my previous comments. Let's wait for a review from the maintainers.

then what are you trying to convey ?

ssahani:
> Let me try to see again if I can understand what you are saying.
> You are suggesting to documetn in Netplan's man how networkd behaves respect to the UseMTU setting.
> 
> If this is not what you suggested, could you clarify again what is your concern here?

Let me try this again too . I gave a reference  what is the behavior  of upstream and we should note that. 
I failed to understand what is your concern ?

src/networkd.c
```diff
@@ -475,6 +476,15 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
...
 
+        /* Only set MTU from DHCP if use-mtu dhcp-override is not false. */
+        if (!combined_dhcp_overrides.use_mtu) {
+            /* isc-dhcp dhclient compatible UseMTU, networkd default is to
+             * not accept MTU, which breaks clouds */
+            g_string_append_printf(s, "\n[DHCP]\nUseMTU=false\n");
+        } else {
+            g_string_append_printf(s, "\n[DHCP]\nUseMTU=true\n");
+        }
+
```
cyphermox: Please keep [DHCP] title added earlier, and simply write out UseMTU=whatever. There are examples of how this has been done elsewhere. Otherwise some other values will be included in another section than the "DHCP" section.

Please also adjust the tests to make sure they still pass (will involve moving the UseMTU= line in expected outputs).

cyphermox:
The code changes generally look good to me, so this seems fine for merging, and while it does include a syntax change, it's a straightforward thing that I'm convinced we do want to include.

That said, I'm going to need to spend a bit of time to read the discussion; and tests appear to be broken by the change. Could you make sure to adjust the output generation in src/networkd.c and tests (see my inline comment)?

shashankram:
> The code changes generally look good to me, so this seems fine for merging, and while it does include a syntax change, it's a straightforward thing that I'm convinced we do want to include.
> 
> That said, I'm going to need to spend a bit of time to read the discussion; and tests appear to be broken by the change. Could you make sure to adjust the output generation in src/networkd.c and tests (see my inline comment)?

Thanks for the review. It seems to me the test is broken even without the change. I tried to run _make check-coverage_ without my change and it failed. Could you check if this is the case?

src/networkd.c
shashankram: Ah, good catch. Will fix this.

cyphermox:
Tests certainly do work; the same tests are run on the branch as those that are run on master (see the green checkmark after the commit message). It looks to me like the tests are failing for this branch simply because of the changes in `src/networkd.c` to move both "UseMTU" and the "[DHCP]" section header, and will need to be adjusted anyway since the fields won't be in the same order as before.

cyphermox:
@ssahani I get what you're saying:

`If <varname>MTUBytes=</varname> is set, then this setting is ignored.`

It's already broken in this way in netplan, since we do set MTUBytes= in some cases already. Let's address this in a separate patch adding a warning (I don't think we should error out, since that would break existing valid configs that have been in use).

ssahani:
@cyphermox  it's a story actually. To short there were a BZ when we try to set the MTU they rebooted. For example you can test with virtual box. 
1. we set MTU
2. Interface get rebooted.
3. interface gains carrier try to get DhCP with UseMTU
4. gets DHCP and try to set MTU . gets rebooted.
5. It's a loop.

Now let's take a consideration 100 machines doing the same the network flooded with DHCP requests.
and in a loop.

So we came up with solution 
1. chose static MTU
2. chose DHCP MTU

src/networkd.c
shashankram: Does the new diff look okday?

shashankram:
@cyphermox What needs to be fixed to address the failure in https://semaphoreci.com/cyphermox/netplan/branches/pull-request-61/builds/2 ?

cyphermox:
Yes, it's now https://semaphoreci.com/cyphermox/netplan/branches/pull-request-61/builds/3

If you run `make check-coverage` locally, you can see in test-coverage/C/ the output of lcov. You should find there that there are 3-4 lines in `src/networkd.c` that are not covered; so you might be missing a test there to make sure errors and caught and whatnot.

I'll try to set things up so it's easier for figure out in the future.

For now, you can also see the current coverage report for your branch on src/networkd.c here:
https://codecov.io/gh/CanonicalLtd/netplan/src/b1b2d53ff685db3f9aca099d23a6112285816abe/src/networkd.c

You'd get that URL from the output of the bash line for codecov runs: https://semaphoreci.com/cyphermox/netplan/branches/pull-request-61/builds/3

shashankram:
> Yes, it's now https://semaphoreci.com/cyphermox/netplan/branches/pull-request-61/builds/3
> 
> If you run `make check-coverage` locally, you can see in test-coverage/C/ the output of lcov. You should find there that there are 3-4 lines in `src/networkd.c` that are not covered; so you might be missing a test there to make sure errors and caught and whatnot.
> 
> I'll try to set things up so it's easier for figure out in the future.
> 
> For now, you can also see the current coverage report for your branch on src/networkd.c here:
> https://codecov.io/gh/CanonicalLtd/netplan/src/b1b2d53ff685db3f9aca099d23a6112285816abe/src/networkd.c
> 
> You'd get that URL from the output of the bash line for codecov runs: https://semaphoreci.com/cyphermox/netplan/branches/pull-request-61/builds/3

Thanks, it seems like I need to update the common base patterns in tests/generator/base.py. With this change, UseMTU=true is not a valid pattern for all inputs.

shashankram:
@cyphermox Mind taking a look at the latest diff? The tests have passed.

src/networkd.c
```diff
@@ -475,6 +476,15 @@ write_network_file(net_definition* def, const char* rootdir, const char* path)
         dhcp_overrides combined_dhcp_overrides;
         combine_dhcp_overrides(def, &combined_dhcp_overrides);
 
+        /* Only set MTU from DHCP if use-mtu dhcp-override is not false. */
+        if (!combined_dhcp_overrides.use_mtu) {
+            /* isc-dhcp dhclient compatible UseMTU, networkd default is to
+             * not accept MTU, which breaks clouds */
```
raharper: Do you have a bug reference for the "breaks clouds" comment?

shashankram: This is an existing comment, I retained it to keep the context on why UseMTU defaults to true.

shashankram:
Can I get a review for this again?

PR #63, opened 2018-12-07
Refactor integration tests to get rid of the single-file-of-doom

cyphermox:
## Description

I want integration tests to be easier to deal with, and we can also more easily navigate the scripts when debugging failing autopkgtests.

At this point, only 'scenarios.py' (if even) might be flaky; as it contains the two "complex scenarios" tests that have previously shown to be flaky.

## Checklist

- [n/a] Retains 100% code coverage (`make check-coverage`).
- [n/a] \(Optional\) Closes an open bug in Launchpad.


cyphermox:
Still missing doc for the various tests; I think they should include a proper docstring to describe the test being run.

PR #71, opened 2019-02-06
Initializing self.t to test before self.t.reset

bcavns01:
`self.t` is only defined within the `try` block.  If an exception is thrown and the try bails out, `self.t` is undefined.  This causes the `finally` block to throw its own exception when it tries to call `self.t.reset`, and netplan crashes.

## Checklist

### DIFF
--- a/netplan/cli/commands/try_command.py
+++ b/netplan/cli/commands/try_command.py
@@ -92,7 +92,6 @@ class NetplanTry(utils.NetplanCommand):
             print("\nReverting.")
             self.revert()
         finally:
-            self.t.reset(self.t_settings)
             self.cleanup()
 
     def backup(self):  # pragma: nocover (requires user input)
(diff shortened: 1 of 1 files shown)


cyphermox:
That's fine, but this doesn't seem like the right fix -- self.t.reset() is there to make sure the terminal is returned to the state it was before 'netplan try' started. In different shells, the terminal might behave differently. Have you tried all the permutations, say, what happens when you run 'netplan try' in csh, in bash, or in zsh? What about when you Control-C out, or hit a key, or let the timeout expire?

bcavns01:
Sorry, please forgive my haste.  I had initially tried to change as little as possible, but CTRL+C specifically does indeed show some undesirable behavior in bash.  I've updated the PR to instead initiatlize self.t and test it before trying to call `reset`.  This appears to handle all the cases well.

PR #75, opened 2019-02-24
allow *.yaml files as well as *.yml

Kriechi:
## Description
netplan currently only accepts *.yaml files.
However, it is quite common for YAML files to have a .yml file extension: https://en.wikipedia.org/wiki/YAML

A user with YAML knowledge, but no previous netplan experience, might simply create a `/etc/netplan/01-foobar.yml` config file, and then spend hours debugging why `netplan apply` doesn't pick up the changes described in this file.

This PR allows `*.yaml` and `*.yml` for all netplan config files in all three locations.
Since the filename (including extension) is different, the order & precedent behaviour does not change.

## Checklist

~~Retains 100% code coverage (`make check-coverage`).~~ (no, but I think thats because my test setup is not working properly...)~~
~~New/changed keys in YAML format are documented.~~ (no changes in this PR)
- [X] This PR will close https://bugs.launchpad.net/netplan/+bug/1815734

### DIFF
--- a/doc/netplan-generate.md
+++ b/doc/netplan-generate.md
@@ -48,9 +48,9 @@ For details of the configuration file format, see **netplan**(5).
 
 There are 3 locations that netplan generate considers:
 
- * /lib/netplan/*.yaml
- * /etc/netplan/*.yaml
- * /run/netplan/*.yaml
+ * /lib/netplan/*.yaml and /lib/netplan/*.yml
+ * /etc/netplan/*.yaml and /etc/netplan/*.yml
+ * /run/netplan/*.yaml and /run/netplan/*.yml
 
 If there are multiple files with exactly the same name, then only one
 will be read. A file in /run/netplan will shadow - completely replace
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -3,8 +3,8 @@ Distribution installers, cloud instantiation, image builds for particular
 devices, or any other way to deploy an operating system put its desired
 network configuration into YAML configuration file(s). During
 early boot, the netplan "network renderer" runs which reads
-``/{lib,etc,run}/netplan/*.yaml`` and writes configuration to ``/run`` to hand
-off control of devices to the specified networking daemon.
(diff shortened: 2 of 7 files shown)


vorlonofportland:
I don't agree that it is at all common to use .yml as an extention on Unix, I think this is exclusive to Windows with its history of three-letter extensions at the filesystem level.  I also do not think that it's less confusing to have multiple file extensions supported for configuration files.

If it is confusing to have `netplan apply` silently ignore files in the directory, let's address that instead by having netplan output debugging messages instead about ignored files.

Kriechi:
I think more debug messages about which config files are read & parsed, and which are ignored, would be a great benefit for the user!

Please let me at least give a few reasons why I think `.yml` IS a common file extension, even on Linux and other systems.

I'm aware that the official YAML documentation states we should yaml as extension:
https://yaml.org/faq.html

But let me give you a few examples from my own little bubble I'm working in:

Ansible, a well-known IT automation platform and configuration management tool uses mostly yml files, while also having yaml files:
```
➜  ansible git:(devel) find . -type f | grep ".yml" | wc -l
    2867
➜  ansible git:(devel) find . -type f | grep ".yaml" | wc -l
    2027
```

Applying the same to all of my open-source projects yields a 3:1 ratio in favour of `yml` over `yaml`, sampled from roughly 70 git repositories.

TravisCI and Appveyor, two popular CI/CD systems: use yml as their default file extension.
https://docs.travis-ci.com/user/customizing-the-build/#what-yaml-version-can-i-use-in-travisyml
https://www.appveyor.com/docs/appveyor-yml/

Codecov, the very coverage reporting tool that netplan uses, also uses yml in their documentation:
https://docs.codecov.io/docs/codecov-yaml#section-repository-yaml

Semaphore, the very build & CI system that netplan uses, also uses yml in their documentation:
https://semaphoreci.com/docs/adding-configuration-files.html#adding-a-new-configuration-file

Also, borrowing from https://github.com/interagent/prmd/pull/166 and https://en.wikipedia.org/wiki/Robustness_principle in general:

"Be conservative in what you do, be liberal in what you accept from others"

PR #80, opened 2019-03-15
Add suport for WPA Wired driver for 802.1x

fluffy:
## Description

To use 802.1x on a wired interface, there needs to be a way to specify to use the wired driver. This adds a new Boolean flag to support that. 

## Checklist

- [ x] Runs `make check` successfully.
- [ x] New/changed keys in YAML format are documented.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ install: default
 	install -m 644 doc/*.5 $(DESTDIR)/$(MANDIR)/man5/
 	install -m 644 doc/*.8 $(DESTDIR)/$(MANDIR)/man8/
 	install -D -m 644 src/netplan-wpa@.service $(DESTDIR)/$(SYSTEMD_UNIT_DIR)/netplan-wpa@.service
+	install -D -m 644 src/netplan-wpa-wired@.service $(DESTDIR)/$(SYSTEMD_UNIT_DIR)/netplan-wpa-wired@.service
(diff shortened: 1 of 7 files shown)


fluffy:
Bug for this at https://bugs.launchpad.net/netplan/+bug/1821182

fluffy:
Sure that makes sense as a better way of doing it.

fluffy:
So I assume the way to detect to use the wired driver is if " (def->type == ND_WIFI || def->has_auth)" is true - Is that the right way to do it ? 

I'm not really clear how the %I gets expanded to the interface in the  netplan-wpa@.service files. Any hints on how I might go about doing something simular to set the -D flag? 

Thanks, Cullen

cyphermox:
Well, sadly I think there would not be a way around writing the file ourselves rather than using a pre-made file like this. It should be file to write it on the fly anyway, since we already do so for the network devices.

jaymzh:
What's the status of this? Without this netplan can't do wired auth which is a shame.

jaymzh:
Oh, btw @fluffy - the %I is interpolated by systemd when the service is brought up as foo@interface

@cyphermox so when you say writing out a file you mean drop the symlinking part and just write a per-interface file? Thus dropping the %I as well?

sil2100:
Ok, I guess this has been resolved by the #109 pull request. Let's close this.

PR #83, opened 2019-04-10
Bionic Netplan Support Point-to-Point networks

khokooli:
hi 
please add point to point option for use instead of gateway 
i use this command : ip address add x.x.x.x peer z.z.z.z dev eth0


cyphermox:
Closing as rejected; this is unnecessary as all the changes are being backported already; they are available in the Proposed repositories; which need to be separately enabled. See https://wiki.ubuntu.com/Testing/EnableProposed

PR #84, opened 2019-04-11
nm: support route attributes "from", "onlink", and "table" and "match.interface-name"

thom311:
For NetworkManager, implement some routing attributes and matching interfaces by glob pattern.

On problem is, that certain of these features require a recent enough NetworkManager. Otherwise, NetworkManager will ignore the setting or fail to load it. I don't think that general problem is solved (nor do I see it solved for networkd).

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -440,6 +440,7 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
     g_autofree char* conf_path = NULL;
     mode_t orig_umask;
     char uuidstr[37];
+    const char *match_interface_name = NULL;
 
     if (def->type == NETPLAN_DEF_TYPE_WIFI)
         g_assert(ap);
@@ -475,12 +476,10 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
         else if (!def->has_match)
             g_string_append_printf(s, "interface-name=%s\n", def->id);
         else if (def->match.original_name) {
-            /* NM does not support interface name globbing */
-            if (strpbrk(def->match.original_name, "*[]?")) {
-                g_fprintf(stderr, "ERROR: %s: NetworkManager definitions do not support name globbing\n", def->id);
-                exit(1);
-            }
-            g_string_append_printf(s, "interface-name=%s\n", def->match.original_name);
+            if (strpbrk(def->match.original_name, "*[]?"))
+                match_interface_name = def->match.original_name;
+            else
+                g_string_append_printf(s, "interface-name=%s\n", def->match.original_name);
         }
         /* else matches on something other than the name, do not restrict interface-name */
     } else {
@@ -614,6 +613,11 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
 
     if (def->type == NETPLAN_DEF_TYPE_TUNNEL)
         write_tunnel_params(def, s);
+
+    if (match_interface_name) {
+        g_string_append(s, "\n[match]\n");
+        g_string_append_printf(s, "interface-name=%s;\n", match_interface_name);
+    }
 
     g_string_append(s, "\n[ipv4]\n");
 
--- a/src/nm.c
+++ b/src/nm.c
@@ -205,8 +205,7 @@ write_routes(const NetplanNetDefinition* def, GString *s, int family)
                 /* For IPv6 addresses, kernel and NetworkManager don't support a scope.
                  * For IPv4 addresses, NetworkManager determines the scope of addresses on its own
                  * ("link" for addresses without gateway, "global" for addresses with next-hop). */
-                g_fprintf(stderr, "ERROR: %s: NetworkManager does not support setting a scope for routes\n", def->id);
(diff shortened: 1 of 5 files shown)


slyon:
Hello Thomas,

please excuse my late reply to this PR. Actually, it looks really nice, as it fills some spots where the `NetworkManager` backend was lacking behind the `networkd` backend.

We just need to work on the tests and documentation, to adopt them accordingly and make sure it merges with current `master`.

thom311:
hi

Invalid route options get silently ignored while parsing keyfile:

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/801c895aa64edfcfd3365c4eebee6afa67bcc9b0/shared/nm-keyfile/nm-keyfile.c#L565

That is to ensure forward compatibility. Of course, it causes a different kind of problem, where users don't get notified of invalid settings.

slyon:
> Of course, it causes a different kind of problem, where users don't get notified of invalid settings.

Yes, we definitely have to note the relevant NM versions in our documentation for `onlink` then.

slyon:
Globbing of interface names seems to be supported since NM 1.14: https://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/NEWS?h=1.14.0#n11

slyon:
Looks like I cannot push to the PR branch directly... (even though it seems to be working via the web UI). I've been following this without any success: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork Maybe because it's not a 'user owned' repository...

I updated this PR to adopt tests and documentation: https://github.com/NetworkManager/netplan/pull/1

thom311:
Can I do something to help? Please ping me on freenode/IRC if yes.

slyon:
Integration tests OK on Ubuntu Groovy (NM-1.22)

src/nm.c
```diff
@@ -226,6 +217,21 @@ write_routes(const NetplanNetDefinition* def, GString *s, int family)
             if (cur_route->metric != NETPLAN_METRIC_UNSPEC)
                 g_string_append_printf(s, ",%d", cur_route->metric);
             g_string_append(s, "\n");
+
+            if (   cur_route->onlink
+                || cur_route->table != NETPLAN_ROUTE_TABLE_UNSPEC
+                || cur_route->from) {
+                g_string_append_printf(s, "route%d_options=", j);
```
sil2100, line 224: I don't know the exact syntax of route_options=, so just want to make sure: does it accept multiple key=value pairs separated with spaces? So, let's say, "route1_options=onlink=true from=192.168.14.2" would be valid syntax?

thom311: yes. In a slightly different syntax.


```
nmcli connection modify "$PROFILE" ipv4.routes \
     '192.168.14.0/24 onlink=1 src=192.168.14.2'
```
gives:

```
[ipv4]
route1=192.168.14.0/24
route1_options=onlink=true,src=192.168.14.2
```

 That's the only way to configure multiple route attributes for one route in keyfile format.

See https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/7b9cb1f952e8c8a0001144f457fcea1c7c853218/shared/nm-keyfile/nm-keyfile.c#L547

thom311: ah... then the patch is wrong :)

Sorry. Pushed a fixup commit...

thom311:
> You from->src change makes the unit tests fail.
> Could you please adopt the `test_route_reject_from` test case in `tests/generator/test_routing.py` accordingly?

done.

Please before merge squash the `fixup!` commits, or I can do it, if you prefer. I just leave them so it's easier to see the difference.

slyon:
Thanks, looking good now. Unit tests running OK, 100% coverage.

src/nm.c
sil2100: Ok, this looks good now, thanks! We could add a test-case for this I think, but I also don't consider it a blocker for getting this merged. Maybe something to follow up with later.

sil2100:
Note to self: as per our discussions with Lukas, we need to make sure that if for some reason we decide to do a full backport of latest netplan to bionic, we need to make sure to distro-patch the disabling some of the features that would cause NM to fail (I guess interface-name matching would one?).

slyon:
@sil2100 Yes, I was thinking about this as well. Fixed with 0c2babc648ad34d3a7fa70395393be37a48ab3e7

vorlonofportland:
On Fri, May 22, 2020 at 07:10:15AM -0700, Łukasz Zemczak wrote:

> Note to self: as per our discussions with Lukas, we need to make sure that
> if for some reason we decide to do a full backport of latest netplan to
> bionic, we need to make sure to distro-patch the disabling some of the
> features that would cause NM to fail (I guess interface-name matching
> would one?).

Are there tests for these features, such that we can rely on autopkgtests to
fail if trying to run against an incompatible version of NM?

slyon:
This is a good point, Steve!
And probably also the best way not to forget about our "note to self".
I'm working on it here: https://github.com/CanonicalLtd/netplan/pull/140

PR #85, opened 2019-04-23
Rework renderer parsing / netdef validation

cyphermox:
## Description

Rework renderer parsing and netdef validation to revert commit fe1fac85 and work around the resulting failures in validating tunnels (which have renderer-specific rules). Seems more logical to go about validating in a single pass anyway, once we're sure we do have all the required information.

This would fix https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1825206

## Checklist

### DIFF
--- a/src/validation.c
+++ b/src/validation.c
@@ -1,0 +1,223 @@
+/*
+ * Copyright (C) 2019 Canonical, Ltd.
+ * Author: Mathieu Trudel-Lapierre <mathieu.trudel-lapierre@canonical.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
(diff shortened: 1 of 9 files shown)


src/validation.c
```diff
@@ -0,0 +1,223 @@
...
+            if (!is_ip4_address(nd->tunnel.remote_ip))
+                return yaml_error(node, error, "%s: 'remote' must be a valid IPv4 address for this tunnel type", nd->id);
+            break;
+    }
+
+    return TRUE;
+}
+
+static gboolean
+validate_tunnel_backend_rules(net_definition* nd, yaml_node_t* node, GError** error)
```
mwhudson: I don't think node here can ever be non-NULL can it? I might be missing something.

vorlonofportland:
This looks sensible to me.

PR #86, opened 2019-05-11
Rescan network interfaces after `nmcli device disconnect` runs

Deltik:
When using the NetworkManager renderer, the logic in `NetplanApply.command_apply()` tells `nmcli` to disconnect the devices previously populated by `netifaces.interfaces()`.

In LP#1810043, interface "eth0:0" was deleted by NetworkManager, but other NetworkManager-managed devices such as "vlan557" or "br0" may also be deleted.

This means that the `devices` variable may be out-of-date with more devices than actually exist after `nmcli device disconnect` runs.

The attached patch fixes this bug by rescanning the devices after `nmcli device disconnect` so that the upcoming `udevadm test-builtin net_setup_link /sys/class/net/XXX` works with existing devices.

Fixes [Launchpad bug #1810043](https://bugs.launchpad.net/netplan/+bug/1810043)


YamiYukiSenpai:
Seems this patch isn't applied in Debian Buster

I'm using Netplan on Raspbian Buster, and I had to apply the patch manually.

ep4sh:
Still no changes in Debian Buster =(

Deltik:
@YamiYukiSenpai and @ep4sh: Comments about packaging `netplan.io` on Debian won't go anywhere on this thread.  If you want to bring this issue to the attention of the package's maintainers, you can find their contact information on this page: https://packages.debian.org/buster/netplan.io

PR #88, opened 2019-05-18
netplan info command (feature-flags spec)

cyphermox:
## Description

One important aspect is that this means there would need to be various updates to the existing code to add feature flags for existing content (where it makes sense to), but the jist of it is that I wanted to make it as easy as possible to add the feature flags without having to edit some other random file, to avoid forgetting about it. A comment in the source code seems appropriate:

If we always follow the `<comment chars> netplan-feature: <feature_flag_name> [comment_chars] format` then this is quite flexible, and allows adding new features as a new function, somewhere inside existing code, etc. to support all the kinds of things that might pop up: new device types, some new knob for an existing type, etc.

## Example output

Output from the command would look something like this (I added website on a whim, because there are relatively few places where people are currently pointed to it):

```
$ netplan info --json                                   
{"netplan.io": {"version": "2.98.1", "website": "http://netplan.io/", "features": ["syntax_error_context", "parser_error_context", "error_context_lineno"]}}

$ netplan info --yaml
netplan.io:
  features:
  - syntax_error_context
  - parser_error_context
  - error_context_lineno
  version: 2.98.1
  website: http://netplan.io/
```

```
method return time=1565793335.852693 sender=:1.979 -> destination=:1.977 serial=4 reply_serial=2
   array [
      struct {
         string "Version"
         variant             string "2.98.1"
      }
      struct {
         string "Features"
         variant             array [
               string "renderer_nm"
               string "renderer_networkd"
            ]
      }
   ]
```

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,10 +104,10 @@ install: default
 	install -D -m 644 src/netplan-wpa@.service $(DESTDIR)/$(SYSTEMD_UNIT_DIR)/netplan-wpa@.service
 	install -T -D -m 644 netplan.completions $(DESTDIR)/$(BASH_COMPLETIONS_DIR)/netplan
 	# dbus
-	mkdir -p $(DESTDIR)/share/dbus-1/system.d $(DESTDIR)/share/dbus-1/system-services
+	mkdir -p $(DESTDIR)$(PREFIX)/share/dbus-1/system.d $(DESTDIR)$(PREFIX)/share/dbus-1/system-services
 	install -m 755 netplan-dbus $(DESTDIR)/$(ROOTLIBEXECDIR)/netplan/
-	install -m 644 dbus/io.netplan.Netplan.conf $(DESTDIR)/share/dbus-1/system.d/
-	install -m 644 dbus/io.netplan.Netplan.service $(DESTDIR)/share/dbus-1/system-services/
+	install -m 644 dbus/io.netplan.Netplan.conf $(DESTDIR)$(PREFIX)/share/dbus-1/system.d/
+	install -m 644 dbus/io.netplan.Netplan.service $(DESTDIR)$(PREFIX)/share/dbus-1/system-services/
 
 %.service: %.service.in
 	sed -e "s#@ROOTLIBEXECDIR@#$(ROOTLIBEXECDIR)/#" $< > $@
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ src/_features.h: src/[^_]*.[hc]
 
 netplan/_features.py: src/[^_]*.[hc]
 	echo "# Generated file" > $@
-	echo "NETPLAN_VERSION = $(NETPLAN_VERSION)" >> $@
+	echo "NETPLAN_VERSION = \"$(NETPLAN_VERSION)\"" >> $@
 	echo "NETPLAN_FEATURE_FLAGS = [" >> $@
 	awk 'match ($$0, /netplan-feature:.*/ ) { $$0=substr($$0, RSTART, RLENGTH); print "    \""$$2"\"," }' $^ >> $@
 	echo "]" >> $@
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ SYSTEMD_GENERATOR_DIR=$(shell pkg-config --variable=systemdsystemgeneratordir sy
 SYSTEMD_UNIT_DIR=$(shell pkg-config --variable=systemdsystemunitdir systemd)
 BASH_COMPLETIONS_DIR=$(shell pkg-config --variable=completionsdir bash-completion || echo "/etc/bash_completion.d")
 
+GCOV ?= gcov
 ROOTPREFIX ?= /
 PREFIX ?= /usr
 ROOTLIBEXECDIR ?= $(ROOTPREFIX)/lib
@@ -79,7 +80,7 @@ check-coverage: coverage
 	python3-coverage report --omit=/usr* --show-missing --fail-under=100
 
 c-coverage:
-	lcov --directory . --capture -o generate.info
+	lcov --directory . --capture --gcov-tool=$(GCOV) -o generate.info
 	lcov --remove generate.info "/usr*" -o generate.info
 	genhtml -o test-coverage/C/ -t "generate test coverage" generate.info
 
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,21 @@ PYFLAKES3 ?= $(shell which pyflakes-3 || which pyflakes3 || echo true)
 PYCODESTYLE3 ?= $(shell which pycodestyle-3 || which pycodestyle || which pep8 || echo true)
 NOSETESTS3 ?= $(shell which nosetests-3 || which nosetests3 || echo true)
 
-default: generate doc/netplan.html doc/netplan.5 doc/netplan-generate.8 doc/netplan-apply.8 doc/netplan-try.8
+default: version_info.so generate doc/netplan.html doc/netplan.5 doc/netplan-generate.8 doc/netplan-apply.8 doc/netplan-try.8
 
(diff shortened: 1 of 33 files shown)


xnox:
Should be `httpS://`

netplan/cli/commands/info.py
```diff
@@ -0,0 +1,67 @@
...
+        super().__init__(command_id='info',
+                         description='Show current netplan version and available features',
+                         leaf=True)
+
+    def run(self):  # pragma: nocover (covered in autopkgtest)
+        format_group = self.parser.add_mutually_exclusive_group(required=False)
+        format_group.add_argument('--json', dest='version_format', action='store_const',
+                                  const='json',
+                                  help='Output version and features in JSON format')
+        format_group.add_argument('--yaml', dest='version_format', action='store_const',
```
raharper: We could just dump json;  yaml parsers can read json just fine.  Having both is fine as well.

netplan/cli/commands/info.py
```diff
@@ -0,0 +1,67 @@
...
+        if self.version_format is None:
+            self.version_format = 'yaml'
+
+        if self.version_format == 'json':
+            import json
+            print(json.dumps(netplan_version))
+
+        elif self.version_format == 'yaml':
+            import yaml
+            print(yaml.dump(netplan_version, default_flow_style=False))
```
raharper: Could we get a small indent? maybe indent=2 ?

Makefile
```diff
@@ -24,12 +27,21 @@ PYFLAKES3 ?= $(shell which pyflakes-3 || which pyflakes3 || echo true)
...
 generate: src/generate.[hc] src/parse.[hc] src/util.[hc] src/networkd.[hc] src/nm.[hc] src/validation.[hc] src/error.[hc]
 	$(CC) $(BUILDFLAGS) $(CFLAGS) -o $@ $(filter %.c, $^) `pkg-config --cflags --libs glib-2.0 gio-2.0 yaml-0.1 uuid`
 
+src/features.h: src/*.[hc]
+	echo "#include <stddef.h>\nstatic const char *feature_flags[] __attribute__((__unused__)) = {" > $@
+	awk 'match ($$0, /netplan-feature:[[:space:]]*.*/ ) { $$0=substr($$0, RSTART, RLENGTH); print "\""$$2"\"," }' $^ >> $@
+	echo "NULL, };" >> $@
+
+netplan/version_info.so: src/version_info.[hc] src/features.h
+	$(CC) -shared -fPIC $(BUILDFLAGS) $(CFLAGS) -o $@ $(filter %.c, $^) src/features.h `pkg-config --cflags --libs python3`
```
vorlonofportland: Why build this as a .so?  Given that all of the interesting source is an autogenerated .h file, it seems straightforward to also ship autogenerated .py instead of having to build and load a .so.

In fact, given that the src/features.h is *only* used in building the .so, I think this could be greatly simplified by outputting python source instead of a .h file.

src/error.c
```diff
@@ -157,7 +158,7 @@ yaml_error(const yaml_node_t* node, GError** error, const char* msg, ...)
     if (node != NULL) {
         error_context = get_syntax_error_context(node->start_mark.line, node->start_mark.column, error);
         g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_PARSE,
-                    "%s:%zu:%zu: Error in network definition: %s\n%s",
+                    "%s:%zu:%zu: Error in network definition: %s\n%s",  // netplan-feature: error_context_lineno
```
vorlonofportland: Should we require stricter parsing, and enforce the full comment string as '/* netplan-feature: <token> */'?  (the mix of C and C++ style comments here made me go double-check the parser)

src/networkd.c
```diff
@@ -27,6 +27,8 @@
 #include "parse.h"
 #include "util.h"
 
+/* netplan-feature: renderer_networkd */
```
vorlonofportland: I would recommend not exposing flags for features that have been present "forever" as this just makes the output more verbose without providing value.

There will never be a version of netplan that both supports the 'netplan info' command, and does not have support for a networkd renderer... unless you intend to make this future-proof in case later versions of netplan *drop* support for renderers?

Either way, I think any of these features that are present in the schema in 0.36.1 (the version in bionic release) don't need to be declared.

jay-vosburgh: Would it make sense to have feature flags for "forever" features that might not necessarily be available?  I'm thinking specifically of network manager; would it make sense to only provide a network-manager renderer feature flag if network manager is actually installed?

vorlonofportland:
On Tue, Jun 25, 2019 at 05:05:11PM -0700, Jay Vosburgh wrote:
> jay-vosburgh commented on this pull request.

> > @@ -27,6 +27,8 @@
>  #include "parse.h"
>  #include "util.h"

> +/* netplan-feature: renderer_networkd */

> Would it make sense to have feature flags for "forever" features that
> might not necessarily be available?  I'm thinking specifically of network
> manager; would it make sense to only provide a network-manager renderer
> feature flag if network manager is actually installed?

I'm unconvinced of this.  In general I think the kinds of things (e.g. 
cloud-init) that would care about this question would want to enforce the
presence of NetworkManager, rather than changing behavior in the absence of
NetworkManager.  Do you have a counterexample?

src/networkd.c
raharper: It certainly makes sense to me to emit which *renderers* the current netplan binary supports.  Cloud-init currently assumes that installing netplan implies that the target system which has netplan can support *all* renderers that netplan does.  Looking forward, they may not be reasonable unless we want netplan itself to depend on any renderer it supports configuring.  If a renderer backed were optional, then I think cloud-init would be interested in knowing if the target netplan would accept a particular yaml with backend specific bits.

raharper:
> I'm unconvinced of this. In general I think the kinds of things (e.g. cloud-init) that would care about this question would want to enforce the presence of NetworkManager, rather than changing behavior in the absence of NetworkManager. Do you have a counterexample?

cloud-init itself doesn't delve into netplan internal implementations (which backend is selected) in what yaml it emits.  If cloud-init were to do so, it would certainly want to query netplan for which backends it currently supports.

vorlonofportland:
On Thu, Jun 27, 2019 at 11:35:31AM -0700, Ryan Harper wrote:
> > I'm unconvinced of this.  In general I think the kinds of things (e.g. 
> > cloud-init) that would care about this question would want to enforce
> > the presence of NetworkManager, rather than changing behavior in the
> > absence of NetworkManager.  Do you have a counterexample?

> cloud-init itself doesn't delve into netplan internal implementations
> (which backend is selected) in what yaml it emits.  If cloud-init were to
> do so, it would certainly want to query netplan for which backends it
> currently supports.

Ok, but I cannot imagine any case in which cloud-init *would* want to make a
choice about the backend which is selected; I would expect it to always use
the default renderer, except in the case that cloud-init was fed in some
network configuration that specified NetworkManager.  And if cloud-init was
fed configuration that specified NetworkManager (for one or more network
interfaces), it would be very surprising to me if cloud-init would decide
after introspecting netplan that it should not honor this input merely
because NetworkManager is not currently available, and emit different
netplan config instead.

raharper:
> Ok, but I cannot imagine any case in which cloud-init *would* want to make a choice about the backend which is selected; I would expect it to always use the default renderer, except in the case that cloud-init was fed in some network configuration that specified NetworkManager. And if cloud-init was fed configuration that specified NetworkManager (for one or more network interfaces), it would be very surprising to me if cloud-init would decide after introspecting netplan that it should not honor this input merely because NetworkManager is not currently available, and emit different netplan config instead.

Actually I think we would definitely emit different netplan depending on what backends are available; in multiple cases, there are features that are only available to certain backends;  If the config sent to cloud-init include config for a specific feature that's only present in a particular backend, cloud-init may choose a different renderer.

We already have scenarios where the image includes both netplan and ifupdown;  things like administratively down interfaces can already be done in ifupdown, but not yet in netplan.  Cloud-init would query for the feature (which we should be able to tell if a particular backend has this feature, I don't think NetworkManager supports such a mode) and adjust how it renders the netplan.

PR #93, opened 2019-07-24
add dbus based "netplan apply" interface

mvo5:
## Description

This PR adds a new io.netplan.Netplan dbus activatable service
that can be used to Apply a netplan configuration.

The main use-case for this is Ubuntu Core where we want to allow
snaps to control the netplan network configuration via an interface.

However we do not want to give access to systemctl to restart
systemd-networkd or network-manager. The dbus service will run
outside of the confinement will archive this goal.

The interface is probably useful beyond Ubuntu Core.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 generate
+netplan-dbus
 test-coverage
 doc/*.html
 doc/*.[1-9]
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 BUILDFLAGS = \
 	-std=c99 \
 	-D_XOPEN_SOURCE=500 \
+	-DSBINDIR=\"$(SBINDIR)\" \
 	-Wall \
 	-Werror \
 	$(NULL)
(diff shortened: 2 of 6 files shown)


mvo5:
Hm, hm, it looks like somehow libsystem-dev needs to be added to the CI build system - any hints how I can do that?

cyphermox:
I've reworked the semaphoreci build rules to be closer to an actual Bionic environment (it's unfortunately based on either trusty or xenial...). For the meantime, that should suffice until we can use a different method of running the CI that doesn't have this kind of limitation.

There is, however, really an issue with 'capture_output'. That's something new in python 3.7. I think we should make sure this is all working correctly on 3.6 at least (what's in bionic). I'll merge this and then fixup for 3.6.

mvo5:
@cyphermox Thank you for updating the CI. I updated the tests to not use any of the modern subprocess things. It works now on 16.04 (and possibly older versions). 

About a different CI - can I help in any way? I could add a .travis.yml in a separate PR?

PR #96, opened 2019-08-02
netplan/apply: use dbus when inside a snap

anonymouse64:
## Description

Continuation of #93 to have the `netplan apply` command use the D-Bus activated service to actually run the privileged version of `netplan apply` outside of snap confinement. After this is merged, we will update the `network-setup-control` interface in snapd to allow calling this D-Bus service so that running `netplan apply` inside a strictly confined snap will result in `netplan apply` really being called outside of confinement. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ __pycache__
 *.pyc
 .coverage
 .vscode
+dbus/io.netplan.Netplan.service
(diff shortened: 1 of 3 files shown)


anonymouse64:
Actually, I just finished a snapd test and realize this needs a slight patch, closing while I finish that patch...

anonymouse64:
Okay, I fixed the issue. The problem was that I was hard-coding `/usr/bin/busctl`, which default core snap policy doesn't allow, so rather than require `/usr/bin/busctl` to be allowed I figured it was better to lookup `busctl` on $PATH so a snap could ship `systemd` as a stage-package and use `$SNAP/usr/bin/busctl` instead without changes to the snap policy. 

I also added `--system` to the list of arguments to `busctl` for clarity and added a new test to make sure that netplan apply uses busctl with expected arguments.

PR #97, opened 2019-08-03
Bonding mode validation

ipchama:
## Description
This changes adds checks for bond modes to prevent the acceptance of invalid modes or aliases.  For example, supplying `mode: 4` instead of `mode: 802.3ad`

Any existing code coverage that originally handled bonding covers this.

## Checklist

- [* ] Runs `make check` successfully.
- [?] Retains 100% code coverage (`make check-coverage`).
- [N/A] New/changed keys in YAML format are documented.
- [N/A] \(Optional\) Closes an open bug in Launchpad.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -725,14 +725,16 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
      ``up-delay`` (scalar)
      :    Specify the delay before enabling a link once the link is physically
           up. The default value is ``0``. This maps to the UpDelaySec= property
-          for the networkd renderer. If no time suffix is specified, the value
-          will be interpreted as milliseconds.
+          for the networkd renderer. This option is only valid for the miimon
+          link monitor. If no time suffix is specified, the value will be
+          interpreted as milliseconds.
 
      ``down-delay`` (scalar)
      :    Specify the delay before disabling a link once the link has been
           lost. The default value is ``0``. This maps to the DownDelaySec=
-          property for the networkd renderer. If no time suffix is specified,
-          the value will be interpreted as milliseconds.
+          property for the networkd renderer. This option is only valid for the
+          miimon link monitor. If no time suffix is specified, the value will
+          be interpreted as milliseconds.
 
      ``fail-over-mac-policy`` (scalar)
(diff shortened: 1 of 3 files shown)


ipchama:
As others have mentioned... seeing failures even on unchanged master branch.

cyphermox:
No, this looks like a valid test "failure". The coverage has dropped, and 3 lines are uncovered by tests in parse.c now. I'm looking at this, because codecov.io doesn't seem to highlight them for me.

ipchama:
Ok, all clear now.  Codecov and local make check/check-coverage weren't being helpful. :(

cyphermox:
Yeah, it's a little clunky to use. Michael made some extra changes very recently that should help with the local coverage reports (in test-coverage/ directory)

lukeyeager:
For future reference, this was a pretty brutal change. I upgraded 18 bionic servers. netplan v0.97 to v0.99 seemed like a pretty safe upgrade, especially since this was pushed to a LTS release of Ubuntu. After reboot, none of the servers were reachable at all (ping, ssh, etc.) because netplan failed to bring up the NICs after rebooting. I had to go in one-by-one through the BMC to edit a netplan file manually on each server. Please consider the consequences when pushing breaking changes like this to LTS releases. Yes, technically my config file was probably invalid, but I'd argue this was still a breaking change.

vorlonofportland:
@lukeyeager thanks for bringing this to our attention.  I agree that we should not be SRUing changes into stable releases which cause netplan to reject any config files that were previously considered valid.

When you say that you upgraded from netplan 0.97 to 0.99, does that mean you are upgrading to the version currently in bionic-proposed?

lukeyeager:
I misspoke. The upgrade was to 0.98, not 0.99.
`netplan.io:amd64 (0.97-0ubuntu1~18.04.1, 0.98-0ubuntu1~18.04.1)`
https://packages.ubuntu.com/bionic-updates/netplan.io

vorlonofportland:
@lukeyeager thanks, I've confirmed as well that this commit is present in 0.98.  I have opened https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1877643 for this issue.

PR #100, opened 2019-08-06
add support for IPv6MTUBytes

mwhudson:
## Description

Allow configuration of networkd's IPv6MTUBytes. network manager does not appear to allow this to be set except via router advisement.

## Checklist

### DIFF
--- a/tests/generator/test_common.py
+++ b/tests/generator/test_common.py
@@ -79,8 +79,9 @@ UseMTU=true
               version: 2
               ethernets:
                 eth1:
-                  mtu: 1280
+                  mtu: 9000
                   dhcp4: n
+                  ipv6-mtu: 2000
               bonds:
                 bond0:
                   interfaces:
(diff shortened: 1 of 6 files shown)


tests/generator/test_common.py
```diff
@@ -576,6 +578,17 @@ def test_mtu_all(self):
...
+    def test_ipv6_mtu(self):
+        self.generate(textwrap.dedent("""
+            network:
+              version: 2
+              renderer: NetworkManager
+              ethernets:
+                eth1:
+                  mtu: 9000
+                  parameters:
+                    ipv6-mtu: 2000"""), expect_fail=True)
```
vorlonofportland: This test case displays perfectly a concern that I have with creating a new "parameters" section, which is that this looks to me like a key that should be at the same level as mtu (and many other settings).  Why do we want this in a separate "parameters" sub-mapping?

cyphermox: MTU for IPv6 is a very different concept than the network-level MTU we already support. It's essentially a per-protocol MSS setting. I know this isn't perfect, but I'd really rather this be clearly separate in the schema such that it's obvious at a glance (because it's in a separate section) that it's advanced things you really shouldn't be playing with unless you know what you're doing. Having this segmented out means as a user you can think of the 10 or so knobs you typically need to work with, and see them easily in the doc. _ipv6-mtu_ is still easy to find in the doc; but now it takes just a tiny bit more of a conscious effort to include in the YAML, and stands out if we're to look at config, since it's likely something to cause issues.

Ultimately, I don't make the decision on this (you do, Steve), but this is consistent with what I suggested to Michael and with my vision for the schema.

mwhudson:
I've pushed changes so that ipv6-mtu is now alongside all the other stuff (much simpler!). PTAL.

PR #106, opened 2019-10-22
Allow parsing .yaml and .yml files

johschmitz:
## Description

Currently only .yaml files can be used, .yml is not allow. This is confusing for the user and causes a lot of unnecessary debugging time.

## Checklist

### DIFF
--- a/netplan/configmanager.py
+++ b/netplan/configmanager.py
@@ -95,7 +95,10 @@ class ConfigManager(object):
         # /run/netplan shadows /etc/netplan/, which shadows /lib/netplan
         names_to_paths = {}
         for yaml_dir in ['lib', 'etc', 'run']:
+            # Allow .yaml and .yml extension
             for yaml_file in glob.glob(os.path.join(self.prefix, yaml_dir, 'netplan', '*.yaml')):
+                names_to_paths[os.path.basename(yaml_file)] = yaml_file
+            for yaml_file in glob.glob(os.path.join(self.prefix, yaml_dir, 'netplan', '*.yml')):
                 names_to_paths[os.path.basename(yaml_file)] = yaml_file
 
         files = [names_to_paths[name] for name in sorted(names_to_paths.keys())]


johschmitz:
Wait @vorlonofportland  isn't YAML a standard that allows both .yaml and .yml? 
See "filename extensions" in the Wikipedia article: https://en.wikipedia.org/wiki/YAML
Do you honestly believe only partially supporting the standard is "good design"?

Let me give you my user story to make it more credible:
I was working on an Ansible playbook. Ansible apparently uses .yml more often judging by their documentation so I didn't want to have a mixture of .yml and .yaml in my little project. But then just nothing happend after Ansible coppied the netplan .yml to the netplan folder and called `netplan generate && netplan apply`. So I thought, what the hell is going on? After some time I figured could it be so that they really just parse .yaml files??
This is a typical case of Murphy's law. If you don't allow this or at least throw a warning many people will run into this problem and get frustrated.

johschmitz:
Here is somebody else who stumbled over this:
https://stackoverflow.com/a/57772502

vorlonofportland:
> Wait @vorlonofportland isn't YAML a standard that allows both .yaml and .yml?
> See "filename extensions" in the Wikipedia article: https://en.wikipedia.org/wiki/YAML

A wikipedia article is not a standard.

I don't see that yaml.org defines file extensions as part of the standard.

> This is a typical case of Murphy's law. If you don't allow this or at least throw a warning many people will run into this problem and get frustrated.

The rules for file names are documented in netplan(5) and on https://netplan.io.  Is there somewhere else you think it should have been documented so that users know the rules?

I would be supportive of a patch to netplan that would throw an error about the presence of any files in the netplan directories which are neither a) files ending in .yaml that it will read as configuration, nor b) files with known editor or system backup extensions.

johschmitz:
> I would be supportive of a patch to netplan that would throw an error about the presence of any files in the netplan directories which are neither a) files ending in .yaml that it will read as configuration, nor b) files with known editor or system backup extensions.

I think that would be good enough to help users not run into the trap. Although there could be unwanted side effects to this approach I could imagine. Personally I have no time now to dive deep enough into your code to write such a patch though. I believe that I have at least lived up to my responsibility towards my fellow human beings by coming up with one suggestion/pull request which I think is the correct approach.

Regardless of that, I find it a bit extreme in general to not support something that is out there in the wild already like a million times, just check the use of .yml in countless Docker and Ansible examples. But that is your project's decision and I am going to respect it. You must go ahead and do whatever you think is best for your users.

johschmitz:
I was just working with docker-compose and was using _ instead of - so I saw a nice example of how they handle it:

        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml

Kriechi:
ref https://bugs.launchpad.net/netplan/+bug/1815734

neodon:
This just burned a chunk of my time.  At least now I know where we are regarding YAML file extensions (seems like a tabs vs spaces thing).  Still would have been nice to get some kind of warning instead of netplan processing 0 files and providing no useful information, even in debug mode.

Also thank you all for everything you do.  I always feel bad complaining about stuff like this, but I hope it's useful to you.  Awesome work.

PR #107, opened 2019-10-23
Allow `critical` to be used without `dhcp4`/`dhcp6` enabled

cedws:
## Description
This change allows the `critical` option to be used without `dhcp4` or `dhcp6` also being enabled. My understanding is that the underlying networkd option `CriticalConnection` also can prevent static IPs being dropped, not just ones assigned via DHCP. In my case, I needed to prevent IPs being assigned by keepalived from being dropped.

## Checklist

- [*] Runs `make check` successfully.
- [*] Retains 100% code coverage (`make check-coverage`).
- [*] New/changed keys in YAML format are documented.

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -198,7 +198,7 @@ Virtual devices
 
 :   (networkd backend only) Designate the connection as "critical to the
     system", meaning that special care will be taken by systemd-networkd to
-    not release the IP from DHCP when the daemon is restarted.
+    not release the assigned IP when the daemon is restarted.
 
 ``dhcp-identifier`` (scalar)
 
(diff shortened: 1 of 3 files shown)


sil2100:
Could you also merge latest trunk so that this is mergeable again?
Thanks!

cedws:
Thanks for taking a look. I saw that there's already a test for this option, so I just modified it to not required `dhcp4` or `dhcp6` anymore. I couldn't actually figure out how to run the generator tests, but hopefully that should work. I've also rebased against master.

I might be able to help out with adding support for `KeepConfiguration` when the time comes.

cedws:
Great, thanks again.

PR #108, opened 2019-11-14
Add `phase2` keyword to `auth` section

jaymzh:
## Description

For corporate wired 802.1x networks, a common config is PEAP plus an
MSCHAPv2 phase2. This is supported by wpa_supplicant, and NM, but not
netplan. This adds support for that.

Signed-off-by: Phil Dibowitz <phil@ipom.com>

## Checklist

- [X] Runs `make check` successfully.
- [X] Retains 100% code coverage (`make check-coverage`).
- [X] New/changed keys in YAML format are documented.
- [x] \(Optional\) Closes an open bug in Launchpad.

PR #109, opened 2019-11-14
Handled wired wpa_supplicant driver automatically

jaymzh:
Currently netplan can't handle authentication for wired connections
because it doesn't force the loading of the wired driver.

This is based on @fluffy's work in #80 but re-working it to drop
the existing pre-written unitfiles and instead to generate our own,
as well as automatically determine when to load the wired driver.

**NOTE**: This is *NOT* yet tested, I want to make sure this is what @cyphermox is looking for.

## Checklist


jaymzh:
I'm a little rusty on C and have not used glib and friends very much so free to let me know where my codestyle is off...

src/networkd.c
```diff
@@ -725,6 +725,24 @@ append_wpa_auth_conf(GString* s, const authentication_settings* auth)
...
 }
 
+static void
+write_wpa_unit(net_definition* def, const char* rootdir)
+{
+    GString* s = g_string_new("[Unit]\n");
+    g_autofree char* path = g_strjoin(NULL, "/run/systemd/system/netplan-wpa-", def->id, ".service", NULL);
+    g_string_append_printf(s, "Description=WPA supplicant for netplan %s\n", def->id);
+    g_string_append(s, "DefaultDependencies=no\n");
+    g_string_append_printf(s, "After=sys-subsystem-net-devices-%s.device\n", def->id);
```
jaymzh: it's worth noting here that we're not getting systemd's escaping here, and what we probably want is to shell out to `systemd-escape` to escape `def->id` as that's what these units will be named.

jaymzh:
Since @cyphermox tagged @vorlonofportland on #108, I'll go ahead and tag them on this one too. :)

cyphermox:
I think this generally looks pretty good; but I think it should keep using netplan-wpa@ form, since we'll really have multiple instances of wpa for each interface; this would be proper parlance for systemd jobs.

jaymzh:
@cyphermox you can't do that, because the files are, necessarily, different per interface, so having one Unit file for all interfaces won't work... that was the whole point of generating them in the first place. :)

jaymzh:
Also, while you're at it, @cyphermox - #108 should be super straighforward.

jaymzh:
rebased and tests updated. @vorlonofportland lmk if it needs anything else.

src/networkd.c
cyphermox: Yes, let's make sure the names are escaped. I don't think we're likely to run into the issue often, but there might be the odd device that is handled special, like VLANs.

jaymzh: Oh shit, forgot about that. Will do.

jaymzh:
OK I had to fork off to `system-escape` since the API isn't exposed (there's an RFE for it https://github.com/systemd/systemd/issues/14202). But it works. The coverage test say there's only 99% of lines covered now, and I have **no** clue why any help would be appreciated.

PR #113, opened 2019-12-04
Add wireguard support, take two.

lxnt:
## Description

Adds wireguard support.

Rework of https://github.com/CanonicalLtd/netplan/pull/82

Changes:
 - l2tp and dummy support split off
 - multiple peer support
 - full code coverage

## Checklist


lxnt:
We were talking about reusing 'remote' key for endpoint in wireguard.

This is now moot since wireguard supports multiple peers per interface, each possibly having different endpoints.

Niek:
What is the status of this being merged? It seems there is no way to get a wireguard client up&running with netplan at the moment - I get messages from networkd-dispatcher similar to ``ERROR:Unknown interface index 7 seen even after reload``. Is there a known workaround?

lxnt:
Guys, any ideas on this being merged?

Any changes needed?

src/parse.c
```diff
@@ -1648,6 +1648,184 @@ handle_tunnel_key_mapping(yaml_document_t* doc, yaml_node_t* node, const void* _
...
+        assert_type(value, h->type);
+
+        if (!h->handler(doc, value, (void *)peer + GPOINTER_TO_UINT(h->data), error))
+            return FALSE;
+    }
+    return TRUE;
+}
+
+static gboolean
+handle_wireguard_peers(yaml_document_t* doc, yaml_node_t* node, const void* _, GError** error)
```
slyon, line 1806: I'd prefer that the Wireguard peers would be handled the same way as `NetplanWifiAccessPoint`/`handle_wifi_acccess_points`, using a global `cur_wireguard_peer` variable.

This way we could avoid storing the whole `NetplanWireGuardPeer`/`wireguard_peer` struct data inside the GArray and store just the pointers to the corresponding data instead.

src/parse.c
```diff
@@ -1648,6 +1648,184 @@ handle_tunnel_key_mapping(yaml_document_t* doc, yaml_node_t* node, const void* _
...
+{
+    cur_netdef->wireguard_peers = g_array_new(FALSE, TRUE, sizeof(wireguard_peer));
```
slyon, line 1808: This code might be run multiple times, as some YAML configuration require multiple passes to parse all netplan definitions. Please make sure that we're not overwriting the array each time, e.g.:
```
if (!cur_netdef->wireguard_peers)
    cur_netdef->wireguard_peers = g_array_new(FALSE, TRUE, sizeof(NetplanWireGuardPeer*));
```

This is also applies to the `peer.allowed_ips` array.

src/parse.c
```diff
@@ -1648,6 +1648,184 @@ handle_tunnel_key_mapping(yaml_document_t* doc, yaml_node_t* node, const void* _
...
+        g_autofree char* addr = NULL;
+        yaml_node_t *entry = yaml_document_get_node(doc, *i);
+        assert_type(entry, YAML_MAPPING_NODE);
+
+        wireguard_peer peer = {0};
+        peer.allowed_ips = g_array_new(FALSE, FALSE, sizeof(char*));
+
+        gboolean ret = process_wireguard_peer(doc, entry, &peer, error);
+        if (ret) {
+            g_array_append_val(cur_netdef->wireguard_peers, peer);
```
slyon, line 1819: Please make sure not to insert the same peer multiple times, if the code is run through a 2nd pass. Similar to how I just recently fixed the same issue for IP4/IP6 addresses: https://github.com/CanonicalLtd/netplan/pull/147/commits/2f2971b7935c9ff6ae14cd68907cfb87b68472c4

src/validation.c
```diff
@@ -57,6 +57,13 @@ is_ip6_address(const char* address)
     return FALSE;
 }
 
+gboolean
+is_hostname(const char *hostname)
+{
+    static const gchar *pattern = "^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])$";
```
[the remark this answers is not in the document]
slyon: Nevermind. The double `\\` is needed to escape within the C string itself.

doc/netplan.md
```diff
@@ -1007,6 +1008,51 @@ Examples:
...
+    ``public-key`` (scalar) – since **0.100**
+    :   Peer's base64-encoded public key, required.
+
+    ``preshared-key`` (scalar) – since **0.100**
+    :   Optional preshared key for a peer.
+
+    ``preshared-key-file`` (scalar) – since **0.100**
+    :   Path to a file with an optional preshared key for a peer.
+
+    ``endpoint`` (scalar) – since **0.100**
```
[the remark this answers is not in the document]
slyon: I renamed it to `remote`, while keeping `endpoint` as an alias, see: 444cb93

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
+    ``remote`` (scalar) – since **0.100**
+    :   Remote endpoint IPv4/IPv6 address or a hostname, followed by a colon
+        and a port number.
+
+    ``allowed-ips`` (sequence of scalars) – since **0.100**
+    :    A list of IP (v4 or v6) addresses with CIDR masks from which this peer
+         is allowed to send incoming traffic and to which outgoing traffic for
+         this peer is directed. The catch-all 0.0.0.0/0 may be specified for
+         matching all IPv4 addresses, and ::/0 may be specified for matching
+         all IPv6 addresses.
```
vorlonofportland, line 1058: I wonder if this should actually be expressed as 'routes' on the tunnel, possibly also reusing the 'use-routes: false' syntax from the dhcp*-overrides.  Does the backend implement this via managing the routing table?

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
 
+Wireguard-specific keys:
+
+    ``private-key`` (scalar) – since **0.100**
+    :   Base64-encoded private key, prefixed with ``base64:``. Or an absolute
+        path to a file with the private key.
+
+    ``fwmark`` (scalar) – since **0.100**
+    :   Firewall mark for outgoing WireGuard packets from this interface,
+        optional.
```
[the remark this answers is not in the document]
slyon: Yes, it should! Fixed.

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
+
+    ``port`` (scalar) – since **0.100**
+    :   UDP port to listen at or ``auto``. Optional, defaults to ``auto``.
+
+    ``peers`` (sequence of mappings) – since **0.100**
+    :   A list of peers, each having keys documented below.
```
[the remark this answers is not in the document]
slyon: From what I understand, Wireguard has pretty similar server & client definitions. A server interface/tunnel can have multiple peers (i.e. clients), while a client interface/tunnel can have one peer (the server).

doc/netplan.md
slyon: Indeed.. there is quite some overlap of the `routes` and `peers` stanza. For Wireguard the backends do **not** implement it via routing tables, tho. There are explicit settings for both backends, to set "AllowedIPs" and "Endpoints".

Lets see if I can implement `routes` for Wireguard tunnels instead of introducing this `peers` setting.

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
 
 :   Alternate name for the ``key`` field. See above.
 
+Wireguard-specific keys:
+
+    ``private-key`` (scalar) – since **0.100**
+    :   Base64-encoded private key, prefixed with ``base64:``. Or an absolute
+        path to a file with the private key.
```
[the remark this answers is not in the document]
slyon: Fixed. We can make use of `input_key` here as the private key in wireguard is basically the key which is used to decrypt the incoming packages.

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
+                  allowed-ips: [0.0.0.0/0, "2001:fe:ad:de:ad:be:ef:1/24"]
+                  keepalive: 23
+                  remote: 1.2.3.4:5
+                - public-key: /other/public.key
+                  allowed-ips: [0.0.0.0/0, "10.10.10.20/24"]
+                  keepalive: 22
+                  remote: 5.4.3.2:1
+
+    ``public-key`` (scalar) – since **0.100**
+    :   Peer's base64-encoded public key, required.
```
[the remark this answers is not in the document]
slyon: There are 2 levels of keys for wireguard:
* **private key**: defined directly for the tunnel interface (e.g. `wg0`) – basically the "input-key" decrypting incoming packages
* **public & shared keys**: defined for each individual peer (remote endpoint) of that tunnel interface

I've rebuilt the wireguard key handling so that on the tunnel level one can specify the `key:` (private key) as scalar (or use `keys.input-key` respectively), while on the wireguard peers level one can use the `key:` scalar to define the public key (always required) or optionally define the `keys.public:` & `keys.shared` mapping if an optional shared key shall be defined:

Example1:
```
    tunnels:
      wg0:
        mode: wireguard
        addresses: [...]
        key: mNb7OIIXTdgW4khM7OFlzJ+UPs7lmcWHV7xjPgakMkQ=
        peers:
          - keys:
              public: rlbInAj0qV69CysWPQY7KEBnKxpYCpaWqOs/dLevdWc=
              shared: 7voRZ/ojfXgfPOlswo3Lpma1RJq7qijIEEUEMShQFV8=
            ...
```

Example2:
```
    tunnels:
        wg0:
            mode: wireguard
            keys:
                input-key: /some/private.key
            peers:
                - key: rlbInAj0qV69CysWPQY7KEBnKxpYCpaWqOs/dLevdWc=
                  allowed-ips: [0.0.0.0/0, "2001:fe:ad:de:ad:be:ef:1/24"]
                  keepalive: 23
                  remote: 1.2.3.4:5
                - keys:
                      public: M9nt4YujIOmNrRmpIRTmYSfMdrpvE7u6WkG8FY8WjG4=
                      shared: /some/shared.key
                  allowed-ips: [10.10.10.20/24]
                  keepalive: 22
                  remote: 5.4.3.2:1
```

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
+
+    ``shared-key`` (scalar) – since **0.100**
+    :   Optional base64-encoded pre shared key for a peer, prefixed with
+        ``base64:``. Or an absolute path to a file with a pre shared key.
```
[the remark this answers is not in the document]
slyon: I reworked the wireguard key handling. See my comment about `private-key` above.

doc/netplan.md
```diff
@@ -1007,6 +1008,60 @@ Examples:
...
+
+    ``remote`` (scalar) – since **0.100**
+    :   Remote endpoint IPv4/IPv6 address or a hostname, followed by a colon
+        and a port number.
```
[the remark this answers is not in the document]
slyon: I removed all of the special wireguard `peers:` stanza and merged it with the `routes:` stanza instaed.
`remote` (endpoint) is now defined by the `via` field (i.e. the gateway).
`allowed-ips` is now defined by the `to` field (i.e. networks where traffic shall be routed to)

New keys include:
* routes.keys.public
* routes.keys.shared
* routes.key (optional short form of routes.keys.public)
* routes.keepalive
* routes.mark
* routes.port

doc/netplan.md
slyon: `peers:` is now merged with `routes:`, see comment above.

slyon:
Hey @vorlonofportland I did quite a big refactoring today, in order to get rid of the `peers:` stanza and merge it with the `routes:` stanza instead (adding a few extra keys there); see inline comments:

I would be glad if you could have another look!

src/parse.c
```diff
@@ -1776,15 +1840,15 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"routes", YAML_SEQUENCE_NODE, handle_routes},                                            \
     {"routing-policy", YAML_SEQUENCE_NODE, handle_ip_rules}
 
-#define COMMON_BACKEND_HANDLERS							\
+#define COMMON_BACKEND_HANDLERS							                                      \
```
sil2100, line 1843: Completely minor: I guess the whitespace changes here are not needed?

src/networkd.c
```diff
@@ -151,6 +152,72 @@ write_tunnel_params(GString* s, const NetplanNetDefinition* def)
...
+            g_string_append_printf(peer_s, "PersistentKeepalive=%d\n", peer->keepalive);
+        if (peer->via)
+            g_string_append_printf(peer_s, "Endpoint=%s\n", peer->via);
+        /* The key was already validated via validate_tunnel_grammar(), but we need
+         * to differentiate between base64 key VS absolute path key-file. And a base64
+         * string could (theoretically) start with '/', so we use is_wireguard_key()
+         * as well to check for more specific characteristics (if needed). */
+        if (peer->preshared_key) {
+            if (peer->preshared_key[0] == '/' && !is_wireguard_key(peer->preshared_key))
+                g_string_append_printf(peer_s, "PresharedKeyFile=%s\n", peer->preshared_key);
```
sil2100, line 211: Is the `PresharedKeyFile` similarly only available in systemd 242+ ? Thinking about it now, maybe we should include some info in the documentation about when, with which dependency versions, the absolute keyfile paths can be used for it to work.

src/nm.c
```diff
@@ -327,6 +333,63 @@ write_bridge_params(const NetplanNetDefinition* def, GString *s)
...
+            g_string_append_printf(s, "persistent-keepalive=%d\n", peer->keepalive);
+        if (peer->via)
+            g_string_append_printf(s, "endpoint=%s\n", peer->via);
+        /* The key was already validated via validate_tunnel_grammar(), but we need
+         * to differentiate between base64 key VS absolute path key-file. And a base64
+         * string could (theoretically) start with '/', so we use is_wireguard_key()
+         * as well to check for more specific characteristics (if needed). */
+        if (peer->preshared_key) {
+            if (peer->preshared_key[0] == '/' && !is_wireguard_key(peer->preshared_key)) {
+                g_fprintf(stderr, "%s: shared key needs to be base64 encoded when using the NM backend\n", def->id);
```
sil2100, line 375: Similarly here. Maybe we should include information in the docs that using a keyfile path is only supported for the networkd backend?

src/parse.c
slyon: Definitely not needed, but I liked to clean it up on the way when I stumbled upon it.

src/networkd.c
slyon: +1 I updated the docs.

src/nm.c
slyon: +1 I updated the docs.

src/parse.c
```diff
@@ -1185,29 +1229,73 @@ handle_routes_type(yaml_document_t* doc, yaml_node_t* node, const void* data, GE
...
+handle_routes_ip(yaml_document_t* doc, yaml_node_t* node, const void* data, GError** error)
+{
+    /* If a sequence node is given, it defines the allowed_ips for a wireguard route/peer */
+    if (node->type == YAML_SEQUENCE_NODE)
+        return handle_generic_addresses(doc, node, FALSE, &(cur_route->allowed_ips), &(cur_route->allowed_ips), error);
+    else if (node->type == YAML_SCALAR_NODE) {
+        /* if this is IP4/IP6 address in 'addr' or 'addr/prefixlen' format */
+        if (is_ip4_address(scalar(node)) || is_ip6_address(scalar(node)) || strrchr(scalar(node), '/')) {
+            guint offset = GPOINTER_TO_UINT(data);
+            int family = get_ip_family(scalar(node));
```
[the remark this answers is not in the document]
slyon: Good catch! I can't remember why I removed that check... So I added it back in, incl. a test case.

slyon:
Thanks a lot @sil2100 ! Your review was much appreciated for this big changeset. I resolved the comments you had inline.

When @vorlonofportland signs-off the final schema, I think this should be ready to merge!

doc/netplan.md
```diff
@@ -556,7 +564,7 @@ These options are available for all types of interfaces.
 
      ``type`` (scalar)
      :    The type of route. Valid options are "unicast" (default),
-          "unreachable", "blackhole" or "prohibit".
+          "unreachable", "blackhole",  "prohibit" or "wireguard".
```
vorlonofportland, line 567: So a wireguard interface may have both 'unicast' routes that specify actual routing tables; and 'wireguard' routes that map to peers?

This seems rather confusing to me; I'm inclined to take back my earlier suggestion here, in favor of using a separate 'peer' notation.

doc/netplan.md
```diff
@@ -572,6 +580,31 @@ These options are available for all types of interfaces.
...
+
+         ``public`` (scalar) – since **0.100**
+         :    A base64-encoded public key, requried for Wireguard peers.
+
+         ``shared`` (scalar) – since **0.100**
+         :    A base64-encoded preshared key. Optional for Wireguard peers.
+              When the ``systemd-networkd`` backend (v242+) is used, this can
+              also be an absolute path to a file containing the preshared key.
+
+     ``key`` (scalar) – since **0.100**
```
vorlonofportland, line 603: I would prefer not to see alternate schema options for expressing the same content.

slyon: Okay. I removed `key` setting in favor of just `keys` within wireguard-peers.
We will still have the tunnel `key[s]` for historical reason (I don't want to break the backwards compatibility there).

slyon:
Thanks @vorlonofportland for your comments. I removed the `key` setting in favor of just `keys` within wireguard-peers.

I changed back to the original schema for wireguard peers, which now looks like this:
```
    tunnels:
        wg0:
            mode: wireguard
	    key: mNb7OIIXTdgW4khM7OFlzJ+UPs7lmcWHV7xjPgakMkQ=
            peers:
                - keys:
                      public: rlbInAj0qV69CysWPQY7KEBnKxpYCpaWqOs/dLevdWc=
                  allowed-ips: [0.0.0.0/0, "2001:fe:addead:be:ef:1/24"]
                  keepalive: 23
                  endpoint: 1.2.3.4:5
                - keys:
                      public: M9nt4YujIOmNrRmpIRTmYSfMdrpvE7u6WkG8FY8WjG4=
                      shared: /some/shared.key
                  allowed-ips: [10.10.10.20/24]
                  keepalive: 22
                  endpoint: 5.4.3.2:1
````

doc/netplan.md
slyon: Done. I switched back to the original `peers:` schema.

slyon:
Thank you very much Alexander for your initial work on this Wireguard support and Lukasz and Steve for your thorough reviews.
This was a pretty big one, but I think we nailed it now.

I did some more cleanup and verified that all unit- & integration-tests PASS.

Ready for merging! :tada:

lxnt:
My pleasure.

PR #115, opened 2019-12-04
Add dummy interfaces

lxnt:
## Description

Split from #82 

This is more like a place to discuss how to put dummy interfaces in netplan rather than a straight PR.
Besides, I'm not sure with what exactly to bring code coverage back to 100%
Let's maybe work on this? I'm using this code to force bridge intefaces up when they don't have any other interfaces added on configuration time. This is a real use case.

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -883,6 +883,11 @@ Examples:
 :    netplan ID of the underlying device definition on which this VLAN gets
      created.
 
+## Properties for device type ``dummies:``
+
+    None.
+
+
 Example:
 
     ethernets:
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -280,6 +280,10 @@ write_netdev_file(net_definition* def, const char* rootdir, const char* path)
             g_string_append_printf(s, "Kind=vlan\n\n[VLAN]\nId=%u\n", def->vlan_id);
             break;
 
+        case ND_DUMMY:
+            g_string_append_printf(s, "Kind=dummy\n");
+            break;
+
         case ND_TUNNEL:
             switch(def->tunnel.mode) {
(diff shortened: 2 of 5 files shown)


cyphermox:
But why is this required? I thought we'd already fixed bringing up VLANs and bridges with no members, as this requires setting "ConfigureWithoutCarrier=" and other keys at the systemd level?

I'm not fundamentally opposed to dummy devices, but I wonder if they wouldn't better fit elsewhere (like under ethernet, simply) since dummies are indeed a thing, but once you move into network hardware, you'll much more likely want to create loopback devices, which is something else that is not currently straightforward.

lxnt:
Hm.
This was a straightforward split from the #82 request, since the configuration I currently have does rely on them, but only for forcing bridges up. 

I must have missed the bit with bridges being fixed already. I'll recheck and if it works and if no one other use case comes up this can of course be dropped.

Frankkkkk:
Hi @cyphermox ,
In our company, we use dummy interfaces to store the IPs we want to announce via BGP. The workaround is to set them on the `lo` interface but this is not scalable as we only have one loopback per machine.

Our `frr` router running on the hosts announces all the ips on the `bgp` dummy interface and all the IPs with a higher metric that are on the `bgp-failover` dummy interface. They effectively act as sets but with the advantage of being understood by the kernel.

This seems to me to be a good use case. Don't hesitate if you need more info.
Cheers

attila123:
Any updates please?

slyon:
> Any updates please?

This is a feature request, it has not been prioritized so far: https://bugs.launchpad.net/netplan/+bug/1774203

PRs are welcome (this PR would need some updates, to be ready for discussions).

slyon:
Closing this in favor of #361

PR #118, opened 2020-03-17
Fix quotation of WPA PSK hex passthrough for networkd

pdulapalli:
## Description

This changeset provides the ability to supply a WPA PSK compatible hex string in the `netplan` YAML configuration, which then `networkd` can use for the `wpa_supplicant` configuration. Essentially, this ensures that a hex string, indicated by the prefix `hex:`, will be passed through unquoted to retain `wpa_supplicant` configuration compatibility

This directly addresses the `netplan` ticket here: https://bugs.launchpad.net/netplan/+bug/1867690

## Checklist

### DIFF
--- a/examples/wireless_wpa_psk.yaml
+++ b/examples/wireless_wpa_psk.yaml
@@ -1,0 +1,14 @@
+network:
+  version: 2
+  renderer: networkd
+  wifis:
+    wlp2s0b1:
+      dhcp4: no
+      dhcp6: no
+      addresses: [192.168.0.21/24]
+      gateway4: 192.168.0.1
+      nameservers:
+        addresses: [192.168.0.1, 8.8.8.8]
+      access-points:
+        "network_ssid_name":
+          password: "hex:****************************************************************"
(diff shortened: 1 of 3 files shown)


sil2100:
I have filled in PR #120 with the other approach that I mentioned. Could you take a look and see if that would work for you? Thanks!

pdulapalli:
> Hello Anurag! First of all, thank you for reporting the issue and preparing a proposed solution.
> 
> I don't think this is the right to go though. The psk can contain various printable characters, so adding special meaning to the "hex:" string, potentially, can cause regressions for users with psk's starting with hex:. We do something similar for hash:, but that is a different case as we then still copy-over the whole passphrase string without any further modification.
> 

Hi Łukasz, really appreciate you taking a look at this pull request! 😄  
Looking at your approach, that was actually my plan A, and then I realized the edge case when someone has a 64 character ASCII password, and so went to plan B above. 

I believe that your approach addresses that edge case smoothly, so looks good to me!

sil2100:
Thank you for your review! I have merged the other branch, so let's close this one for clarity!

PR #120, opened 2020-03-19
Fix quotation of WPA PSK 64 hex-digit keys for networkd 

sil2100:
## Description

Per definition seen on wpa_supplicant's documentation:
"psk: WPA preshared key; 256-bit pre-shared key
The key used in WPA-PSK mode can be entered either as 64 hex-digits, i.e., 32 bytes or as an ASCII passphrase (in which case, the real PSK will be generated using the passphrase and SSID). ASCII passphrase must be between 8 and 63 characters (inclusive)."

Whenever we see a 64 hex-digit passphrase, let's make sure it's passed unquoted. Since normal passwords need to be between 8 and 63 characters long, we also add some additional checks to be sure those conditions are fulfiled.
While at it, I changed most wifi test passwords as we were using exemplary passphrases that were shorter than 8 characters, which is not a good thing to demonstrate anyway.

This hopefully fixes: https://bugs.launchpad.net/netplan/+bug/1867690

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -724,7 +724,7 @@ append_wpa_auth_conf(GString* s, const authentication_settings* auth, const char
                     }
                 /* this is required to be unquoted */
                 g_string_append_printf(s, "  psk=%s\n", auth->password);
-            } else if (len < 8 || len > 64) {
+            } else if (len < 8 || len > 63) {
                 /* per wpa_supplicant spec, passphrase needs to be between 8
                    and 63 characters */
                 g_fprintf(stderr, "ERROR: %s: ASCII passphrase must be between 8 and 63 characters (inclusive)\n", id);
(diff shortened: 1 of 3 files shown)


src/networkd.c
```diff
@@ -708,7 +710,24 @@ append_wpa_auth_conf(GString* s, const authentication_settings* auth)
...
+            if (len == 64) {
+                /* must be a hex-digit key representation */
+                for (unsigned i = 0; i < 64; ++i)
+                    if (!isxdigit(auth->password[i])) {
+                        g_fprintf(stderr, "ERROR: %s: PSK length of 64 is only supported for hex-digit representation\n", id);
+                        exit(1);
+                    }
+                /* this is required to be unquoted */
+                g_string_append_printf(s, "  psk=%s\n", auth->password);
+            } else if (len < 8 || len > 64) {
```
slyon: Here we should check for `len < 8 || len > 63`, even though we already check the ` len == 64` case above, passphrase are still only valid for 8-63 characters.

PR #121, opened 2020-03-20
Adopt integration tests for NetworkManager v1.22 (focal)

slyon:
## Description
NetworkManager changed the "startup complete" behavior in v1.22, this is  why in some cases the test runs, even though the network is not yet fully up and running (i.e. only IPv6, not IPv4 & IPv6).

see: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/1e5206414af83ede1e4e1ff6f525161b83519845

This is why we use libnm and pygobject to check for the n-m internal NM_ACTIVATION_STATE_FLAG_IP4_READY flag via d-bus, instead of just waiting for the `nm-online` binary ("startup complete" state).

In order to make the tests run, we **need the following dependencies** inside the Debian package (`debian/tests/control`), for the "ethernets" test:
    +  libnm0,
    +  python3-gi,
    +  gir1.2-nm-1.0,

## Checklist

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -362,7 +362,7 @@ class IntegrationTestsBase(unittest.TestCase):
                                                 stderr=subprocess.PIPE, universal_newlines=True)
                 self.fail('timed out waiting for networkd to settle down:\n%s\n%s\n%s' % (st, st_e, st_e2))
 
-        if subprocess.call(['nm-online', '--quiet', '--timeout=120', '--wait-for-startup']) != 0:
+        if subprocess.call(['nm-online', '--quiet', '--timeout=240', '--wait-for-startup']) != 0:
             self.fail('timed out waiting for NetworkManager to settle down')
 
     def nm_wait_connected(self, iface, timeout):
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -28,6 +28,7 @@ import subprocess
 import tempfile
 import unittest
 import shutil
+import gi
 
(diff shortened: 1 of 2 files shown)


slyon:
How can we best integrate the package/autopkgtest dependencies on Launchpad?

https://code.launchpad.net/~ubuntu-core-dev/netplan/+git/ubuntu

tests/integration/base.py
```diff
@@ -362,9 +363,25 @@ def generate_and_settle(self):
...
 
+    def nm_online_full(self, iface, timeout=60):
+        '''Wait for NetworkManager connection to be completed (incl. IP4 & DHCP)'''
+
+        gi.require_version('NM', '1.0')
+        from gi.repository import NM
+        for t in range(timeout):
+            c = NM.Client.new(None)
+            con = c.get_device_by_iface(iface).get_active_connection()
+            if not con: self.fail('no active connection for %s by NM' % iface)
```
sil2100: Just a stylistic nit-pick: could we get this split into two lines? Since then it will fit better with the rest of the code.

slyon:
Thanks for the review! Merged.
I'll commit the packaging changes to the `ubuntu/master` branch of the packaging repo then.

slyon:
I don't have access to the upstream packaging repo, yet. (https://code.launchpad.net/~ubuntu-core-dev/netplan/+git/ubuntu)

So I've pushed the changes here: https://github.com/slyon/netplan/commits/packaging-focal

PR #124, opened 2020-03-31
Rename gsms to modems and enable CDMA

slyon:
## Description
Rename `gsms` to `modems` to handle GSM and CDMA connections in the same way.
GSM connections are auto detected by checking for specific config options, which are only applicable to GSM connections.

Also, added the `number` field, which was missing to enable cmda (and legacy gsm/ppp) connections. And the `modem-mtu` field as required by the spec.

## Checklist


doc/netplan.md
```diff
@@ -567,9 +567,15 @@ not support GSM modems.
 :    Specify the device ID (as given by the WWAN management service) of the modem to match.
      This can be found using ``mmcli``.
 
+``modem-mtu`` (scalar)
+:    If non-zero, only transmit packets of the specified size or smaller, breaking larger packets up into multiple frames.
+
```
vorlonofportland: why would this be modem-mtu, instead of using the 'mtu' key which is already supported for all interface types? (your latest commit says "according to the spec" - what spec?)

src/nm.c
```diff
@@ -70,17 +70,37 @@ g_string_append_netdef_match(GString* s, const NetplanNetDefinition* def)
     }
 }
 
+/**
+ * Infere if this is a modem netdef of type GSM.
```
vorlonofportland: "Infer"

slyon:
Thank you for your valuable feedback and also the stylistic nitpicking, this is what will improve the maintainability in the long run!

@vorlonofportland I was referring to the "NetworkManager readwrite plugin" spec, which requires the ability to set a `mtu` field for GSM and/or CDMA connections. But you are right, that we should be using the generic `mtu` field for the modem interface, instead of introducing a new `modem-mtu` field!

@sil2100 I've deleted the `modem-mtu` field and use the generic `mtu` field now. Though, I had to make sure not to inject an `[ethernet]` or `[802-3-ethernet]` section within the `[gsm/cdma]` config, which is auto appended for the common link handlers. I'm not sure if I handled this correctly and would appreciate your feedback on this!

Also, I'm not sure about the docs formatting no-op.

PR #125, opened 2020-04-02
Implement WiFi flags for bssid/band/channel

slyon:
## Description
This PR implements the following WiFi fields, required by the "NetworkManager readwrite plugin" spec, for the `nm` and `networkd` backends. systemd-networkd does not support _seen-bssids_.

- band
- bssid
- channel
- seen-bssids

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -638,6 +638,10 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           numbers overlap between bands, this property takes effect only if
           the ``band`` property is set, too.
 
+     ``seen-bssids`` (sequence of scalars)
+     :    A list of BSSIDs (each BSSID formatted as a MAC address) that have
+          been detected as part of the Wi-Fi network.
+
 ## Properties for device type ``bridges:``
 
 ``interfaces`` (sequence of scalars)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -624,6 +624,20 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           and ``adhoc`` (peer to peer networks without a central access point).
           ``ap`` is only supported with NetworkManager.
 
+     ``bssid`` (scalar)
+     :    If specified, directs the device to only associate with the given
+          access point.
+
+     ``band`` (scalar)
+     :    Possible bands are ``a`` (for 5GHz 802.11a) and ``bg``
+          (for 2.4GHz 802.11), do not restrict the 802.11 frequency band of the
+          network if unset (the default).
+
+     ``channel`` (scalar)
+     :    Wireless channel to use for the Wi-Fi connection. Because channel
+          numbers overlap between bands, this property takes effect only if
+          the ``band`` property is set, too.
+
 ## Properties for device type ``bridges:``
 
 ``interfaces`` (sequence of scalars)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -638,12 +638,6 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           numbers overlap between bands, this property takes effect only if
           the ``band`` property is also set.
 
-     ``seen-bssids`` (sequence of scalars)
-     :    A list of BSSIDs (each BSSID formatted as a MAC address) that have
-          been detected as part of the Wi-Fi network. This property is only
-          meant for reading and reflects an internal state of NetworkManager.
-          The changes you make to this property will not be preserved.
-
 ## Properties for device type ``bridges:``
 
 ``interfaces`` (sequence of scalars)
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -640,8 +640,9 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
 
      ``seen-bssids`` (sequence of scalars)
      :    A list of BSSIDs (each BSSID formatted as a MAC address) that have
-          been detected as part of the Wi-Fi network. This is only supported
-          with NetworkManager.
+          been detected as part of the Wi-Fi network. This is a read-only
+          option, used only to track NetworkManager's state when using the
+          netplan read-write plugin.
 
 ## Properties for device type ``bridges:``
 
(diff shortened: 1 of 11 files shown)


slyon:
I adopted NetworkManager's language for those fields, not sure if we should have this confirmed by Steve?

sil2100:
Just a heads up: I started reviewing this PR but I'd like to sit down a bit longer on it. Same for the rest of the PRs.

src/networkd.c
```diff
@@ -809,6 +818,41 @@ write_wpa_conf(const NetplanNetDefinition* def, const char* rootdir)
...
+                    if (ap->channel) {
+                        g_string_append_printf(s, "  freq_list=%d\n", wifi_get_freq24(ap->channel));
+                    } else {
+                        g_string_append_printf(s, "  freq_list=");
+                        g_hash_table_foreach(wifi_frequency_24, wifi_append_freq, s);
+                        // overwrite last whitespace with newline
+                        s = g_string_overwrite(s, s->len-1, "\n");
+                    }
+                }
+                if (ap->band == NETPLAN_WIFI_BAND_5) {
```
sil2100, line 838: I suppose this would be better of becoming an 'else if'.
I was also thinking of maybe making this less boilerplate, as we basically have the same code (with different variables) executed twice, but then again it would become less readable. So maybe let's leave it as is for now.

doc/netplan.md
```diff
@@ -624,6 +624,25 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+     :    Possible bands are ``a`` (for 5GHz 802.11a) and ``bg``
+          (for 2.4GHz 802.11), do not restrict the 802.11 frequency band of the
+          network if unset (the default).
+
+     ``channel`` (scalar)
+     :    Wireless channel to use for the Wi-Fi connection. Because channel
+          numbers overlap between bands, this property takes effect only if
+          the ``band`` property is set, too.
+
+     ``seen-bssids`` (sequence of scalars)
```
sil2100, line 641: I'd like to ask Steve for an opinion. So I wanted to include this key because of the NM read-write plugin, so that when the seen-bssids state changes, users can see that also in the netplan configuration and not only via NM. Of course that was the idea, but I'd like to hear from Steve if maybe it's not really needed?
Anyway, for now the one thing I'd recommend either way is to not include it in the documentation, or explicitly mentioning that it's a read-only variable that is used only to track NM state when using the read-write plugin.

src/util.c
```diff
@@ -86,3 +86,63 @@ unlink_glob(const char* rootdir, const char* _glob)
...
+{
+    int channels[] = { 7, 8, 9, 11, 12, 16, 32, 34, 36, 38, 40, 42, 44, 46, 48,
+                       50, 52, 54, 56, 58, 60, 62, 64, 68, 96, 100, 102, 104,
+                       106, 108, 110, 112, 114, 116, 118, 120, 122, 124, 126,
+                       128, 132, 134, 136, 138, 140, 142, 144, 149, 151, 153,
+                       155, 157, 159, 161, 165, 169, 173 };
+    gboolean found = FALSE;
+    for (unsigned i = 0; i < sizeof(channels) / sizeof(int); i++) {
+        if (channel == channels[i])
+            found = TRUE;
```
sil2100, line 130: Should we maybe 'break' after we find the first match, not to iterate over the whole list when not needed?

slyon: Absolutely!

doc/netplan.md
slyon: Done. Stated it explicitly. If we do not include it, our test suite will an error about missing/non-documented fields.

src/networkd.c
slyon: Right. And the surrounding "if(ap->band)" can be removed, too.

doc/netplan.md
```diff
@@ -624,6 +624,26 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+     :    A list of BSSIDs (each BSSID formatted as a MAC address) that have
+          been detected as part of the Wi-Fi network. This is a read-only
+          option, used only to track NetworkManager's state when using the
```
vorlonofportland, line 644: I'm not sure what it means in this context for it to be "read-only".  It's clearly not read only from the perspective of the NM read-write plugin, else we would not need to store it.  If the meaning is that we will not emit it to NetworkManager via the netplan renderer, then I would say /that/ explicitly.

doc/netplan.md
```diff
@@ -624,6 +624,26 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
           ``ap`` is only supported with NetworkManager.
 
+     ``bssid`` (scalar)
+     :    If specified, directs the device to only associate with the given
+          access point.
+
+     ``band`` (scalar)
+     :    Possible bands are ``a`` (for 5GHz 802.11a) and ``bg``
+          (for 2.4GHz 802.11), do not restrict the 802.11 frequency band of the
+          network if unset (the default).
```
vorlonofportland, line 634: 'a' and 'bg' may be the native terms used in NetworkManager, but I don't think this is a very clear way to label the bands.  I think it's more common to refer to them as 2.4G and 5G, and that these are the terms we should use here explicitly - either 2.4G / 5G or 2.4GHz / 5GHz.  'a' and 'bg' is just going to confuse users who are configuring 802.11ac or 802.11n.

doc/netplan.md
```diff
@@ -624,6 +624,26 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
...
+
+     ``channel`` (scalar)
+     :    Wireless channel to use for the Wi-Fi connection. Because channel
+          numbers overlap between bands, this property takes effect only if
+          the ``band`` property is set, too.
```
vorlonofportland, line 639: --> if the ``band`` property is also set.

doc/netplan.md
slyon: I opted for 5GHz/2.4GHz as the documented default, also accepting 5G/2.4G as a fallback.

This is because I want to avoid confusion with future mobile communication protocol standards (3G/4G/5G).

doc/netplan.md
slyon: Actually, the `wifi.seen-bssids` setting is deprecated. NetworkManager will not read this option or use it internally. Instead, it reads this data from it's internal state db at:
`/var/lib/NetworkManager/seen-bssids`
    
This is why we do not need to support this field for netplan.io. I added a commit to remove the corresponding code.

doc/netplan.md
slyon: Fixed.

PR #126, opened 2020-04-02
Add IPv6 addr-gen-mode field for NM backend

slyon:
## Description
Allow to choose the configure method for creating the address for use with RFC4862 IPv6 Stateless Address Autoconfiguration. Possible values are: `eui64` or `stable-privacy`. If unset, the backend's default value stays in place.

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -239,7 +239,7 @@ Virtual devices
 
     Example: ``addresses: [192.168.14.2/24, "2001:1::1/64"]``
 
-``addr-gen-mode`` (scalar)
+``ipv6-address-generation`` (scalar)
 
 :   Configure method for creating the address for use with RFC4862 IPv6
     Stateless Address Autoconfiguration. Possible values are ``eui64``
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -239,6 +239,12 @@ Virtual devices
 
     Example: ``addresses: [192.168.14.2/24, "2001:1::1/64"]``
 
+``addr-gen-mode`` (scalar)
+
+:   Configure method for creating the address for use with RFC4862 IPv6
+    Stateless Address Autoconfiguration. Possible values are ``eui64``
+    or ``stable-privacy``.
+
 ``gateway4``, ``gateway6`` (scalar)
 
 :   Set default gateway for IPv4/6, for manual address configuration. This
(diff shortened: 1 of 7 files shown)


slyon:
I am not yet sure how to properly implement this for the `systemd-networkd` backend...

networkd's `IPv6Token=` might be an option [0], but this seems to be a combination with NM's `token` field, c.f.: https://bugs.launchpad.net/ubuntu/+source/nplan/+bug/1737976

Other tests seem to be using the Kernel's ioctls directly, to set this mode with `systemd-networkd` [1].

[0] https://www.freedesktop.org/software/systemd/man/systemd.network.html
[1] https://fossies.org/linux/systemd/test/test-network/systemd-networkd-tests.py

slyon:
So... addr-gen-mode was proposed for `systemd-networkd` [0] and merged as of v245 [1], which is included in Ubuntu Focal. Unfortunately, networkd seems to apply `stable-privacy` (aka `prefixstable` / RFC 7217) only to specific networks, whose prefix needs to be given. It falls back to EUI64 otherwise.

[0] https://github.com/systemd/systemd/issues/4625
[1] https://github.com/systemd/systemd/pull/14415

doc/netplan.md
```diff
@@ -239,6 +239,12 @@ Virtual devices
 
     Example: ``addresses: [192.168.14.2/24, "2001:1::1/64"]``
 
+``addr-gen-mode`` (scalar)
```
sil2100, line 242: All good! Let's just get Steve looking at the key-name - maybe he has something better, seeing that networkd has its own thing for naming it.

src/networkd.c
```diff
@@ -487,6 +487,13 @@ write_network_file(const NetplanNetDefinition* def, const char* rootdir, const c
     if (def->ip6_addresses)
         for (unsigned i = 0; i < def->ip6_addresses->len; ++i)
             g_string_append_printf(network, "Address=%s\n", g_array_index(def->ip6_addresses, char*, i));
+    if (def->addr_gen_mode) {
+        /* TODO: figure out if/how we can configure addr-gen-mode for networkd
+         *       IPv6Token= might be an option:
+         *       https://www.freedesktop.org/software/systemd/man/systemd.network.html */
```
sil2100, line 493: Let's maybe update this to include more info, like what you found just now (link to the systemd PR might be good as well).

doc/netplan.md
vorlonofportland: Which addresses are affected by this?  AIUI it's ipv6-only, but does it affect the behavior of both link-local ipv6 addresses and `accept-ra` ipv6 addresses?

At minimum I think this key should mention ipv6, e.g. `ipv6-address-generation`.

PR #127, opened 2020-04-03
bridges: implement vlans & port-vlans options for NM backend

slyon:
## Description
Implements the `vlans` & `port-vlans` option keys for bridges, as required by the "NetworkManager readwrite plugin" spec. Currently only supports the `nm` backend. I need to further research the `networkd` backend...

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -667,6 +667,12 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           a number between ``0`` and ``63``. This metric is used in the
           designated port and root port selection algorithms.
 
+     ``port-vlans`` (sequence of scalars)
+     :    Array of bridge VLAN objects. The VLAN list can be specified with the
+          following syntax: $vid [pvid] [untagged] [, $vid [pvid] [untagged]]..
+          where $vid is either a single id between 1 and 4094 or a range,
+          represented as a couple of ids separated by a dash.
+
      ``forward-delay`` (scalar)
      :    Specify the period of time the bridge will remain in Listening and
           Learning states before getting to the Forwarding state. This field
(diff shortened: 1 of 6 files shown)


sil2100:
Ok, so I still need to think about this one for a bit. For now let's not consider getting this merged as a blocker for the NM read-write pieces, if anything.

nafallo:
For systemd, using default VLAN 1, using vlan 2 for the bridge and adding 2 and 3 to eth0...
```
br0.netdev:
[Bridge]
VLANFiltering=1
DefaultPVID=1

br0.network:
[BridgeVLAN]
PVID=2
EgressUntagged=2

eth0.network:
[BridgeVLAN]
PVID=1
EgressUntagged=1
VLAN=2
VLAN=3
```

slyon:
> For systemd, using default VLAN 1, using vlan 2 for the bridge and adding 2 and 3 to eth0...

I do not really get this comment. Could you please give some more detail?

nafallo:
That was in response to needing further research into the networkd backend, so I thought a comment with how my set up is looking might be helpful?

For the .netdev file of the bridge you need to turn on VLAN filtering and set a default VLAN to use, then tag VLANs in the .network files. You do need to tag the VLANs you want to use from the bridge on the interface going towards the switch, I've learnt the hard way.

slyon:
Ah, I see. So this is an example of how it could be implemented for the networkd backend. Thank you very much! I guess we need to adopt the PR accordingly, as right now this feature is not implemented at all for the networkd backend.

gaby:
Any updates on this? Having VLAN aware bridges as shown in the `networkd` example would be very useful.

slyon:
> Any updates on this?

Unfortunately, no. This is an old PR which needs to be updated to the current codebase. Furthermore, we need to implement the networkd backend as described above. If anybody wants to work on this, feel free to pick it up.

As a workaround you should always be able to use networkd override configs manually, in addition to your normal eth0/br0 netplan definition, e.g.:
```
$ cat /etc/systemd/network/10-netplan-br0.network.d/override.conf
br0.network:
[BridgeVLAN]
PVID=2
EgressUntagged=2
$ cat /etc/systemd/network/10-netplan-eth0.network.d/override.conf
[BridgeVLAN]
PVID=1
EgressUntagged=1
VLAN=2
VLAN=3
```

slyon:
Closing due to inactivty, please re-open if this is still needed in the future.

Cellebyte:
@chdxD1 maybe we add support for this?

PR #130, opened 2020-04-14
SR-IOV support for explicitly defining the VF count

sil2100:
## Description

This is a possible feature request to the just-landed SR-IOV support in netplan. This PR introduces an optional 'virtual-functions:' parameter that can be defined for physical functions to force the allocation of a given number of VFs, regardless of how many are actually used in the netplan config. There are of course safety checks to ensure that we can't request less VFs than actually needed in netplan.

This feature request came from the OpenStack team. We did consider such a thing originally, but then decided that it's better if we let netplan handle it. This PR adds this as an option (not recommended for everyday usage tho).

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -578,9 +578,11 @@ Example:
 
 ``virtual-functions`` (scalar)
 
-:    (SR-IOV devices only) Explicit number of Virtual Functions for the given
-     Physical Function. This is optional, as normally netplan will allocate as
-     many VFs as there are needed.
+:    (SR-IOV devices only) In certain special cases VFs might need to be
+     configured outside of netplan. For such configurations ``virtual-functions``
+     can be optionally used to set an explicit number of Virtual Functions for
+     the given Physical Function. This should be used for special cases only, as
+     normally netplan allocates as many VFs as there are needed.
 
 ## Properties for device type ``modems:``
 GSM/CDMA modem configuration is only supported for the ``NetworkManager`` backend. ``systemd-networkd`` does
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -576,13 +576,14 @@ Example:
       enp1s16f1:
         link: enp1
 
-``virtual-functions`` (scalar)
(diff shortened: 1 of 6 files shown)


doc/netplan.md
```diff
@@ -576,6 +576,12 @@ Example:
       enp1s16f1:
         link: enp1
 
+``virtual-functions`` (scalar)
+
+:    (SR-IOV devices only) Explicit number of Virtual Functions for the given
+     Physical Function. This is optional, as normally netplan will allocate as
+     many VFs as there are needed.
```
slyon, line 583: I think we should state more explicitly that this is only for special/advanced configurations, e.g. something like:

"Additional VFs can be configured outside of netplan. This should be used for special cases only."

sil2100: +1, let me adjust this a bit better, as proposed.

doc/netplan.md
```diff
@@ -576,6 +576,14 @@ Example:
       enp1s16f1:
         link: enp1
 
+``virtual-functions`` (scalar)
+
+:    (SR-IOV devices only) In certain special cases VFs might need to be
+     configured outside of netplan. For such configurations ``virtual-functions``
+     can be optionally used to set an explicit number of Virtual Functions for
+     the given Physical Function. This should be used for special cases only, as
+     normally netplan allocates as many VFs as there are needed.
```
vorlonofportland, line 585: "If unset, the default is to create only as many VFs as are defined in the netplan configuration."

doc/netplan.md
```diff
@@ -576,6 +576,14 @@ Example:
       enp1s16f1:
         link: enp1
 
+``virtual-functions`` (scalar)
```
vorlonofportland, line 579: maybe this should be 'virtual-function-count' rather than 'virtual-functions'

sil2100:
Thank you for the review! I have changed the name as proposed - I took 'virtual-functions' as this was what the original draft naming was. It does look better when it's called 'count'!
Merging!

PR #132, opened 2020-04-20
Add `hidden` to connect to non-broadcast SSIDs

aganders3:
## Description
We got bit this week by not being able to connect to a "hidden" WiFi network with Netplan. A quick search brought up this top result, and we were able to work around the issue with the posted hack:
https://askubuntu.com/questions/1111494/netplan-not-connecting-to-hidden-ssid-server-18-04-1-lts

I'm a first-time contributor, but it didn't take me too long to get this working. The existing code is pleasantly clear and organized! I did notice some repeated code in the `handle_[foo]_bool` methods - I think this is now the third such method. Should some of this logic be moved into a `handle_generic_bool` function? I'm also not familiar with the other backends, so please let me know if there is work to be done to make this more broadly applicable.

The first two checklist items are pending, as I did this development in Docker and could not get the `TestNetplanDBus` tests to pass. I will try to do this in a VM but other feedback is appreciated in the meantime.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: [#1866100](https://bugs.launchpad.net/netplan/+bug/1866100)

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -701,7 +701,7 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
             g_string_append_printf(s, "bssid=%s\n", ap->bssid);
         }
         if (ap->hidden) {
-            g_string_append(s, "hidden=TRUE\n");
+            g_string_append(s, "hidden=true\n");
         }
         if (ap->band == NETPLAN_WIFI_BAND_5 || ap->band == NETPLAN_WIFI_BAND_24) {
             g_string_append_printf(s, "band=%s\n", wifi_band_str(ap->band));
--- a/src/nm.c
+++ b/src/nm.c
@@ -700,6 +700,9 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
         if (ap->bssid) {
             g_string_append_printf(s, "bssid=%s\n", ap->bssid);
         }
+        if (ap->hidden) {
+            g_string_append(s, "hidden=TRUE\n");
+        }
         if (ap->band == NETPLAN_WIFI_BAND_5 || ap->band == NETPLAN_WIFI_BAND_24) {
             g_string_append_printf(s, "band=%s\n", wifi_band_str(ap->band));
             /* Channel is only unambiguous, if band is set. */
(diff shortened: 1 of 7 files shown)


aganders3:
> The first two checklist items are pending, as I did this development in Docker and could not get the TestNetplanDBus tests to pass. I will try to do this in a VM but other feedback is appreciated in the meantime.

Oh nice, the helpful bot seems to have done this for me!

src/parse.c
```diff
@@ -572,6 +572,29 @@ get_default_backend_for_type(NetplanDefType type)
     return NETPLAN_BACKEND_NETWORKD;
 }
 
+static gboolean
+handle_access_point_bool(yaml_document_t* doc, yaml_node_t* node, const void* data, GError** error)
+{
+    guint offset = GPOINTER_TO_UINT(data);
+    gboolean v;
+
+    if (g_ascii_strcasecmp(scalar(node), "true") == 0 ||
```
sil2100, line 581: I know that refactoring should not really be part of new functionality PRs, but here I feel like we might want to get rid of the redundancy. This bool checking is happening too frequently and has just copy-pasted code all the time. I'd look into splitting this out instead.

aganders3: Yep, I agree as mentioned in the main PR description. I'm happy to do the refactor either way - in this PR or in a separate one, just wasn't sure what the preference would be.

aganders3:
> Another thing: I see this only adds support for it the networkd backend - is there any reason not doing the same for NetworkManager? Guess that would mean using the 'hidden' setting.

I don't know anything about NetworkManager since we don't use it. I had assumed it does not need special configuration for "hidden" SSIDs and that this would not be necessary, but I can look into it. Of course if it does need something this would be the appropriate place to add it!

aganders3:
OK - I think I found that for NetworkManager you pass `hidden=TRUE`, so I can add that.

Another design question: if a user passes a default value to Netplan, do you make sure to write that default value in the output? For example if a user given `hidden: no` in Netplan, should NetworkManager file have `hidden=FALSE`? Likewise, would the networkd backend add `scan_ssid=0` to the wpa_supplicant config?

src/parse.c
aganders3: I refactored this in 3aacdfc

src/nm.c
```diff
@@ -700,6 +700,9 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
         if (ap->bssid) {
             g_string_append_printf(s, "bssid=%s\n", ap->bssid);
         }
+        if (ap->hidden) {
+            g_string_append(s, "hidden=TRUE\n");
```
slyon, line 704: Just a little stylistic nitpick: Could you please change this to "hidden=true", we use lower case boolean values in the rest of the code, so I'd like to keep this consistent.

aganders3: No problem - quick fix!

aganders3:
Thanks! I appreciate the feedback and guidance. Hopefully people can make use of these changes. Let me know if there's anything else needed to get this merged.

PR #133, opened 2020-04-23
Fix LP#1874377: Not connect to WiFi after 'netplan apply'

slyon:
## Description
Seems like the 'netplan apply' command was not properly adopted in https://github.com/CanonicalLtd/netplan/pull/109 when wired wpa_supplicant support was introduced.

Fixes: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1874377

## Checklist

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -109,6 +109,9 @@ class NetplanApply(utils.NetplanCommand):
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, restarting networkd')
             utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa-*.service'])
+            # Historically (up to v0.98) we had netplan-wpa@*.service files, in case of an
+            # upgraded system, we need to make sure to stop those.
+            utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa@*.service'])
         else:
             logging.debug('no netplan generated networkd configuration exists')
 
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -108,7 +108,7 @@ class NetplanApply(utils.NetplanCommand):
         # stop backends
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, restarting networkd')
-            utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa@*.service'])
+            utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa-*.service'])
         else:
             logging.debug('no netplan generated networkd configuration exists')
 
@@ -169,7 +169,7 @@ class NetplanApply(utils.NetplanCommand):
 
         # (re)start backends
         if restart_networkd:
-            netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa@*.service')]
+            netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa-*.service')]
             utils.systemctl_networkd('start', sync=sync, extra_services=netplan_wpa)
         if restart_nm:
             utils.systemctl_network_manager('start', sync=sync)
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -108,13 +108,13 @@ class NetplanApply(utils.NetplanCommand):
         # stop backends
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, restarting networkd')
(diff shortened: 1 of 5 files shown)


slyon:
@xnox found that we need to consider the following scenario:

> assuming the scenario is: used 2GHz network, upgraded netplan to 0.99, changed yaml to use 5GHz network, run netplan apply => @*.service units need to be stopped, but the new -* units need to be started

netplan/cli/commands/apply.py
```diff
@@ -108,7 +108,7 @@ def command_apply(run_generate=True, sync=False, exit_on_error=True):  # pragma:
         # stop backends
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, restarting networkd')
-            utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa@*.service'])
+            utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa-*.service'])
```
xnox, line 111: i'd rather have `['netplan-wpa@*.service', 'netplan-wpa-*.service']`

After netplan upgrade, netplan generate will wipe old units, but they might still be running. Netplan apply then needs to stop them. It's ok to stop non-existent / unknown units. But maybe need to test that with older systemctls. If old systemctls doesn't handle the calls to stop non-existent / unknown units we then should do as two calls:
```
utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa@*.service'])
utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa-*.service'])
``

Comment (slyon) [reply netplan/cli/commands/apply.py:111]: Yes, I'd rather go the safe way, using two calls.
This way we can also better separate the two calls with a comment about this scenario.

**Comment by slyon** [comment]
I think `netplan apply` should be fine now... but how do we handle upgrade scenarios, where the old service file is enabled/linked in the filesystem (e.g. via `/lib/systemd/system/netplan-wpa@.service`) – so it would auto-start after reboot?

I think we need some kind of migration handler (at netplan startup), which can rename/move or disable the old service files.

**Comment by slyon** [comment]
With my latest commits the upgrade problem should be fixed.
It now cleans up old systemd service files in `netplan generate`, which is also run as a systemd generator via `/lib/systemd/system-generators/netplan` at early boot time.

So now at each reboot the old systemd dependencies are removed and new ones are generated.

File: netplan/cli/commands/apply.py [hunk 2]
```diff
@@ -108,7 +108,13 @@ def command_apply(run_generate=True, sync=False, exit_on_error=True):  # pragma:
...
+            # Historically (up to v0.98) we had netplan-wpa@*.service files, in case of an
+            # upgraded system, we need to make sure to stop those.
+            try:
+                utils.systemctl_networkd('stop', sync=sync, extra_services=['netplan-wpa@*.service'])
```
sil2100, line 115: I am wondering a bit about this now. So even though this practically is alright, I don't think I'm a big fan of doing an extra systemctl call just for legacy reasons everytime we run `netplan apply`. I mean, this is fine, but feels ugly. Maybe we could do this in the packaging somehow? I am not a systemd expert, but maybe we could somehow, on package upgrade, symlink the old name in /run/systemd to the new one? This would then be persistent only for one 'run', and would be transparent to the user.

slyon: You are right and I like the way you are thinking! I'm not an systemd expert either, but after doing some experiments, I think this is unfortunately not possible.

Basically we need to "rename" a running service inside systemd's memory. `netplan generate` will already do the renaming (unlink & re-create) of the legacy & new systemd unit-files for us at reboot or at any `netplan apply` run.

But the problem is, that the old units are still running (can be seen via `systemctl show netplan-wpa@*`), even though their unit-files/symlinks are not existing in the filesystem anymore. Only the systemd process knows the PID of those old, running `netplan-wpa@*` units and can send the proper kill command to stop them.

I was experimenting with `systemctl set-property netplan-wpa@wlan0 Id=netplan-wpa-wlan0 Names=netplan-wpa-wlan0`, trying to rename the running units, but this fails with `Unknown assignment:` errors.

sil2100: That's too bad then, guess we have no choice for now. Was thinking about maybe first querying if there are any services with the old name before running systemctl, but then again it's not a very time wasteful operation.

slyon: I did some more experiments:

Stopping non-existing services takes approx. 40ms on my machine, while 'is-active' checking takes approx. 10ms. So we should actually opt to only call the stop command, iff they are running (only first 'netplan apply' run after v0.98->v0.99 upgrade).

Also, I verified `systemctl stop` of non-existing services works as expected on older systemd (v237 / bionic) and implemented Dimitri's prefered way of stopping those services, in order to avoid calling `systemctl stop systemd-networkd.service` twice.

PR #134, opened 2020-05-06
Implement DBus try-commit handling

slyon:
## Description
Implement the `Try(uint32 timeoutInSeconds)` and `Cancel()` command inside `netplan-dbus` and adopt the `Apply()` command according to spec. "Try(...)" calls `netplan try`, keeping `stdin` open. The "Apply()" command uses the `SIGUSR1` signal, to send a confirmation to the `netplan try` process to confirm/commit the settings. The "Cancel()" command uses the `SIGINT` signal, to send a rejection to the `netplan try` child process.
The child process is being monitored, using events on the sd-event mainloop.

Codecov reports a reduced coverage, but that is only because the code in `terminal.py` cannot be tested from a script, as it uses the `isatty()` condition. And that part is excluded/skipped from our test-suite if run inside a script. If `make check-coverage` is executed locally, it retains 100% coverage.

## Checklist

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -20,7 +20,6 @@
  * correctly capture tests being run over a DBus bus.
  */
 
-static sd_bus *bus = NULL;
 static sd_event_source *_try_child_es = NULL;
 static gint _try_child_stdin = -1;
 static GPid _try_child_pid = -1;
@@ -46,7 +45,7 @@ static int method_apply(sd_bus_message *m, void *userdata, sd_bus_error *ret_err
     if (err != NULL) {
        return sd_bus_error_setf(ret_error, SD_BUS_ERROR_FAILED, "netplan apply failed: %s\nstdout: '%s'\nstderr: '%s'", err->message, stdout, stderr);
     }
-    
+
     return sd_bus_reply_method_return(m, "b", true);
 }
 
@@ -204,7 +203,7 @@ static int _try_child_handler(sd_event_source *es, const siginfo_t *si, void* us
     _try_child_clear();
     _try_child_stdin = -1;
     _try_child_pid = -1;
-    return _config_changed_signal(bus);
+    return _config_changed_signal(userdata);
 }
 
 static int method_try(sd_bus_message *m, void *userdata, sd_bus_error *ret_error) {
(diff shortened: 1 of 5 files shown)


sil2100:
On it, but I might need a bit to properly review!

src/dbus.c
```diff
@@ -183,24 +332,23 @@ int main(int argc, char *argv[]) {
...
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGCHLD);
+    sigprocmask(SIG_BLOCK, &mask, NULL);
+
+    /* Start the event loop, wait for requests */
+    r = sd_event_loop(event);
+    if (r < 0)
+        fprintf(stderr, "Failed mainloop: %s\n", strerror(-r));
 finish:
+    g_free(data);
```
sil2100, line 351: Do we also need/should do a `sd_event_unref(event);` here? Or maybe it's not really necessary?

src/dbus.c
```diff
@@ -18,11 +20,86 @@
...
     g_autoptr(GError) err = NULL;
     g_autofree gchar *stdout = NULL;
     g_autofree gchar *stderr = NULL;
     gint exit_status = 0;
+    NetplanData *d = userdata;
+
+    /* Accept the current 'netplan try', if active.
+     * Otherwise execute 'netplan apply' directly. */
+    if (d->try_pid > 0)
+        return _try_accept(TRUE, m, userdata, ret_error);
```
sil2100, line 102: Oh, good catch with this case here!

sil2100 (approved):
I must say that I'm not a DBus expert, so this took longer for me to process. But architecture-wise and code-wise this looks like a good hit. Will still play around with it locally myself later today, but it's all good as is. +1

src/dbus.c
slyon: Absolutely! This is the proper way to clean it up. The application terminates afterwards, so it doesn't make that big of a difference, but still we should have it properly cleaned up in the code.

Fixed.

slyon:
Thank you for the review!

I've fixed the `sd_event` cleanup and removed some unneeded safeguards, for easier testability.

Now the code can be executed locally via:
`sudo LD_LIBRARY_PATH=/home/lukas/canonical/netplan PYTHONPATH=/home/lukas/canonical/netplan DBUS_TEST_NETPLAN_CMD=/home/lukas/canonical/netplan/src/netplan.script ./netplan-dbus`

And called via DBus:
`sudo busctl call io.netplan.Netplan /io/netplan/Netplan io.netplan.Netplan Try u 20` (or `Apply`, `Cancel`, ...)

PR #135, opened 2020-05-13
Call daemon-reload after we touched systemd unit files (LP: #1874494)

slyon:
##  Description
When running `netplan apply` systemd unit files might be created/deleted/changed on disk. We need to run `systemctl daemon-reload`, to make systemd aware of those new units and calculate the new unit/service dependencies. Otherwise it will throw warnings at us:
```
ubuntu@ubuntu:~$ sudo netplan apply
Warning: The unit file, source configuration file or drop-ins of netplan-wpa-wlan0.service changed on disk. Run 'systemctl daemon-reload' to reload units.
Warning: The unit file, source configuration file or drop-ins of netplan-wpa-wlan0.service changed on disk. Run 'systemctl daemon-reload' to reload units.
```

Fixes: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1874494

## Checklist

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -175,6 +175,10 @@ class NetplanApply(utils.NetplanCommand):
 
         # (re)start backends
         if restart_networkd:
+            # Running 'systemctl daemon-reload' will re-run the netplan systemd generator,
+            # so let's make sure we only run it iff we're willing to run 'netplan generate'
+            if run_generate:
+                utils.systemctl_daemon_reload()
             netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa-*.service')]
             utils.systemctl_networkd('start', sync=sync, extra_services=netplan_wpa)
         if restart_nm:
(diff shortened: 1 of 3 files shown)


slyon:
Integration tests looking good! Also, I verified it manually on RaspberryPi, all OK.

netplan/cli/commands/apply.py
```diff
@@ -108,6 +108,10 @@ def command_apply(run_generate=True, sync=False, exit_on_error=True):  # pragma:
         # stop backends
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, restarting networkd')
+            # Running 'systemctl daemon-reload' will re-run the netplan systemd generator,
+            # so let's make sure we only run it iff we're willing to run 'netplan generate'
```
sil2100, line 112: Typo, 'iff'.

slyon: 'iff' was used on purpose, with the meaning of 'if and only if'.

xnox:
When wifi was a template @.service unit, we didn't change any units.

PR #137, opened 2020-05-14
Doc: Mark versions of new config keys

slyon:
## Description
Documenting the first version when a new feature (config key) was introduced down to `v0.96`. No new keys were introduced in 0.96.

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -133,7 +133,7 @@ Virtual devices
 
 :    Enable wake on LAN. Off by default.
 
-``emit-lldp`` (bool)
+``emit-lldp`` (bool) _– since 0.99_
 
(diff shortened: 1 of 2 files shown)


vorlonofportland:
I agree that an underline of a dash is a strange rendering choice.  What about either putting the dash outside the underline, or omitting the dash completely given that the underline itself will already provide visual separation?

vorlonofportland:
+1 on the latest proposal.

slyon:
Thanks, I will adopt the documentation accordingly!

slyon:
@sil2100 This should be ready for merging now.
I've also updated the `validate_docs.sh` test, to account for the new number of config keys.

PR #139, opened 2020-05-22
Added TTL option for tunnels (LP: #1846783)

kev1989:
## Description
Some protocols set the TTL field of the packet to 1; when passing through the tunnel, the packet is discarded. To solve the problem, the tunnel has the TTL option, but it was not in netplan.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/netplan/+bug/1846783

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -583,6 +583,14 @@ These options are available for all types of interfaces.
      :    The MTU to be used for the route, in bytes. Must be a positive integer
           value.
 
+    ``congestion-window`` (scalar) – since **0.102**
+     :    The congestion window to be used for the route, represented by number
+          of segments. Must be a positive integer value.
+
+    ``advertised-receive-window`` (scalar) – since **0.102**
+     :    The receive window to be advertised for the route, represented by
+          number of segments. Must be a positive integer value.
+
 ``routing-policy`` (mapping)
 
 :    The ``routing-policy`` block defines extra routing policy for a network,
--- a/src/parse.h
+++ b/src/parse.h
@@ -425,6 +425,8 @@ typedef struct {
     gboolean has_auth;
 } NetplanWifiAccessPoint;
 
+#define NETPLAN_ADVERTISED_RECEIVE_WINDOW_UNSPEC 0
+#define NETPLAN_CONGESTION_WINDOW_UNSPEC 0
 #define NETPLAN_MTU_UNSPEC 0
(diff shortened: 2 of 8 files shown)


slyon:
Hey, thank you for your contribution.
To better understand the problem, could you please specify the protocol, where you can observe this behavior?

doc/netplan.md
```diff
@@ -967,6 +967,10 @@ more general information about tunnels.
 
 :   Defines the address of the remote endpoint of the tunnel.
 
+``ttl`` (scalar)
```
slyon, line 970: Could you please append ` – since **0.100**` to this line, so we can document that this setting will be introduced with the current development version (0.100)

src/parse.h
```diff
@@ -329,6 +329,7 @@ struct net_definition {
         char *remote_ip;
         char *input_key;
         char *output_key;
+	guint ttl;
```
slyon, line 332: Please fix the indentation to be in line with the other variables of this struct (8 spaces).

src/validation.c
```diff
@@ -72,6 +72,11 @@ validate_tunnel_grammar(NetplanNetDefinition* nd, yaml_node_t* node, GError** er
     if (!nd->tunnel.remote_ip)
         return yaml_error(node, error, "%s: missing 'remote' property for tunnel", nd->id);
 
+    if (nd->tunnel.ttl) {
+	if ((nd->tunnel.ttl) < 1 || (nd->tunnel.ttl) > 255)
```
slyon, line 76: I think we can combine the two `if`s.. if `(nd->tunnel.ttl)` is true it will be > 0 (as this is a uint variable), which makes the check for `nd->tunnel.ttl < 1` useles.

So we could just combine it as: `if (nd->tunnel.ttl && nd->tunnel.ttl > 255)`

wolfy1339:
This would close https://bugs.launchpad.net/netplan/+bug/1846783

slyon:
According to https://bugs.launchpad.net/netplan/+bug/1846783 this is required for IPIP/SIT/GRE tunnels.
I addressed my concerns mentioned above.

slyon:
I accidentaly closed this PR, I cannot re-open as I do not have access to @kev1989 's `master` branch. So I've setup a new PR containing the changes, incl. my fixes: #194

slyon:
This PR was now merged. Thank you very much for your contribution @kev1989 !

PR #140, opened 2020-05-25
Add itegration tests for interface globbing and NM routing

slyon:
## Description
This should actually fail if used with NetworkManager < 1.14, e.g. on Ubuntu Bionic.

## Checklist

### DIFF
--- a/tests/integration/ethernets.py
+++ b/tests/integration/ethernets.py
@@ -77,6 +77,9 @@ class _CommonTests():
                                'address', self.dev_e2_client_mac])
 
     def test_eth_glob(self):
+        '''Supposed to fail if tested against NetworkManager < 1.14
+
+        Interface globbing was introduced as of NM 1.14+'''
         self.setup_eth(None)
         self.start_dnsmasq(None, self.dev_e2_ap)
         with open(self.config, 'w') as f:
--- a/tests/integration/ethernets.py
+++ b/tests/integration/ethernets.py
@@ -89,6 +89,10 @@ class _CommonTests():
 ''' % {'r': self.backend}) # globbing match on "eth42", i.e. self.dev_e_client
         self.generate_and_settle()
         self.assert_iface_up(self.dev_e_client, ['inet 172.16.42.99/18', 'inet6 1234:ffff::42/64'])
+        out = subprocess.check_output(['ip', 'a', 'show', 'dev', self.dev_e_client],
+                                      universal_newlines=True)
+        self.assertIn('inet 172.16.42.99/18', out)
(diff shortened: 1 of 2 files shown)


slyon:
Okay, so I've added integration tests for interface globbing and route options, which are run on the `networkd` and `NetworkManager` backends.

The globbing test works as expected (interface globbing is supported as of NM 1.14):
- FAIL on Ubuntu Bionic (NM 1.10.6-2ubuntu1.4)
- PASS on Ubuntu Focal (NM 1.22.10-1ubuntu1)
- PASS on Ubuntu Groovy (1.22.10-1ubuntu3)

But, the route options tests behave in a really odd way...
`test_route_on_link` (should be supported as of NM 1.12+)
- PASS on Ubuntu Bionic (NM 1.10.6-2ubuntu1.4)
- PASS on Ubuntu Focal (NM 1.22.10-1ubuntu1)
- PASS on Ubuntu Groovy (1.22.10-1ubuntu3)

`test_route_table` (should be supported as of NM 1.10+)
- PASS on Ubuntu Bionic (NM 1.10.6-2ubuntu1.4)
- FAIL on Ubuntu Focal (NM 1.22.10-1ubuntu1)
- FAIL on Ubuntu Groovy (1.22.10-1ubuntu3)

Let me provide full logs (once all of them are done)... maybe someone else can have another look then.

slyon:
I've added the full logs and partial logs for the relevant tests. Would be glad if somebody could have a look at it.

We could skip the `test_route_table` test on `NetworkManager` for now to make use of the tests, or maybe @thom311 could have another look of how the route options correlate with the different versions of NetworkManager.

thom311:
from the provided test output it's not clear to me what happened. Can you show 

 - the generated keyfile
 - if necessary, what does NM say about the profile with `nmcli connection show filename "$KEYFILE_NAME"`
 - if necessary, what happens when activating the profile (full level=TRACE log and `ip route` output)

sil2100:
Before we proceed with this one I'd like to better understand what's going on with the routing tests. Could you maybe hack in the tests to provide a bit more info about the networking stack state when it fails? Guess what Thomas requested seems like something we could use while debugging.

slyon:
Thanks for your input! After checking with some more debugging output, I was able to fix the two problems (mostly related to different default settings). Our integration tests produce the expected results now – FAIL on Ubuntu Bionic (NM 1.10), PASS on Focal/Groovy.

@sil2100 I would be glad if you could have a final look at my code and merge if you are OK with it.

slyon:
Thanks, merged!

PR #141, opened 2020-05-26
Work on a better first-boot experience with SR-IOV

sil2100:
## Description
Currently there's no real first boot experience in SR-IOV. I propose adding some simple udev rules so that SR-IOV initialization can be performed on first boot and whenever a new SR-IOV-enabled device appears. The rule then calls netplan apply SR-IOV-only part, which is fine to be executed multiple times. We could add more fine grained control and only perform initialization of the device that's appearing, but that would require additional code changes - and it's fine as is like it is right now.

Related packaging changes: TBD

## Checklist

### DIFF
--- a/netplan/cli/sriov.py
+++ b/netplan/cli/sriov.py
@@ -30,25 +30,35 @@ import netifaces
 def _get_target_interface(interfaces, config_manager, pf_link, pfs):
     if pf_link not in pfs:
         # handle the match: syntax, get the actual device name
-        pf_match = config_manager.ethernets[pf_link].get('match')
+        pf_dev = config_manager.ethernets[pf_link]
+        pf_match = pf_dev.get('match')
         if pf_match:
-            by_name = pf_match.get('name')
-            by_mac = pf_match.get('macaddress')
-            by_driver = pf_match.get('driver')
+            # now here it's a bit tricky
+            set_name = pf_dev.get('set-name')
+            if set_name:
+                # if we had a match: stanza and set-name: this means we should
+                # assume that, if found, the interface has already been
+                # renamed - use the new name
+                pfs[pf_link] = set_name
+            else:
+                # no set-name, so we need to do the matching ourselves
+                by_name = pf_match.get('name')
+                by_mac = pf_match.get('macaddress')
+                by_driver = pf_match.get('driver')
 
-            for interface in interfaces:
-                if ((by_name and not utils.is_interface_matching_name(interface, by_name)) or
-                        (by_mac and not utils.is_interface_matching_macaddress(interface, by_mac)) or
-                        (by_driver and not utils.is_interface_matching_driver_name(interface, by_driver))):
-                    continue
-                # we have a matching PF
-                # store the matching interface in the dictionary of
-                # active PFs, but error out if we matched more than one
-                if pf_link in pfs:
-                    raise ConfigurationError('matched more than one interface for a PF device: %s' % pf_link)
-                pfs[pf_link] = interface
+                for interface in interfaces:
+                    if ((by_name and not utils.is_interface_matching_name(interface, by_name)) or
+                            (by_mac and not utils.is_interface_matching_macaddress(interface, by_mac)) or
+                            (by_driver and not utils.is_interface_matching_driver_name(interface, by_driver))):
+                        continue
+                    # we have a matching PF
+                    # store the matching interface in the dictionary of
+                    # active PFs, but error out if we matched more than one
+                    if pf_link in pfs:
+                        raise ConfigurationError('matched more than one interface for a PF device: %s' % pf_link)
+                    pfs[pf_link] = interface
         else:
-            # no match field, assume entry name is interface name
+            # no match field, assume entry name is the interface name
             if pf_link in interfaces:
                 pfs[pf_link] = pf_link
 
@@ -219,11 +229,13 @@ def apply_vlan_filter_for_vf(pf, vf, vlan_name, vlan_id, prefix='/'):
             'failed setting SR-IOV VLAN filter for vlan %s (ip link set command failed)' % vlan_name)
 
 
-def apply_sriov_config(interfaces, config_manager):
+def apply_sriov_config(config_manager):
     """
     Go through all interfaces, identify which ones are SR-IOV VFs, create
     them and perform all other necessary setup.
     """
+    config_manager.parse()
+    interfaces = netifaces.interfaces()
 
     # for sr-iov devices, we identify VFs by them having a link: field
     # pointing to an PF. So let's browse through all ethernet devices,
--- a/netplan/cli/sriov.py
+++ b/netplan/cli/sriov.py
@@ -35,13 +35,14 @@ def _get_target_interface(interfaces, config_manager, pf_link, pfs):
         if pf_match:
             # now here it's a bit tricky
             set_name = pf_dev.get('set-name')
-            if set_name:
+            if set_name and set_name in interfaces:
                 # if we had a match: stanza and set-name: this means we should
                 # assume that, if found, the interface has already been
                 # renamed - use the new name
                 pfs[pf_link] = set_name
             else:
-                # no set-name, so we need to do the matching ourselves
+                # no set-name (or interfaces not yet renamed) so we need to do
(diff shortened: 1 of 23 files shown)


sil2100:
So I'm still thinking about this one a bit, because I need to actually figure out what is the best moment to run all the SR-IOV setup. Currently it's written in such a way that it's best to run it early. But maybe that's a bug. Let me think about it.

sil2100:
Ok, this should now be good for review. I will be writing two simple integration tests, but I guess I can do that in a separate PR if needed. Lukas, could you take a look?
As far as the approach goes, I was thinking of maybe the systemd generator generating (or copying) the .service file only in the case if there's an SR-IOV configuration detected, but that would add additional code that I don't think is actually needed. But if you think this is the right way to go indeed, I can look into that.

systemd/netplan-sriov-setup.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Netplan one-shot initial SR-IOV configuration service
+DefaultDependencies=no
+After=systemd-udevd-settle.service
+Wants=systemd-udevd-settle.service
```
xnox, line 5: `systemd-udevd-settle.service` is a deprecated service which should not be used anymore.

Do you mean ordering after `systemd-udev-trigger.service` unit instead?

Overall the unit feels very wrong. For example, shouldn't you simply install a udev rule that is sensitive to network device with sriov functionality, which should then declare `SYSTEMD_WANTS=netplan-sriov-config@%i.service` which would then start that unit and ask netplan to configure sriov on the device? Possibly as a remainafterexit & oneshot service.

Because this needs to happen whenever netplan config appears/ changed, and/or applied whenever network devices appear. Hence, it is more natural to have it udev driven with a custom udev rule.

Or like a networkd-dispatcher hook sensitive to devices appearing? But then not sure if this will work with network-manager.

sil2100: Ah, indeed, thanks for pointing this out!

We basically only need this to run once on boot time in theory, since for a running system it is assumed that, whenever the network/device setup changes, one has to call `netplan apply` explicitly. That being said, I guess this is also an option - and I have an idea how to make an udev rule only for SR-IOV devices. Let me get back to you in a bit.

sil2100: Ok, so even though I found a nice way to write an udev rule that would work for any supported sriov_device (by matching by `ATTRS{sriov_totalvfs}`), I do not think this is a very easy way to go right now.

The problem I see with this is that right now it's most optimal to deal with SR-IOV devices as a whole when they are all available and ready to be configured. I could refactor the SR-IOV initialization so that it could be called for a single device, sure, but then netplan needs to do a lot of parsing and matching multiple times because of how the whole thing has been designed. Like, each time netplan would have to match which VF comes from which PF, then evaluate which PF has been added, then apply the config for the given set of PFs and VFs. Then another device pops up and the same parsing would have to happen for no reason. I prefer to just wait for all the devices to come up and then call SR-IOV initialization once, if possible. Sure, we don't have the fine granularity, but how often do we hot-plug SR-IOV devices?

Another thing I'd have to wrap my head around is the timing. Like, I explicitly set `systemd-udevd-settle` as the target as I need the SR-IOV initialization to happen *after* all the interface renaming has happened. I don't know these parts too well, but how could I make sure that all the related network device renaming per netplan's set-name have been done and only then do my SR-IOV initialization? Since this is the goal.

sil2100: On the other hand, guess I'd live with the extra parsing, but I wonder if there's a way to just make sure the interface has been renamed when the netplan trigger is happening. Let me experiment some more.

xnox: @sil2100 you can make netplan generate create a target. Then from neptlan yaml, generate dependencies on all of the sriov devices that netplan says to configure. Then make the target depend on all of those devices. Then make the unit that does `netplan-sriov-setup` to be after the target, this way it will be run once, after all the sriov devices are up.

sil2100: Argh, this is terrible. The worst thing is that even though I can get the right timing done for networkd (as I can create udev rules that would fire after the interfaces have final names), I can't really do that easily for NetworkManager. And it's important in this approach as the VFs need to exist when NetworkManager starts configuring them as otherwise they won't get all the usual settings. So argh, I'd probably have to restart NM after applying SR-IOV config, which really sucks.

Ok, so maybe instead I should revert to my original idea of running SR-IOV initialization really early, before the renames. I didn't like that approach as even though it would look really nice during boot, on runtime its a bit confusing as some interfaces can have new names and others - old. But maybe it makes sense to make this smarter? Like, take into consideration process_link_changes() and then check in the SR-IOV code which name should be considered? Let me look into that then.

src/sriov.c
```diff
@@ -0,0 +1,40 @@
...
+#include <glib-object.h>
+
+#include "util.h"
+
+void
+write_sriov_conf_finish(const char* rootdir)
+{
+    /* For now we execute apply --sriov-only everytime there is a new
+       SR-IOV device appearing, which is fine as it's relatively fast */
+    GString *udev_rule = g_string_new("ACTION==\"add\", SUBSYSTEM==\"net\", ATTRS{sriov_totalvfs}==\"?*\", RUN+=\"/sbin/netplan apply --sriov-only\"\n");
```
slyon, line 31: I'm not sure if this changes during early boot... but I think it should be `/usr/sbin/netplan ...`

netplan/cli/sriov.py
```diff
@@ -30,25 +30,36 @@
...
-                if pf_link in pfs:
-                    raise ConfigurationError('matched more than one interface for a PF device: %s' % pf_link)
-                pfs[pf_link] = interface
+            # now here it's a bit tricky
+            set_name = pf_dev.get('set-name')
+            if set_name and set_name in interfaces:
+                # if we had a match: stanza and set-name: this means we should
+                # assume that, if found, the interface has already been
+                # renamed - use the new name
+                pfs[pf_link] = set_name
```
slyon, line 42: I had some problems recently with `set-name:`... it does not seem to always work reliably. The assumption you make here is still valid and needs no changes!

I just want to point out that we should observe the overall `set-name:` functionality in the future and see if there is any debugging or fixes needed.

src/parse.c
```diff
@@ -1888,6 +1903,7 @@ handle_network_type(yaml_document_t* doc, yaml_node_t* node, const void* data, G
             g_hash_table_insert(netdefs, cur_netdef->id, cur_netdef);
             netdefs_ordered = g_list_append(netdefs_ordered, cur_netdef);
             cur_netdef->sriov_vlan_filter = FALSE;
+            cur_netdef->sriov_explicit_vf_count = -1;
```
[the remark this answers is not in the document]
sil2100: Oh, I forgot we did that there! Yeah, that would be MUCH better. The only reason why I did the switch to gint from guint was being able to have an invalid value. Good catch, on it.

src/sriov.c
```diff
@@ -0,0 +1,40 @@
...
+    g_string_free_to_file(udev_rule, rootdir, "run/udev/rules.d/999-netplan-sriov-setup.rules", NULL);
```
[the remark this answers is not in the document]
sil2100: I guess we basically can. The only reason I set the priority so high was that I wanted to be safe and run those after out regular netplan udev rules, which are usually priority `99-*`. This is of course probably just paranoid, but I wanted to make sure that if we write any udev rules for interface renames etc., those get executed first, before the `netplan apply --sriov-only`.

But maybe we don't actually need to be so paranoid. I'll look into that!

sil2100:
Hey! Thanks for the review - I have pushed changes per your comments. As I still want the SR-IOV config to run as late as possible, I still use priority `99-*`, but renamed it slightly so that (hopefully) if there's an alphanumerical sort, they should still execute after the `99-netplan-*` ones.

Could you take a look now?

sil2100:
Thanks! I just did a quick re-test of this on an SR-IOV enabled instance - added `virtual-function-count: 3` to the SR-IOV interface, rebooted and observed that 3 VFs have been created on boot time. Didn't have time for more testing, but this looks good so far.

PR #143, opened 2020-05-28
Fix process_link_changes handling 'up' interfaces

hrasiq:
## Description
b7f1d9b04212 refactored `process_link_changes` with helper methods to get
the interface driver name and MAC address. This new code introduced a
regression where it's possible for an interface in the up state to be
included in the changelist by its MAC address.

This patch restores the previous behaviour, skipping interfaces that
don't have driver_name set.

Fixes: https://bugs.launchpad.net/bugs/1875411

## Checklist

- [X] \(Optional\) Closes an open bug in Launchpad.


slyon:
Should we have an integration test for this, to make sure it keeps working in the future?

PR #145, opened 2020-06-09
OVS: Add support for bridge controller configuration

slyon:
## Description
Adds support for the OpenVSwitch Bridge `controller.addresses` config key to specify OpenFlow controller targets and `controller.connection-mode` config key, to define if the controller shall be used in a `in-band` or `out-of-band` configuration.

Additionally, it adds support for the following global `openvswitch.ssl` configs, which need to be used if a `ssl:` or `pssl:` OpenFlow controller target is to be used.
* ca-cert
* certificate
* private-key

It adds a new integration test in `tests/integration/ovs.py`, which we need to make sure to integrate into the debian package (via `debian/tests/control`), to have it run during `autopkgtest`.

## Checklist

### DIFF
--- a/src/openvswitch.c
+++ b/src/openvswitch.c
@@ -224,6 +224,18 @@ write_ovs_protocols(const NetplanOVSSettings* ovs_settings, const gchar* bridge,
     g_string_free(s, TRUE);
 }
 
+static void
+write_ovs_bridge_controller_targets(const NetplanOVSController* controller, const gchar* bridge, GString* cmds)
+{
+    GString* s = g_string_new(g_array_index(controller->addresses, char*, 0));
+
+    for (unsigned i = 1; i < controller->addresses->len; ++i)
+        g_string_append_printf(s, " %s", g_array_index(controller->addresses, char*, i));
+
+    append_systemd_cmd(cmds, OPENVSWITCH_OVS_VSCTL " set-controller %s %s", bridge, s->str);
+    g_string_free(s, TRUE);
+}
+
 /**
  * Generate the OpenVSwitch systemd units for configuration of the selected netdef
  * @rootdir: If not %NULL, generate configuration in this root directory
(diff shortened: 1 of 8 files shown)


src/openvswitch.c
```diff
@@ -224,6 +224,48 @@ write_ovs_protocols(const NetplanOVSSettings* ovs_settings, const gchar* bridge,
...
+    return FALSE;
+}
+
+static void
+write_ovs_bridge_controller_targets(const NetplanOVSController* controller, const gchar* bridge, GString* cmds)
+{
+    gboolean needs_ssl = FALSE;
+    g_autofree gchar* ssl = "";
+    gchar* target = g_array_index(controller->addresses, char*, 0);
+    needs_ssl |= check_ovs_ssl(target);
```
sil2100, line 249: Any reason why we don't simply start off with `gboolean needs_ssl = check_ovs_ssl(target);`?

src/openvswitch.c
```diff
@@ -224,6 +224,48 @@ write_ovs_protocols(const NetplanOVSSettings* ovs_settings, const gchar* bridge,
...
+    GString* s = g_string_new(target);
+
+    for (unsigned i = 1; i < controller->addresses->len; ++i) {
+        target = g_array_index(controller->addresses, char*, i);
+        needs_ssl |= check_ovs_ssl(target);
```
sil2100, line 254: I know this way just looks nice, but currently this means performing 2 unnecessary strcmp's everytime + other conditionals. It's more of a nitpick, but I think I'd prefer this the 'old-fashioned' way, like `if (!needs_ssl) needs_ssl = check_ovs_ssl(target);`. Since if needs_ssl is TRUE, it means we already checked for the ssl settings existence and don't need to do that again (as we'd throw an error if we didn't).

src/validation.h
```diff
@@ -22,6 +22,7 @@
 
 gboolean is_ip4_address(const char* address);
 gboolean is_ip6_address(const char* address);
+gboolean validate_ovs_target(gboolean host_first, gchar* s);
```
sil2100, line 25: I like this additional validator.

src/parse.c
```diff
@@ -1787,6 +1787,77 @@ handle_ovs_bridge_protocol(yaml_document_t* doc, yaml_node_t* node, const void*
...
+        assert_type(entry, YAML_SCALAR_NODE);
+        /* We always need at least one colon */
+        if (!g_strrstr(scalar(entry), ":"))
+            return yaml_error(node, error, "Unsupported OVS controller target: %s", scalar(entry));
+
+        vec = g_strsplit (scalar(entry), ":", 2);
+
+        is_host = !g_strcmp0(vec[0], "tcp") || !g_strcmp0(vec[0], "ssl");
+        is_port = !g_strcmp0(vec[0], "ptcp") || !g_strcmp0(vec[0], "pssl");
+        is_unix = !g_strcmp0(vec[0], "unix") || !g_strcmp0(vec[0], "punix");
```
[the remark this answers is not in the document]
slyon: Yeah... I spent lots of time splitting this whole "OVS target parser" up into not-too-confusing pieces.. I'd like to keep it like this for now, to keep it easily understandable.

src/openvswitch.c
slyon: Nope. Fixed!

src/openvswitch.c
slyon: Thanks, that's a good optimization!

slyon:
Thank you very much for your review @sil2100 ! I've integrated you comments.

PR #146, opened 2020-06-16
Fix autopkgtest on arm64 with NM 1.24

slyon:
## Description
On `arm64` we had a test failure in `tests/integration/ethernets.py` -> `test_manual_addresses`, which was caused by the fact that `netplan apply` was disconnecting ALL available network interfaces via `nmcli`, in order to reload the new netplan config. This way the IP of the 2nd test DHCP server (dnsmasq on `veth43`) was lost and did not provide correct responses for the rest of the test.

It is not necessary to take all interfaces down, only the ones which are (or were) configured via netplan. This was fixed in the `netplan apply` command.

**Note:** This should make its way into the Groovy package as a distro-patch, to avoid future autopkgtest failures.

## Checklist

### DIFF
--- a/netplan/cli/utils.py
+++ b/netplan/cli/utils.py
@@ -23,6 +23,8 @@ import fnmatch
 import argparse
 import subprocess
 import netifaces
+import re
+import glob
 
 NM_SERVICE_NAME = 'NetworkManager.service'
 NM_SNAP_SERVICE_NAME = 'snap.network-manager.networkmanager.service'
@@ -53,6 +55,19 @@ def nm_running():  # pragma: nocover (covered in autopkgtest)
         return True
     except (OSError, subprocess.SubprocessError):
         return False
+
+
+def nm_interfaces(paths=glob.glob('/run/NetworkManager/system-connections/netplan-*')):
+    pat = re.compile('^interface-name=(.*)$')
+    interfaces = []
+    for path in paths:
+        with open(path, 'r') as f:
+            for line in f:
+                m = pat.match(line)
+                if m:
+                    interfaces.append(m.group(1))
+                    break  # skip to next file
+    return interfaces
 
 
 def systemctl_network_manager(action, sync=False):  # pragma: nocover (covered in autopkgtest)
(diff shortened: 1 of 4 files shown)


slyon:
This should make its way into the Groovy package as a distro-patch, to avoid future autopkgtest failures.

slyon:
I verified it does not break stuff, by running the full autopkgtest suite on Groovy/arm64, Groovy/amd64 and Bionic/amd64

netplan/cli/utils.py
```diff
@@ -55,6 +57,19 @@ def nm_running():  # pragma: nocover (covered in autopkgtest)
         return False
 
 
+def nm_interfaces(paths=glob.glob('/run/NetworkManager/system-connections/netplan-*')):
```
sil2100, line 60: Ok, so I don't think the default value is correct here. I'd recommend not setting a default value at all.

As it is right now is actually a tricky common Python mistake in most cases! Default values are evaluated and stored once on start of the Python code and never re-ran. This means that the default value of 'paths' will have the state of the `/run/NetworkManager/system-connections/` directory from one particular time, e.g. it will not be reevaluated with consecutive nm_interfaces() calls within the application's life cycle.

This is why I think we should just not have a default value - we don't use that anyway!

netplan/cli/utils.py
```diff
@@ -23,6 +23,8 @@
 import argparse
 import subprocess
 import netifaces
+import re
+import glob
```
sil2100, line 27: If we decide to ditch the possibly-invalid default value for nm_interfaces, we can remove glob from here.

netplan/cli/utils.py
slyon: Ahh, very good catch!
Let's get rid of that default. As you said, we're not using it after all.

Fixed.

netplan/cli/utils.py
slyon: Fixed.

slyon:
ACK'ed by sil2100 on a side-channel.

PR #148, opened 2020-06-26
Base support for OVS fake VLAN bridges.

sil2100:
## Description

The spec defined fake VLAN bridge structures, but wasn't very clear about it being in scope. Let's add those now.

## Checklist


sil2100:
Pushed some basic doc mention - we can improve that later. Thanks!

PR #152, opened 2020-07-07
Ifupdown2 backend

julienfortin:
## Description

add support for a new backend: ifupdown2

https://packages.debian.org/sid/ifupdown2
https://github.com/CumulusNetworks/ifupdown2

Introducing ifupdown2 a network manager using the "ifupdown" syntax.
It creates a "stanza" for each interface. We store the final configuration
in /etc/network/interfaces. An example bond config looks like this:

auto swp42
iface swp42

auto swp84
iface swp84

auto bond42
iface bond42
      bond-slaves swp42 swp84
      bond-mode 802.3ad
      address 10.0.0.42/32

This backend currently supports:
bond, bridges, vlans, tunnels and ethernet interfaces

ifupdown2 supported attributes:
- address
- gateway
- hwaddress
- mtu

- bridge-ports
- bridge-ageing
- bridge-bridgeprio
- bridge-portprios
- bridge-fd
- bridge-hello
- bridge-maxage
- bridge-pathcosts
- bridge-stp

- tunnel-mode
- tunnel-local
- tunnel-endpoint

- vlan-id
- vlan-raw-device

- bond-slaves
- bond-mode
- bond-lacp-rate
- bond-miimon
- bond-min-links
- bond-xmit-hash-policy
- bond-updelay
- bond-downdelay
- bond-num-grat-arp
- bond-primary-reselect
- bond-primary

## Checklist

- [X] \(Optional\) Closes an open bug in Launchpad.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,8 @@ libnetplan.so.$(NETPLAN_SOVER): parse.o util.o validation.o error.o
 	$(CC) -shared -Wl,-soname,libnetplan.so.$(NETPLAN_SOVER) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ `pkg-config --libs yaml-0.1`
 	ln -snf libnetplan.so.$(NETPLAN_SOVER) libnetplan.so
 
-#generate: src/generate.[hc] src/parse.[hc] src/util.[hc] src/networkd.[hc] src/nm.[hc] src/validation.[hc] src/error.[hc]
-generate: libnetplan.so.$(NETPLAN_SOVER) nm.o networkd.o generate.o
+#generate: src/generate.[hc] src/parse.[hc] src/util.[hc] src/networkd.[hc] src/nm.[hc] src/ifupdown2.[hc] src/validation.[hc] src/error.[hc]
+generate: libnetplan.so.$(NETPLAN_SOVER) nm.o networkd.o ifupdown2.o generate.o
 	$(CC) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ -L. -lnetplan `pkg-config --cflags --libs glib-2.0 gio-2.0 yaml-0.1 uuid`
 
 netplan-dbus: src/dbus.c src/_features.h
--- a/examples/bonding.yaml
+++ b/examples/bonding.yaml
@@ -1,6 +1,11 @@
 network:
   version: 2
   renderer: networkd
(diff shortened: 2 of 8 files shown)


vorlonofportland:
Netplan supports both networkd and NetworkManager backends because neither of these supports all the features of the other.

What is the rationale for supporting ifupdown2 as a third backend?  The provided patch doesn't introduce support for any network features that aren't already supported by both of the other two.  This appears to increase maintenance burden, without any advantages to end users.

julienfortin:
> Netplan supports both networkd and NetworkManager backends because neither of these supports all the features of the other.
> 
> What is the rationale for supporting ifupdown2 as a third backend? The provided patch doesn't introduce support for any network features that aren't already supported by both of the other two. This appears to increase maintenance burden, without any advantages to end users.

hi @vorlonofportland 

ifupdown2 is available on both Ubuntu and Debian, we support a wide range of attributes to customize your network configuration. So far we have about 145 different attributes (for bond, vxlan, vlan, bridges, macvlans, vrf, etc...)

We want our users to be able to configure their system with netplan and ifupdown2 if they desire to. In the future we could easily add support for all the networking features that ifupdown2 supports. So far i've only covered what is available today in netplan for NM and networkd.

vorlonofportland:
Thanks for the reply.  Regarding availability of ifupdown2 in Ubuntu, I find that it has been removed prior to the release of Ubuntu 20.04, due to having dependencies on removed python2-only packages.  (https://bugs.launchpad.net/ubuntu/+source/ifupdown2/+bug/1872590)  So as of this moment, it is actually not available in any release of Ubuntu in which this netplan PR would potentially land.

And when it was in Ubuntu (and if it will be again, once ported to python3), it was in universe, implying a much lesser degree of support than for packages in main.  From an Ubuntu product perspective, we would not want to bring the number of network management backends we need to support up from 2 to 3.

You mention 145 attributes supported by ifupdown2.  Have you done a differential analysis of whether any of these are available in ifupdown2 but not in networkd?  Again, given that the intention of netplan is to have a single common language to use in describing network configurations, if there are no features missing from networkd vs ifupdown2, then adding another backend would seem to only increase the maintenance burden.

The other possible argument I could see for netplan supporting ifupdown2 would be if there were images for some OS that contained ifupdown2 by default, and that someone wanted to provision using cloud-init.  Is this a use case you have in mind?

slyon:
Closing for inactivity of more than one year...

PR #153, opened 2020-07-13
Cleanup OVS code and (old) OVS interfaces after 'netplan apply'

slyon:
## Description
This provides general cleanup of the OVS codebase, such as test and comment improvements.

It also makes sure that left-over OVS interfaces (tagged `netplan=true`) are cleaned up during `netplan apply`, by executing `del-br/del-port` on those interfaces. All `ExecStop=` commands and `RemainAfterExit=true` have been removed and all cleanup is now done via the `netplan-ovs-cleanup.service` unit, which runs prior to the other `netplan-ovs-*.service` units.

I'm adding a new parameter `netplan apply --ovs-only` (called via `netplan-ovs-cleanup.service` as well), which will just clean up all `netplan=true` taged OVS interfaces (Bridges/Ports/Bonds), if they are not defined in the current YAML config.

I reduced the usage of `systemd_escape()` even more, to have it only escape interface names, if they are part of an escaped systemd "path" where they actually need to be escaped (such as `sys-subsystem-net-devices-<ESCAPED_ID>.device`). This avoids having normal interface names, such as `br-data` escaped, which leads to a partly escaped systemd service unit name (e.g. `netplan-ovs-br\x2ddata.service`) and unescaping problems when passing those through subprocess calls.

## Checklist

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -57,7 +57,7 @@ class NetplanApply(utils.NetplanCommand):
         # For certain use-cases, we might want to only apply specific configuration.
         # If we only need OpenVSwitch cleanup, do that and exit early.
         if self.ovs_only:
-            NetplanApply.process_ovs_cleanup(config_manager, False, False, True, exit_on_error)
+            NetplanApply.process_ovs_cleanup(config_manager, False, False, exit_on_error)
             return
 
         # if we are inside a snap, then call dbus to run netplan apply instead
@@ -87,7 +87,10 @@ class NetplanApply(utils.NetplanCommand):
                 return
 
         old_files_networkd = bool(glob.glob('/run/systemd/network/*netplan-*'))
-        old_files_ovs = bool(glob.glob('/run/systemd/system/netplan-ovs-*'))
+        old_ovs_glob = glob.glob('/run/systemd/system/netplan-ovs-*')
+        # Ignore netplan-ovs-cleanup.service, as it is always there
+        old_ovs_glob.remove('/run/systemd/system/netplan-ovs-cleanup.service')
+        old_files_ovs = bool(old_ovs_glob)
         old_nm_glob = glob.glob('/run/NetworkManager/system-connections/netplan-*')
         nm_ifaces = utils.nm_interfaces(old_nm_glob)
         old_files_nm = bool(old_nm_glob)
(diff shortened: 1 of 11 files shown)


slyon:
Thank you for your feedback @sil2100 ! I've tried multiple approaches:
- directly via `netplan apply` and `ExecStop=`
- via a systemd cleanup unit, without `ExecStop=`
- via `netplan apply --ovs-only`, without `ExecStop=`

The 3rd approach is similar to how our SR-IOV implementation will work and avoid the systemd ExecStop commands, which is why why I want to stick with this. I verified all integration tests are passing (incl. the new OVS tests).

I'm looking forward for another review.

netplan/cli/commands/apply.py
```diff
@@ -193,7 +211,7 @@ def command_apply(run_generate=True, sync=False, exit_on_error=True):  # pragma:
         # (re)start backends
         if restart_networkd:
             netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa-*.service')]
-            netplan_ovs = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netpalan-ovs-*.service')]
+            netplan_ovs = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-ovs-*.service')]
```
sil2100, line 214: wow, ok, missed that one! Guess this is the reason why `netplan apply` didn't quite work previously!

netplan/cli/ovs.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Query OpenVSwitch state through 'ovs-vsctl' and filter for netplan=true
+    tagged ports/bonds and bridges. Delete interfaces which are not defined
+    in the current configuration.
+    """
+    config_manager.parse()
+
+    # Tear down old OVS interfacess, not defined in the current config
+    if os.path.isfile(OPENVSWITCH_OVS_VSCTL):
+        for t in [['Port', 'del-port'], ['Bridge', 'del-br']]:
```
sil2100, line 35: Super super minor thing: since those are immutable, maybe we should use tuples instead?

netplan/cli/ovs.py
```diff
@@ -0,0 +1,48 @@
...
+            out = subprocess.check_output([OPENVSWITCH_OVS_VSCTL, '--columns=name,external-ids',
+                                           '-f', 'csv', '-d', 'bare', '--no-headings', 'list', t[0]],
+                                          universal_newlines=True)
+            for line in out.split('\n'):
```
sil2100, line 39: Another minor, stylistic thing: maybe `.splitlines()` could be nicer?

netplan/cli/commands/apply.py
```diff
@@ -38,15 +40,26 @@ def __init__(self):
         super().__init__(command_id='apply',
                          description='Apply current netplan config to running system',
                          leaf=True)
+        self.ovs_only = False
 
     def run(self):  # pragma: nocover (covered in autopkgtest)
-        self.func = NetplanApply.command_apply
+        self.parser.add_argument('--ovs-only', action='store_true',
+                                 help='Only clean up old OpenVSwitch interfaces and exit')
```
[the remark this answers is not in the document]
slyon: +1 I was thinking about this as well, but decided to keep it in line with `--sriov-only`. If you have the impression that explicitly having "cleanup" in the name as well, I'm happily changing it!

netplan/cli/commands/apply.py
slyon: No, I think it is more or less redundant to name the services explicitly here, as they are already triggered as a systemd dependencies via `systemd-networkd.service.wants` symlink, created by netplan `generate`.

netplan/cli/ovs.py
slyon: ACK

PR #156, opened 2020-07-28
Fix OVS cleanup and systemd race conditions

slyon:
## Description
* Fix `only_ovs_cleanup` argument, to make it execute the correct code.
* Execute the `oneshot` OVS systemd units synchronously, to avoid race conditions during `systemctl start ...`.
* Set OVS patch ports atomically, to avoid logging errors.
* Clean up OVS fake VLAN bridges from the `Interface` table via `del-br`

## Checklist

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -41,7 +41,7 @@ class NetplanApply(utils.NetplanCommand):
                          description='Apply current netplan config to running system',
                          leaf=True)
         self.sriov_only = False
-        self.ovs_only = False
+        self.only_ovs_cleanup = False
 
     def run(self):  # pragma: nocover (covered in autopkgtest)
         self.parser.add_argument('--sriov-only', action='store_true',
@@ -63,7 +63,7 @@ class NetplanApply(utils.NetplanCommand):
             NetplanApply.process_sriov_config(config_manager, exit_on_error)
             return
         # If we only need OpenVSwitch cleanup, do that and exit early.
-        elif self.ovs_only:
+        elif self.only_ovs_cleanup:
             NetplanApply.process_ovs_cleanup(config_manager, False, False, exit_on_error)
             return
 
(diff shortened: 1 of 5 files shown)


netplan/cli/commands/apply.py
```diff
@@ -41,7 +41,7 @@ def __init__(self):
                          description='Apply current netplan config to running system',
                          leaf=True)
         self.sriov_only = False
-        self.ovs_only = False
+        self.only_ovs_cleanup = False
```
sil2100, line 44: Holy crap, how did I miss this? I think my brain detached the argument name from the variable holding it completely

netplan/cli/commands/apply.py
```diff
@@ -214,7 +214,9 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True):  # p
         if restart_networkd:
             netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa-*.service')]
             netplan_ovs = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-ovs-*.service')]
-            utils.systemctl_networkd('start', sync=sync, extra_services=netplan_wpa + netplan_ovs)
+            # Run 'systemctl start' command synchronously, to avoid race conditions
+            # with 'oneshot' systemd service units, e.g. netplan-ovs-*.service.
+            utils.systemctl_networkd('start', sync=True, extra_services=netplan_wpa + netplan_ovs)
```
sil2100, line 219: I think this now makes sense. Doing this asynchronously in the past was something we could do, but now we have much much more services that can potentially cause issues. +1

tests/integration/ovs.py
```diff
@@ -322,6 +327,14 @@ def test_missing_ovs_tools(self):
         self.assertIn('ovs0: The \'ovs-vsctl\' tool is required to setup OpenVSwitch interfaces.', err)
         self.assertNotEqual(p.returncode, 0)
 
+    @unittest.skip("For debugging only")
+    def test_zzz_ovs_debugging(self):  # Runs as the last test, to collect all logs
```
sil2100, line 331: Convenient!

PR #159, opened 2020-08-27
Flush IPs of NM managed interfaces (LP: #1870561)

slyon:
## Description
There might be scenarios, where leftover IP addresses are assigned to a
given interface. This confuses NetworkManager and makes it create a new,
non netplan-* connection profile, using this leftover IP.

Flush the IP addresses of all NM managed interfaces, to avoid this
situation.

Fixes LP: #1870561

This PR passes all unit- and integration-test and the manual testing.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/netplan/+bug/1870561

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -86,7 +86,7 @@ class NetplanApply(utils.NetplanCommand):
 
         old_files_networkd = bool(glob.glob('/run/systemd/network/*netplan-*'))
         old_nm_glob = glob.glob('/run/NetworkManager/system-connections/netplan-*')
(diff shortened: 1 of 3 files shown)


sil2100 (approved):
Looking good! Also, good catch with us not doing proper globbing in `nm_interfaces`. I assume that us flushing the IPs before restarting NM will not actually have any bad side-effects.

slyon:
Thanks! I'll merge this. I ran all relevant `autopkgtests` and it did not show any side-effects.
Well... actually, the `nm_interfaces` globbing problem surfaced during that testing, so I fixed it.

PR #160, opened 2020-08-27
Add per-route MTU option (LP: #1860201)

hrasiq:
This patch introduces a way to set MTU for specific routes. It adds an 'mtu'
field to the `NetplanIPRoute` struct, and makes use of already existing route
handling code to add the MTU-related fields for networkd and NM backends.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. ([LP: #1860201](https://bugs.launchpad.net/netplan/+bug/1860201))

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1517,6 +1517,7 @@ static const mapping_entry_handler routes_handlers[] = {
     {"type", YAML_SCALAR_NODE, handle_routes_type},
     {"via", YAML_SCALAR_NODE, handle_routes_ip, NULL, route_offset(via)},
     {"metric", YAML_SCALAR_NODE, handle_routes_guint, NULL, route_offset(metric)},
+    {"mtu", YAML_SCALAR_NODE, handle_routes_guint, NULL, route_offset(mtubytes)},
     {NULL}
 };
 
--- a/tests/generator/test_errors.py
+++ b/tests/generator/test_errors.py
@@ -528,6 +528,21 @@ class TestConfigErrors(TestBase):
         - 192.168.14.2/24
         - 2001:FFfe::1/64''', expect_fail=True)
 
+    def test_device_bad_route_mtu(self):
+        err = self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      routes:
+        - to: 10.10.0.0/16
+          via: 10.1.1.1
+          mtu: -1
+      addresses:
+        - 192.168.14.2/24
+        - 2001:FFfe::1/64''', expect_fail=True)
+
+        self.assertIn("invalid unsigned int value '-1'", err)
+
     def test_device_route_family_mismatch_ipv6_to(self):
         self.generate('''network:
   version: 2
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -574,6 +574,10 @@ These options are available for all types of interfaces.
           see ``/etc/iproute2/rt_tables``.
           (``NetworkManager``: as of v1.10.0)
 
+    ``mtu`` (scalar) – since **0.101**
+     :    The MTU to be used for the route, in bytes. Must be a positive integer
+          value.
+
 ``routing-policy`` (mapping)
 
 :    The ``routing-policy`` block defines extra routing policy for a network,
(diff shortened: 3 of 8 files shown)


src/parse.c
```diff
@@ -1423,6 +1437,7 @@ static const mapping_entry_handler routes_handlers[] = {
     {"type", YAML_SCALAR_NODE, handle_routes_type},
     {"via", YAML_SCALAR_NODE, handle_routes_ip, NULL, route_offset(via)},
     {"metric", YAML_SCALAR_NODE, handle_routes_metric},
+    {"mtu", YAML_SCALAR_NODE, handle_routes_mtu},
```
slyon, line 1440:
```suggestion
    {"mtu", YAML_SCALAR_NODE, handle_routes_guint, NULL, route_offset(mtubytes)},
```
We should be making use of a more generic `handle_routes_guint` function here, to avoid duplication.
Actually, we could/should use a more generic approach for `handle_routes_metric` and `handle_routes_table` as well...

tests/generator/test_errors.py
```diff
@@ -509,6 +509,19 @@ def test_device_bad_route_metric(self):
         - 192.168.14.2/24
         - 2001:FFfe::1/64''', expect_fail=True)
 
+    def test_device_bad_route_mtu(self):
+        self.generate('''network:
```
slyon, line 513:
```suggestion
        err = self.generate('''network:
```

tests/generator/test_errors.py
```diff
@@ -509,6 +509,19 @@ def test_device_bad_route_metric(self):
...
+  version: 2
+  ethernets:
+    engreen:
+      routes:
+        - to: 10.10.0.0/16
+          via: 10.1.1.1
+          mtu: -1
+      addresses:
+        - 192.168.14.2/24
+        - 2001:FFfe::1/64''', expect_fail=True)
```
slyon, line 523:
```suggestion
        - 2001:FFfe::1/64''', expect_fail=True)
        self.assertIn("invalid unsigned int value '-1'", err)
```
Could you please verify the explicit error message here?

hrasiq:
Thanks for the review, @slyon!

On the NM `mtu` key: I didn't find explicit mention of it, just entries like `NM_IP_ROUTE_ATTRIBUTE_MTU` [0]. This seems to be the same for the `onlink` key, and I tried following that for MTU.

I've been doing more tests with those on Ubuntu Bionic and Focal, and even if NetworkManager itself seems to pick up the `mtu` key (it shows up in `nmcli connection show <con>` as an attribute under `ipv4.routes`), it doesn't seem to be reflected in the actual routes (i.e. `ip route` doesn't pick up the new MTU values). I'm a bit confused as to why, but it does seem like something in NM is not passing those correctly down to the actual routes.

I'll implement the changes you've requested, and also give the integration test a try. I'm thinking of removing the MTU sections from the NM backend and just error out early, since I haven't been able to get it to work in my tests. Do you have any other suggestions, or does that make sense?

[0] https://developer.gnome.org/libnm/stable/NMSettingIPConfig.html#NM-IP-ROUTE-ATTRIBUTE-MTU:CAPS

slyon:
Thank you @hrasiq 

Okay... if the MTU shows up in nmcli it must be recognized by NM. I'm not sure how NM is supposed to pass that information down to the routing tables... I found this reference talking about MTUs in the routing cache, maybe you could test if it is showing up there: http://linux-ip.net/html/tools-ip-route.html Or you could try using wireshark/tcpdump to verify what the actual MTU size is on that route, once you configured it via NM...

For the integration test it would be great if you could find a way to verify the MTU is actually passed through to `ip route`. If not, I would not want to drop the `mtu` attribute, though, as it **is picked up by NM**. So if you cannot find a way to verify it via `ip route` (i.e. this might be a bug in NM passing it down to the routing table...), you could at least verify that NM picked it up via the `ipv4.routes` "mtu" attribute in the integration test. Once the (potential) bug is fixed in NM, there are no changes needed on our side.

hrasiq:
@slyon Apologies for the long delays on this one. I've sorted out the NetworkManager issues on picking up the route MTU values, and it seems to be working fine now.

I've addressed the changes you've requested, and added an integration test as well. It passed correctly on my tests with Ubuntu Bionic and Focal, but please let me know if you'd like any changes there.

Thanks again for the review!

slyon (approved):
Thank you very much for your continued work @hrasiq 
This is looking very good!

Let me run a full set of integration test on this, before I merge it.

PR #161, opened 2020-08-28
Implement ipv6-address-token-id key (LP: #1737976)

slyon:
## Description
This allows to statically configure the IPv6 host ID (low 64 bits) when auto-generated IPv6 addressing is used (i.e. DHCPv6 stateless, SLAAC).

It introduces one new YAML key in the schema: `ipv6-address-token-id`.

Fixes LP: #1737976

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/ubuntu/+source/nplan/+bug/1737976

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -272,11 +272,11 @@ Virtual devices
     Stateless Address Autoconfiguration (only supported with `NetworkManager`
     backend). Possible values are ``eui64`` or ``stable-privacy``.
 
-``ipv6-address-token-id`` (scalar) – since **0.100**
+``ipv6-address-token`` (scalar) – since **0.100**
 
 :   Define an IPv6 address token for creating a static interface identifier for
-    IPv6 Stateless Address Autoconfiguration. This implies a slightly modified
-    ``eui64`` mode and is mutually exclusive with ``ipv6-address-generation``.
+    IPv6 Stateless Address Autoconfiguration. This is mutually exclusive with
+    ``ipv6-address-generation``.
 
 ``gateway4``, ``gateway6`` (scalar)
 
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -272,6 +272,12 @@ Virtual devices
     Stateless Address Autoconfiguration (only supported with `NetworkManager`
     backend). Possible values are ``eui64`` or ``stable-privacy``.
 
+``ipv6-address-token-id`` (scalar) – since **0.100**
+
+:   Define an IPv6 address token for creating a static interface identifier for
+    IPv6 Stateless Address Autoconfiguration. This implies a slightly modified
+    ``eui64`` mode and is mutually exclusive with ``ipv6-address-generation``.
+
 ``gateway4``, ``gateway6`` (scalar)
 
 :   Set default gateway for IPv4/6, for manual address configuration. This
--- a/src/parse.c
+++ b/src/parse.c
@@ -1744,7 +1744,7 @@ static const mapping_entry_handler nameservers_handlers[] = {
     {"route-metric", YAML_SCALAR_NODE, handle_netdef_guint, NULL, netdef_offset(overrides.metric)},         \
     {"send-hostname", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.send_hostname)},  \
     {"use-dns", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_dns)},              \
-    {"use-domains", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(overrides.use_domains)},       \
+    {"use-domains", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(overrides.use_domains)},      \
     {"use-hostname", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_hostname)},    \
     {"use-mtu", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_mtu)},              \
     {"use-ntp", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_ntp)},              \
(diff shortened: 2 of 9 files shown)


slyon:
Hey @vorlonofportland I'd like to get your opinion on the the schema key wording: `ipv6-address-token-id`
I wasn't really sure how to name it... what do you think about it?

cluonbeam:
You could drop the `id` portion of the schema key. The noun in this case is `token`; there is no other identifier. 

`ipv6-address-token`

slyon:
> You could drop the `id` portion of the schema key. The noun in this case is `token`; there is no other identifier.
> 
> `ipv6-address-token`

Yeah... I was thinking about this as well... but this token is basically the "IPv6 interface identifier", so I wanted to add the `-id` part.
Maybe `ipv6-address-interface-id` would be a good name?

cluonbeam:
That's actually not what a token is. A token allows the administrator to set the host portion of the address to a static value even though the rest of the configuration is some form of autoconfiguration. Also, interface ID is **very ** ambiguous with other configuration elements. The only truly accurate and unambiguous name for this is `ipv6-address-token`

src/parse.c
```diff
@@ -1735,7 +1746,7 @@ static const mapping_entry_handler nameservers_handlers[] = {
     {"route-metric", YAML_SCALAR_NODE, handle_netdef_guint, NULL, netdef_offset(overrides.metric)},         \
     {"send-hostname", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.send_hostname)},  \
     {"use-dns", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(overrides.use_dns)},              \
-    {"use-domains", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(overrides.use_domains)},      \
+    {"use-domains", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(overrides.use_domains)},       \
```
vorlonofportland, line 1749: correct change, but unrelated?

src/parse.c
```diff
@@ -1780,11 +1792,11 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
...
-#define PHYSICAL_LINK_HANDLERS                                                           \
-    {"match", YAML_MAPPING_NODE, handle_match},                                          \
-    {"set-name", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(set_name)},    \
-    {"wakeonlan", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(wake_on_lan)}, \
-    {"wakeonwlan", YAML_SEQUENCE_NODE, handle_wowlan, NULL, netdef_offset(wowlan)},       \
+#define PHYSICAL_LINK_HANDLERS                                                                \
+    {"match", YAML_MAPPING_NODE, handle_match},                                               \
+    {"set-name", YAML_SCALAR_NODE, handle_netdef_str, NULL, netdef_offset(set_name)},         \
+    {"wakeonlan", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(wake_on_lan)},    \
+    {"wakeonwlan", YAML_SEQUENCE_NODE, handle_wowlan, NULL, netdef_offset(wowlan)},           \
```
vorlonofportland, line 1799: more unrelated whitespace changes

src/validation.c
```diff
@@ -195,6 +195,9 @@ validate_netdef_grammar(NetplanNetDefinition* nd, yaml_node_t* node, GError** er
             goto netdef_grammar_error;
     }
 
+    if (nd->ip6_addr_gen_mode > NETPLAN_ADDRGEN_EUI64 && nd->ip6_addr_gen_token)
+        return yaml_error(node, error, "%s: ipv6-address-generation and ipv6-address-token-id are mutually exclusive", nd->id);
+
```
vorlonofportland, line 200: I'm uncomfortable with this being an inequality comparison. If other values are added to NetplanAddrGenMode in the future, what guarantees these checks are correct?

doc/netplan.md
```diff
@@ -272,6 +272,12 @@ Virtual devices
     Stateless Address Autoconfiguration (only supported with `NetworkManager`
     backend). Possible values are ``eui64`` or ``stable-privacy``.
 
+``ipv6-address-token-id`` (scalar) – since **0.100**
+
+:   Define an IPv6 address token for creating a static interface identifier for
+    IPv6 Stateless Address Autoconfiguration. This implies a slightly modified
+    ``eui64`` mode and is mutually exclusive with ``ipv6-address-generation``.
+
```
vorlonofportland, line 280: I agree with the other PR comments: "token id" is redundant.

I also think that "token" itself is unclear; and at least in networkd, token is used to mean something other than just the id.

https://tools.ietf.org/html/rfc4862 mentions in Appendix B that they /used to/ use the term "interface token" but now say "interface identifier".

At https://developer.gnome.org/NetworkManager/stable/settings-ipv6.html, NetworkManager references draft-chown-6man-tokenised-ipv6-identifiers-02 for its definition of "token". But this also refers to "tokenized interface identifier" - so "interface identifier" is used consistently.

Therefore I recommend "ipv6-interface-identifier" as the key for this.

cluonbeam: The term used in iproute2, NetworkManager, systemd, etc. is "token." The token is **not** an interface identifier; labeling it as such is extremely misleading. The accurate key for this is, as stated, `ipv6-address-token`.

vorlonofportland: I agree it's important that this be discoverable for people who are looking for "token", given that this term is used throughout iproute2, NetworkManager, and networkd as you mention (though, not with a consistent meaning).

Given that the term 'token' is deprecated in the RFCs, however, I don't like `ipv6-token-id` as this propagates the idea of a 'token' as something that new users are expected to understand.

My suggestions are either to use the `ipv6-interface-identifier` and reference 'token' in the documentation; or use `ipv6-address-token`.

src/parse.c
slyon: Yeah... unrelated. I removed them and will prepare another PR for cleaning this up.

src/parse.c
slyon: I removed them and will prepare another PR for cleaning this up.

src/validation.c
slyon: Good point! I switched to a `!= NETPLAN_ADDRGEN_DEFAULT` comparison, which makes it more explicit to fail if `ipv6-address-generation` is set in any way.

doc/netplan.md
slyon: Thank you very much @cluonbeam and @vorlonofportland for your input!
I'm going with `ipv6-address-token` then, to keep it in line with the related `ipv6-address-generation` setting.

sil2100 (approved):
Ok, this looks good. I see all Steve's review comments have also been addressed, so I think this is good to go.

PR #165, opened 2020-10-09
Fix MAAS/OVS first boot for single NIC/PXE systems

slyon:
## Description
Netplan's Open vSwitch helper units are started too late, which makes `systemd-networkd-wait-online.service` fail the first time it is started. This leads to `cloud-init init` running in an invalid state, thus failing, which in turn makes the system (deployed via MAAS) end up with broken networking.

The helper units only depend on OVSDB, which is started in `network-pre.target`, thus we can make them depend on `ovsdb-server.service` instead of `openvswitch-switch.service`, to avoid failures in `wait-online`.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/cloud-init/+bug/1898997

### DIFF
--- a/src/openvswitch.c
+++ b/src/openvswitch.c
@@ -38,8 +38,9 @@ write_ovs_systemd_unit(const char* id, const GString* cmds, const char* rootdir,
     g_string_append_printf(s, "Description=OpenVSwitch configuration for %s\n", id);
     g_string_append(s, "DefaultDependencies=no\n");
     /* run any ovs-netplan unit only after openvswitch-switch.service is ready */
-    g_string_append_printf(s, "Wants=openvswitch-switch.service\n");
-    g_string_append_printf(s, "After=openvswitch-switch.service\n");
+    g_string_append_printf(s, "Wants=ovsdb-server.service\n");
+    g_string_append_printf(s, "After=ovsdb-server.service\n");
+    g_string_append_printf(s, "Before=systemd-networkd.service\n");
     if (physical) {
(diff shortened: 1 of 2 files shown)


xnox:
In theory, systemd-networkd monitors netlink and should respond well to dynamically added/removed/changed interfaces. But it doesn't hurt to start before it.

Note though, some of the commands will fail, if one expects to add something to the ovs-bridge, which is to be created by systemd-networkd for example a regular vlan, veth pair, dummy0 device, tunnel, wireguard connection, etc. So there might be a dependency loop.

I wonder if Before=systemd-netwokrd bit can be dropped, or moved elsewhere. Such that ovsdb-server.service itself is Before=systemd-networkd.service, rather than our executor helpers.

slyon:
Yes, `Before=systemd-networkd.service` can be dropped and it is still working as expected.

Currently ovsdb-server.service contains `Before=network.target networking.service` (so before ENI/ifupdown but not before systemd-networkd) and I assume `Before=systemd-networkd.service` is not needed, due to systemd-networkd's reactive behavior on changed interfaces. So I think it is OK to just drop that part.

My experiments showed that OVSDB will happily accept configurations of non-existing interfaces and activate them once the interface appears. So I don't think stuff will fail, if for example we connect an OVS bridge to a Wireguard interface created afterwards by systemd-networkd.

xnox (approved):

slyon:
All integration tests passed. Ready for merging.

PR #166, opened 2020-10-22
Do not try to match/rename duplicated MAC of vlan (LP: #1888726)

slyon:
## Description
Fix systemd-networkd misconfiguration; when matching on macaddr, if other interfaces in the system might have the same mac (e.g. bridge, vlan, etc), the match filtering must also filter which interface to match, e.g., besides the MACAddress= match it also needs (for example) one or more of:

* Type=!vlan
* Type=!bridge

PS: also cleaning up the `test_vlans.py` file a bit, while working on it.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1888726

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -86,19 +86,18 @@ append_match_section(const NetplanNetDefinition* def, GString* s, gboolean match
             g_string_append_printf(s, "Name=%s\n", def->match.original_name);
     }
 
-    /* Workaround for bug LP: #1804861: something outputs netplan config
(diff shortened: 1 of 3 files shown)


slyon:
Integration tests all passed.

sil2100:
Looking at it today!

sil2100 (approved):
Looks good, thank you for this! Guess we forgot about that list there.

PR #171, opened 2020-11-06
Add support for the activation-mode parameter (LP: #1664844)

sil2100:
## Description

Per an existing specification, add the activation-mode parameter which currently supports the 'off' and 'manual' modes. This allows administrators either force a given interface to be always in the down state ('off', only for networkd right now) or allow the user to manually manage the state ('manual'). This means that a 'manual' interface will only be up/down if the user does that explicitly.

The main target is the networkd backend, but we try supporting at least the `manual` setting for NetworkManager. The networkd support for this has now been merged into v248 (which is still not released as of this writing).

https://bugs.launchpad.net/netplan/+bug/1664844

## Checklist


sil2100:
Ok, so I think this old branch should be good for review now. Not sure if we should merge it already as the feature is only available in the upcoming v248, but a review of both the code and schema would be welcome!

src/nm.c
```diff
@@ -530,6 +530,17 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
         g_string_append_printf(s, "uuid=%s\n", uuidstr);
     }
 
+    if (def->activation_mode) {
+        /* XXX: For now NetworkManager only supports the "manual" activation
+         * mode */
+        if (g_strcmp0(def->activation_mode, "off") == 0) {
+            g_fprintf(stderr, "ERROR: %s: NetworkManager definitions do not support activation_mode off\n", def->id);
```
slyon, line 537: I'd rather like to check for the supported feature (i.e. whitelist) here, and fail for all others, instead of the other way around, to avoid unexpected behavior once new values are added to "activation-mode". Also I'd mention the setting name (activation-mode) vs variable name (activation_mode) in the error message.
```suggestion
        if (!!g_strcmp0(def->activation_mode, "manual")) {
            g_fprintf(stderr, "ERROR: %s: NetworkManager definitions do not support activation-mode %s\n", def->id, def->activation_mode);
```

doc/netplan.md
```diff
@@ -437,6 +437,24 @@ similar to ``gateway*``, and ``search:`` is a list of search domains.
             dhcp6: true
             optional-addresses: [ ipv4-ll, dhcp6 ]
 
+``activation-mode`` (scalar) – since **0.102**
+
+:    Allows specifying the management policy of the selected interface. By
+     default, netplan brings up any configured interface if possible. Using the
+     ``activation-mode`` setting users can override that behavior by either
+     specifying "manual", to hand over control over the interface state to the
+     administrator or (for networkd backend *only*) "off" to force the link
```
slyon, line 446: Just some formatting here, I think we usually emphasize the values as a code-style block.

```suggestion
     specifying ``manual``, to hand over control over the interface state to the
     administrator or (for networkd backend *only*) ``off`` to force the link
```

src/networkd.c
```diff
@@ -566,6 +566,16 @@ write_network_file(const NetplanNetDefinition* def, const char* rootdir, const c
         }
     }
 
+    /* The ActivationPolicy setting is available in systemd v248+ */
+    if (def->activation_mode) {
+        char* mode;
```
slyon, line 571: We're assigning constant string to this pointer, so I'd suggest to mark it as such.
```suggestion
        const char* mode;
```

doc/netplan.md
```diff
@@ -437,6 +437,24 @@ similar to ``gateway*``, and ``search:`` is a list of search domains.
             dhcp6: true
             optional-addresses: [ ipv4-ll, dhcp6 ]
 
+``activation-mode`` (scalar) – since **0.102**
```
slyon, line 440: Depending on when we merge this PR, we might need to remember to update the version string.

slyon (approved):
Thank you very much Łukasz, this feature has been requested many times already!
I think we should mention the LP bug in the commit title (https://bugs.launchpad.net/netplan/+bug/1664844), to make sure we can put it in our release notes and notify all waiting people about this update.

Schema looks nice and simple to me, but I'd leave the final word to @vorlonofportland 

Code-wise I added some tiny notes and one other "blacklist vs whitelist" points, which I'd like to see addressed. I'm fine with merging once those are resolved.

sil2100:
Thank you for the review! Actually I didn't know about the whole suggestion mechanism on github, that's quite sweet! Applied all of them - they made sense to me! I won't rebase the branch yet per master - I'll do that right before we decide to merge it (once the new systemd is out).

slyon:
Would it be possible to implement the "off" state for the NetworkManager renderer by using the `keyfile.unmanaged-devices` setting, as discussed here: https://bugs.launchpad.net/cloud-init/+bug/1914784 ?

sil2100:
hm, I don't think so. The "off" state in this case means "interface forced into a down state at all times", where at least from a quick glance `keyfile.unmanaged-devices` is more of a thing "hey, don't manage this interface at all". Or am I mixing something up?

slyon:
No.. I think you're right. The behavior is different, although it might look the same in a common scenario (system booted with neither NM nor networkd managing the interface => DOWN). But if any other network management comes into play and brings up the interface it will behave in an unexpected way. So it's better to keep this PR as is.

JeffHochberg:
I had opened a bug request (1916062) about this functionality thinking it was a documentation/support issue, but I just reviewed this and it seems to accomplish exactly what I'm looking for. I hope that you're able to add this soon. It would be very helpful!

doc/netplan.md
slyon, line 440: This is now going to be v0.103
```suggestion
``activation-mode`` (scalar) – since **0.103**
```

Additionally, we might choose a more generic name here, like `interface-mode`, in order to support an additional mode: `ignore`, where netplan should not touch the interface at all, but leave it up to the administrator to configure that interface manually. ([LP: #1918198](https://bugs.launchpad.net/netplan/+bug/1918198))

slyon (commented):
I've added an additional comment inline, to possibly rename this new setting in order to support a broader usecase (cf. [LP: #1918198)](https://bugs.launchpad.net/netplan/+bug/1918198)

slyon:
Hmm.. according to our reference [0] "Devices not covered by the network config do not get touched at all."

So maybe this should even be the default, and not a new, explicit "ignore" mode? But IMO it would then interfere with the default experience on desktop systems, when netplan is installed, changing NetworkManager's defaults.

src/parse.h
```diff
@@ -237,6 +237,8 @@ struct net_definition {
     gboolean optional;
     NetplanOptionalAddressFlag optional_addresses;
     gboolean critical;
+    /* netplan-feature: activation-mode */
+    char* activation_mode;
```
slyon, line 241: Please move this to the very end of the `net_definition` struct, otherwise we might break the ABI compatibility (again).

slyon (changes requested):
In lights of the current netplan v0.102 regression problems, double checking this PR, I've found an ABI compatibility issue that needs fixing. See inline comment.

src/parse.h
sil2100: Yes! Done!

src/nm.c
```diff
@@ -596,6 +596,17 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
...
 
+    if (def->activation_mode) {
+        /* XXX: For now NetworkManager only supports the "manual" activation
+         * mode */
+        if (!!g_strcmp0(def->activation_mode, "manual")) {
+            g_fprintf(stderr, "ERROR: %s: NetworkManager definitions do not support activation-mode %s\n", def->id, def->activation_mode);
+            exit(1);
+        }
+        /* "manual" */
+        g_string_append(s, "autoconnect=false\n");
```
slyon, line 607: The recent merging of `master` introduced a build failure here:
```suggestion
        g_key_file_set_boolean(kf, "connection", "autoconnect", FALSE);
```

slyon (changes requested):
Thank you for that changes. It LGTM, but needs one small (but important) change, which was introduced by merging the `master` branch. The `g_string_append(s, ...)` is no more, as we're now using a `GKeyFile` data structure – that introduces a build failure.

Also, I'm still pondering if we should rename `activation-mode` to `interface-mode`, in order to allow supporting future use cases, such as `interface-mode: ignore` (see: https://bugs.launchpad.net/netplan/+bug/1918198).

slyon (approved):
All good from a code POV now, thank you!
We need to make sure to keep `char* activation_mode;` at the very end of the `NetplanNetDefinition` struct when merging.

Pending `interface-mode` renaming decision.

sil2100:
Hey Lukas! Do you think we can get this merged already, so that it's staged in master for 0.103? Or do you want to wait for the `interface-mode` discussion to finish? I'd personally want to merge it as-is and then still resolve it before going out with 0.103, but I think I like activation-mode as is right now (and would simply add a separate schema item). Thanks!

slyon (approved):
Yes. I don't have a strong opinion about that, so I'm fine with that.
I also double-checked the merge of the `netdef` struct from `origin/master` and it is looking good! :+1: 

So let's just merge it!

PR #172, opened 2020-11-18
Fix is_ovs_interface method in cli/ovs.py (LP: #1904633) (LP: #1905156)

iaroslavscript:
## Description
Hi! This is a fix for [1904633](https://bugs.launchpad.net/netplan/+bug/1904633) **netplan** crashes while loading incorrect config. 

## Testcase

Bellow is a minimal reproducible config. It wasn't my aim to create a malfunction config I just thought that it's possible to define a `renderer` for a subgroup of interfaces.

```
root@pc:/# cat /etc/netplan/01-network-manager-all.yaml
network:
  version: 2
  wifis:
    renderer: NetworkManager
```
```
root@pc:/# netplan try

An error occurred: 'str' object has no attribute 'get'

Reverting.
  File "/usr/share/netplan/netplan/cli/commands/try_command.py", line 83, in command_try
    NetplanApply().command_apply(run_generate=True, sync=True, exit_on_error=False)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 154, in command_apply
    NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 324, in process_ovs_cleanup
    apply_ovs_cleanup(config_manager, ovs_old, ovs_current)
  File "/usr/share/netplan/netplan/cli/ovs.py", line 119, in apply_ovs_cleanup
    if (is_ovs_interface(i, config_manager.interfaces)):
  File "/usr/share/netplan/netplan/cli/ovs.py", line 102, in is_ovs_interface
    if interfaces[iface].get('openvswitch') is not None:
AttributeError: 'str' object has no attribute 'get'

During handling of the above exception, another exception occurred:

    netplan.main()
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 211, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/try_command.py", line 66, in run
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 211, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/try_command.py", line 94, in command_try
    self.revert()
  File "/usr/share/netplan/netplan/cli/commands/try_command.py", line 117, in revert
    NetplanApply().command_apply(run_generate=False, sync=True, exit_on_error=False)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 154, in command_apply
    NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 324, in process_ovs_cleanup
    apply_ovs_cleanup(config_manager, ovs_old, ovs_current)
  File "/usr/share/netplan/netplan/cli/ovs.py", line 119, in apply_ovs_cleanup
    if (is_ovs_interface(i, config_manager.interfaces)):
  File "/usr/share/netplan/netplan/cli/ovs.py", line 102, in is_ovs_interface
    if interfaces[iface].get('openvswitch') is not None:
AttributeError: 'str' object has no attribute 'get'
root@pc:/#

```

The exception happens because *interface* supposed to be a `dict` not any other type.

I've found a comment that it's planning to get rid of parsing configs in python [configmanager.py#L93](https://github.com/CanonicalLtd/netplan/blob/0.100/netplan/configmanager.py#L93) So in this pull request a made only a quick fix.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: [LP: #1904633](https://pad.lv/1904633) [LP: #1905156](https://pad.lv/1905156)

### DIFF
--- a/netplan/cli/ovs.py
+++ b/netplan/cli/ovs.py
@@ -99,10 +99,14 @@ def clear_setting(type, iface, setting, value):
 
 
 def is_ovs_interface(iface, interfaces):
-    if interfaces[iface].get('openvswitch') is not None:
+    assert isinstance(interfaces, dict)
+    if not isinstance(interfaces.get(iface), dict):
+        logging.debug('Ignoring special key: {} ({})'.format(iface, interfaces.get(iface)))
+        return False
+    elif interfaces.get(iface, {}).get('openvswitch') is not None:
         return True
     else:
-        return any(is_ovs_interface(i, interfaces) for i in interfaces[iface].get('interfaces', []))
+        return any(is_ovs_interface(i, interfaces) for i in interfaces.get(iface, {}).get('interfaces', []))
 
 
 def apply_ovs_cleanup(config_manager, ovs_old, ovs_current):  # pragma: nocover (covered in autopkgtest)
--- a/netplan/configmanager.py
+++ b/netplan/configmanager.py
@@ -17,7 +17,6 @@
 
 '''netplan configuration manager'''
 
-import collections.abc
 import glob
 import logging
 import os
@@ -235,7 +234,6 @@ class ConfigManager(object):
 
         for ifname in changed_ifaces:
             iface = ports.pop(ifname)
-            self._validate_interface_config(ifname, iface)
             if ifname in orig:
                 logging.debug("{} exists in {}".format(ifname, orig))
                 orig[ifname].update(iface)
@@ -252,7 +250,6 @@ class ConfigManager(object):
 
         for ifname in changed_ifaces:
             iface = new.pop(ifname)
-            self._validate_interface_config(ifname, iface)
             if ifname in orig:
                 logging.debug("{} exists in {}".format(ifname, orig))
                 orig[ifname].update(iface)
@@ -303,24 +300,10 @@ class ConfigManager(object):
                     if 'renderer' in network:
                         self.network['renderer'] = network.get('renderer')
             return new_interfaces
-        except ConfigurationError as e:  # pragma: nocover (validation YAML)
-            logging.error('Validation config {} failed, aborting. {}'.format(yaml_file, e))
-            sys.exit(1)
         except (IOError, yaml.YAMLError):  # pragma: nocover (filesystem failures/invalid YAML)
             logging.error('Error while loading {}, aborting.'.format(yaml_file))
             sys.exit(1)
-        except Exception as e:  # pragma: nocover (unknown error)
-            logging.debug('Exception raised while loading {}. {0!r}'.format(yaml_file, e))
-            logging.error('Unknown error while loading {}, aborting.'.format(yaml_file))
-            sys.exit(1)
 
-    def _validate_interface_config(self, ifname, iface):
-        """
-        The function do a minimal checking of validity iface.
-        """
-        if not isinstance(iface, collections.abc.Mapping):
-            logging.debug('Config validation error - invalid interface {}'.format(ifname))
-            raise ConfigurationError('Interface {} invalide type {}'.format(ifname, type(iface)))
 
 class ConfigurationError(Exception):
     """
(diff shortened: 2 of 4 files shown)


slyon:
Hey Iaroslav, thank you very much for reporting this bug and starting to work on it right away!
I can reproduce the problem on my machine.

Your proposed change looks pretty big, though. I think a simple change in `netplan/cli/ovs.py` should be enough to fix the root cause of this problem (in line 102). Something like this:
```
diff --git a/netplan/cli/ovs.py b/netplan/cli/ovs.py
index a7ae6c8..ab3cb4c 100644
--- a/netplan/cli/ovs.py
+++ b/netplan/cli/ovs.py
@@ -99,7 +99,9 @@ def clear_setting(type, iface, setting, value):
 
 
 def is_ovs_interface(iface, interfaces):
-    if interfaces[iface].get('openvswitch') is not None:
+    if not isinstance(interfaces[iface], dict):
+        return False
+    elif interfaces[iface].get('openvswitch') is not None:
         return True
     else:
         return any(is_ovs_interface(i, interfaces) for i in interfaces[iface].get('interfaces', []))
```

Also, adding a small test case in `tests/generator/test_ovs.py` (using your minimal reproducer) to make tests & coverage happy would be great.
Would you be willing to re-work this PR accordingly?

slyon (changes requested):
Hey Iaroslav, thank you very much for reporting this bug and starting to work on it right away!
I can reproduce the problem on my machine.

Your proposed change looks pretty big, though. I think a simple change in `netplan/cli/ovs.py` should be enough to fix the root cause of this problem (in line 102). Something like this:
```
diff --git a/netplan/cli/ovs.py b/netplan/cli/ovs.py
index a7ae6c8..ab3cb4c 100644
--- a/netplan/cli/ovs.py
+++ b/netplan/cli/ovs.py
@@ -99,7 +99,9 @@ def clear_setting(type, iface, setting, value):
 
 
 def is_ovs_interface(iface, interfaces):
-    if interfaces[iface].get('openvswitch') is not None:
+    if not isinstance(interfaces[iface], dict):
+        return False
+    elif interfaces[iface].get('openvswitch') is not None:
         return True
     else:
         return any(is_ovs_interface(i, interfaces) for i in interfaces[iface].get('interfaces', []))
```

Also, adding a small test case in `tests/generator/test_ovs.py` (using your minimal reproducer) to make tests & coverage happy would be great.
Would you be willing to re-work this PR accordingly?

iaroslavscript:
Hi @slyon,
Thanks for the review.
Sure. I will rework!

slyon (approved):
Hey @iaroslavscript sorry for cutting the line here and pushing my own changes.
There is another bug related to a very similar issue ([LP: #1905156](https://pad.lv/1905156)) and we're working on a new release where we want to have those fixes included.

Thank you very much for your contribution, test case and minimal reproducer. This is much appreciated! Also, let me know what you think about my proposed changes in this PR.

Maybe @sil2100 could have another look as well?

sil2100 (approved):
Looks correct. Indeed this was something we missed considering during the OVS code addition. Thank you Iaroslav, Lukas!

iaroslavscript:
Hi @slyon 

I'm very glad this case is closed and the problem is solved. Thanks for taking care about it, I was a little bit busy on my regular job that week.

LGFM! The only thing I worry about is that the warning of skipped key is displayed only in debug logging mode. You can miss it when applying configuration until you actually verify the result. I would propose to display it always.

iaroslavscript:
@slyon @sil2100 :+1:

marcoBros:
How can we apply this patch whitout waiting for package distribution via apt ?

slyon:
> @iaroslavscript
> How can we apply this patch whitout waiting for package distribution via apt ?

@marcoBros you could modify the file on disk accordingly (`/usr/share/netplan/netplan/cli/ovs.py`) but this is not recommended – do it at your own risk.

PR #174, opened 2020-11-24
Fix rename of matched interfaces at runtime (LP: #1904662)

slyon:
## Description
Renaming of network interfaces was not working when running `netplan apply` and interfaces would only be renamed after a reboot via udev. There is some udev-rename logic inside `cli/commands/apply.py`, which doesn't seem to be working, though.

This seems to be related to systemd-udevd's `NamePolicy=keep` default, which is set explicitly as of systemd v240 [0][1] and prevents renaming of interfaces via udev, if the name was set once (i.e. during boot). Even before v240 this has be the (implicit) default, so I'm not sure if this rename logic has ever worked...

I reworked/refactored the renaming of interfaces in `apply.py` to make use of iproute2 `ip link set eth_OLD set name eth_NEW` command instead and also added an integration test.

In order to rename interfaces via iproute2, they need to be DOWN, therefore the new code is bringing all to-be-renamed interfaces - which are not defined to be `critical` - down during `apply` and renames them accordingly.

Also, there was another issue with the detection of physical interfaces, where the system's current network interface names (from `netifaces.interfaces()`) have been compared to `config_manager.py`'s physical_interfaces list, which are not (always) actual interface names (but netplan IDs), especially for netplan definitions with a `match` condition. So I added another util to lookup/match the actual interface name for a given `match:` condition.

[0] https://www.freedesktop.org/software/systemd/man/systemd.net-naming-scheme.html
[1] https://github.com/systemd/systemd/blob/e62a7fea757f259eb330da5b6d3ab4ede46400a2/NEWS#L25

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: [LP: #1904662](https://pad.lv/1904662)

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -278,15 +278,22 @@ class NetplanApply(utils.NetplanCommand):
                 # may be the same for all interface members.
                 continue
             # Find current name of the interface, according to match conditions and globs (name, mac, driver)
-            current_iface_name = utils.find_matched_name(phy, match)
-            if current_iface_name and current_iface_name != newname:
-                critical = settings.get('critical', False)
-                if critical:
-                    # Skip interfaces defined as critical, as we should not take them down in order to rename
-                    logging.warning('Cannot rename {} ({} -> {}) at runtime (needs reboot), due to being critical'
-                                    .format(phy, current_iface_name, newname))
-                    continue
-                changes[current_iface_name] = {'name': newname}
+            current_iface_name = utils.find_matching_iface(interfaces, match)
+            if not current_iface_name:
+                logging.warning('Cannot find unique matching interface for {}: {}'.format(phy, match))
+                continue
+            if current_iface_name == newname:
+                # Skip interface if it already has the correct name
+                logging.debug('Skipping correctly named interface: {}'.format(newname))
+                continue
+            if settings.get('critical', False):
+                # Skip interfaces defined as critical, as we should not take them down in order to rename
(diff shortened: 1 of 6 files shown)


xnox:
https://github.com/systemd/systemd/blob/e62a7fea757f259eb330da5b6d3ab4ede46400a2/NEWS#L25

I think is the best source of explanation around this stuff. keep policy is the thing that causes this.

netplan/cli/commands/apply.py
```diff
@@ -245,68 +255,49 @@ def is_composite_member(composites, phy):  # pragma: nocover (covered in autopkg
     @staticmethod
     def process_link_changes(interfaces, config_manager):  # pragma: nocover (covered in autopkgtest)
```
sil2100, line 256: Some cleanup: after the recent changes, I think we can drop the `interfaces` parameter, as it doesn't seem to be used anymore.

sil2100 (approved):
Looks crisp, thanks! One inline clean-up, but not really required. It's generally good to go.

netplan/cli/commands/apply.py
slyon: Thanks for the review! The `interfaces` parameter is still used ([here](https://github.com/CanonicalLtd/netplan/pull/174/files#diff-40d0f1535a830c32c1e6884b3ce81254a5e9576c1443e059c74f1f4a89d55179R283)). It is passed to `find_matching_iface()` for matching the name glob onto the available network interfaces, so I'll keep that.

PR #178, opened 2020-12-15
Fix changing of macaddress with systemd v247

slyon:
## Description
Starting with systemd v247 the [.link](https://www.freedesktop.org/software/systemd/man/systemd.link.html#MACAddress=1) file section to change the MAC address does not seem to have an effect anymore:
```
[Link]
MACAddress=00:11:22:33:44:55
```
Changing/clearing `MACAddressPolicy=` did not help to make the .link file set the MAC address.

Whereas the same section applied to a [.network](https://www.freedesktop.org/software/systemd/man/systemd.network.html#MACAddress=1) file seems to fix the situation:
```
[Link]
MACAddress=00:11:22:33:44:55
```
Documentation indicates that those do the same thing, so I'm applying both now. It passes all integration tests this way, for old and new systemd.

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -572,6 +572,8 @@ write_network_file(const NetplanNetDefinition* def, const char* rootdir, const c
     if (def->mtubytes) {
         g_string_append_printf(link, "MTUBytes=%u\n", def->mtubytes);
     }
+    if (def->set_mac)
+        g_string_append_printf(link, "MACAddress=%s\n", def->set_mac);
 
     if (def->emit_lldp) {
         g_string_append(network, "EmitLLDP=true\n");
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -227,7 +227,7 @@ write_link_file(const NetplanNetDefinition* def, const char* rootdir, const char
         return;
 
     /* do we need to write a .link file? */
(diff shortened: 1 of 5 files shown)


sil2100 (approved):
Seems to look fine. I would normally worry that maybe having it in both places at once might cause some issues (it's confusing for sure), but you seem to have checked that this works fine for all systemd versions.
I'm wondering if maybe we should do `MACAddress=` only in the `.network` part, maybe that was the 'correct' way from the start? Like, also working properly with older networkd? But I guess it's safer this way, as we don't know the exact implications.

slyon:
Thank you for your feedback! You're right, it is much cleaner to have it only in one place.
I investigated a bit more about the two options: the .link file controls udev and sets the data/MAC only once after the device is created, while the .network file controls systemd-networkd and can also set the data/MAC during runtime (it seems like the device needs to be down for systemd <= 247, though).

So I think the usage of *only* the .network file is best suited for us, as it will work in more situations (i.e. "netplan generate && reboot" and "netplan apply"). I adopted the PR accordingly, to not write that information into the .link file, but only the .network file.

Running the full integration test-suite on Focal (old systemd) and on Hirsute (+ systemd 247) shows that this approach passes in both environments.

sil2100 (approved):
Awesome! After this investigation I feel much better and happier about the removed redundancy. Let's merge this.

PR #179, opened 2020-12-16
Fix 'networkmanager:' backend options for modem connections

slyon:
## Description
The `COMMON_BACKEND_HANDLERS` have been forgotten for `modem` connections apparently. Add them to allow the definition of the special `networkmanager:` mapping, used for NetworkManager integration. E.g. this `networkmanager.uuid` setting:

```
network:
  version: 2
  renderer: NetworkManager
  modems:
    mycon:
      auto-config: true
      networkmanager:
        uuid: b22d8f0f-3f34-46bd-ac28-801fa87f1eb6
```

We do not (yet) use that information (like uuid) in the current implementation. But reading YAML via NetworkManager will be broken if the `networkmanager:` mapping is not accepted.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -2221,6 +2221,7 @@ static const mapping_entry_handler vlan_def_handlers[] = {
 
 static const mapping_entry_handler modem_def_handlers[] = {
     COMMON_LINK_HANDLERS,
+    COMMON_BACKEND_HANDLERS,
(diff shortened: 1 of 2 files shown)


sil2100 (approved):
Ok, this looks good. I mean, we don't support `openvswitch:` for modems, but it doesn't really matter.

PR #181, opened 2021-01-11
Prepare for NetworkManager YAML backend functionality

slyon:
## Description
This is the first in a series of pull request to be prepared for setting up the functionality for a NetworkManager YAML/netplan backend, to be merged into [slyon/networkmanager-yaml-backend](https://github.com/CanonicalLtd/netplan/tree/slyon/networkmanager-yaml-backend) before it will collectively be merged into [master](https://github.com/CanonicalLtd/netplan/tree/master).

This PR refactors some of the YAML processing logic into *libnetplan* (from the *generate* binary), keeps track of the filename (absolute path) of the YAML file for each *NetplanNetDefinition* and implements a first additional library function `netplan_get_id_from_filename`, which will output the NetplanNetDefinition ID given an path to a netplan generated `.nmconnection` keyfile in `/run/NetworkManager/system-connections/`.

The netplan ID is not stored inside NM's data structures, therefore we need a way to gain netplan's identifier for any given connection profile.

## Checklist


src/parse.c
```diff
@@ -2385,6 +2389,8 @@ handle_network_type(yaml_document_t* doc, yaml_node_t* node, const void* data, G
         } else {
             netplan_netdef_new(scalar(key), GPOINTER_TO_UINT(data), backend_cur_type);
         }
+        g_assert(cur_filename);
+        cur_netdef->filename = g_strdup(cur_filename);
```
slyon, line 2393: This is actually only made use of in the follow-up PR #182

src/parse.c
```diff
@@ -2609,3 +2617,70 @@ netplan_clear_netdefs()
...
+
+    config_keys = g_list_sort(g_hash_table_get_keys(configs), (GCompareFunc) strcmp);
+
+    for (GList* i = config_keys; i != NULL; i = i->next)
+        process_input_file(g_hash_table_lookup(configs, i->data));
+    return TRUE;
+}
+
+gchar*
+netplan_get_id_from_filename(const char* filename, const char* ssid)
```
sil2100, line 2660: This is entirely fine, this function makes sense. One thing I'd like us to consider (now or at some later point, really) is to maybe change the name of this function. Since this is a NetworkManager-backend specific function, only useful whenever NM is in use. The `netplan_get_id_from_filename()` name seems to give off a vibe that it's for some general ID stored in the filename.
So maybe worth considering: changing the name to something like `netplan_get_nm_id_from_filename()` (or maybe some better placement of the `nm` part) and maybe moving this to `util.h` instead?

sil2100 (approved):
Approving with small comments. This is a +1 as none of the comments I left are blockers for this to land. But I'd really like to think about the function name and where it's defined.

src/parse.c
slyon: Yes... I was thinking that this function could be extended to also extract names from networkd filenames. But I agree that we don't have (nor need) this functionality currently, so we shouldn't name the function like this. We can still wrap around a more generic function(-name), once it is implemented in a more generic way.

I renamed it to `netplan_get_id_from_nm_filename` (as it only works for NetworkManager connection profile filenames, generated by netplan) and moved it into util.c/.h

Merging (and rebasing #182 on top of it).

PR #182, opened 2021-01-12
Implement netplan_generate and netplan_delete_connection APIs

slyon:
## Description
This is 2nd in a series of pull request implementing the functionality in `libnetplan` to provide a NetworkManager YAML backend. It builds upon #181 

* It improves the `netplan set` CLI a bit, to delete a file, if only `network: {version: 2}` is left.
* It implements a `netplan_generate` function, which will call `netplan generate` in the background by spawning another process.
  * At some point the `generate` binary should be refactored, to call functions of `libnetplan` instead, so we could use this same (refactored) functionality inside the library directly
  * But for now this is as good as it gets...
* It implements a `netplan_delete_connection` function, which will delete a connection from the YAML structure of a file (or potentially the whole file, if it is empty afterwards), by utilizing the `netplan set network.TYPE.IFNAME=NULL` functionality by spawning another process.
  * At some point we should get rid of this Python/C split of the YAML parsers, to have all the functionality available inside the C library.
  * But for now this is as good as it gets, without duplicating any logic...

## Checklist


slyon:
Rebased, after #181 was merged, to reduce the diff.

sil2100 (changes requested):
Ok, code-wise this looks good. Only potential change I'd like to look into is, as with the previous PR, thinking if `parse.h` is the right place for the two new functions. So far this header (and code file) have only defined things related to the actual netdef definitions and parsing - `netplan_delete_connection()` is for modifying the yaml state, so doesn't seem like the best fit (sure, we use some functions from parse, but that's it). Same for `netplan_generate()`. Those two are a bit problematic to find a good place for though. For now, I'd personally propose `util.h` *again*.

slyon:
Thank you very much for the review Łukasz!
You're right, `parse.[ch]` is supposed to contain parser stuff, which neiter `netplan_delete_connection` nor `netplan_generate` are. I've moved them into `util.c`, which should be okayish. Maybe we can find a better place (possibly a new module) in the future, as the implementation of this feature gets more complete and before we merge the big `slyon/networkmanager-yaml-backend` branch into `master`.

Feel free to merge this if PR if that is OK for you!

sil2100 (approved):
Yes, I think utils right now might be best. We can think of a new module later!

PR #183, opened 2021-01-13
Implement NM keyfile passthrough mode as a fallback mechanism

slyon:
## Description
This is 5th in a series of pull requests to prepare for the NetworkManager YAML backend. It builds upon #187 

It introduces a new (undocumented) top level type `others`, which needs to be used when a NM connection of unsupported type needs to be written. If used the `connection.type` needs to be specified by a `passthrough` setting, e.g.:
```
network:
  version: 2
  others:
    renderer: NetworkManager
    networkmanager:
      uuid: ...
      passthrough:
        connection.type: vxlan
        ipv4.method: auto
        ...
```
* `others` is undocumented on purpose, as this is not supposed to be used in regular netplan configs, but can be used as a fallback mechanism, if a given network connection type is not yet implemented in netplan. It can then still be used via `passthrough` mode.

Furthermore, this PR introduced a new module `nm-keyfile` to libnetplan, which contains some logic to help with integration of the NetworkManager YAML settings backend. Especially the `netplan_render_yaml_from_nm_keyfile` function, which transforms a given `GKeyFile*` structure into a valid `NetplanNetDefinition*` structure and uses the serializer to transform this into a valid netplan YAML and save it to disk. For now this makes heavy use the the `passthrough` fallback mechanism, but it will be extended to place keyfile settings into the correct netplan schema in the future, step by step.

The `nm.c` module was extended to make use of the `passthrough` fallback mechanism, to render a valid NetworkManager keyfile out of this data and write it to disk. It uses the internal netplan data structure to generate the keyfile as usual, but allows to extend (or override) specified settings via `passthrough` if they are not (yet/fully) supported.

As discussed in #181, the `netplan_get_id_from_nm_filename` function is moved into this new `nm-keyfile` module from `utils` as this might be a better place for keyfile specific functionality.

Currently this PR is based upon the [slyon/nm-4](https://github.com/CanonicalLtd/netplan/tree/slyon/nm-4) branch. It will be rebased once PR #187 is merged, but the underlying base shouldn't change too much, so it should be fine to review this already.

## Checklist


src/util.h
```diff
@@ -31,7 +31,6 @@ int wifi_get_freq24(int channel);
 int wifi_get_freq5(int channel);
 
 gchar* systemd_escape(char* string);
-gchar* netplan_get_id_from_nm_filename(const char* filename, const char* ssid);
```
slyon, line 34: This is intentionally moved to the new `nm-keyfile.c/h` module. As it derives the netplan-ID from the keyfile filename and is specific to NM keyfiles.

sil2100 (commented):
Ok, just writing down my thoughts here.

As mentioned, I suppose I had a slightly different idea for how this should be done. Yours might be hm, how to say it - more 'optimal' I guess, but feels slightly detached from the general netplan model we have right now. Thoughts below.

So as I see it done right now, it feels to me that the `nm-keyfile.*` module would be used directly by the thin-layer in NM to save the internal KeyFile into our yaml config file. As said, this is very optimal for the NM case, since it's basically just one call done on the NM side to get the resulting yaml config saved. It's not bad. Might be a straightforward way to go here.

The concept I had in my head was to just add a new `netplan` backend to the set of nm, ovs, sriov and networkd backends, so providing the same set of `write_netplan_conf()` functions for it as for any backend - that would be basically what `netplan_render_netdef()` does already. In this case, NM would basically call something like `netplan_render_yaml_from_nm_keyfile`, but instead called something like `netplan_parse_nm_keyfile()` - doing basically the same thing as now but without actually saving the yaml file, just generating the `NetplanNetDefinitions`. The next call would be to just do the `write_netplan_conf()` on the parsed `nd` we got from the previous step.
So as you can see, it's basically the same thing but less optimal as we're doing two steps instead of one - but in the end we still keep the same per-setting-backend model. This also extends the parser concept of netplan - previously we just had one parser, the yaml netplan one (in `parse.c`), but now that we learned to parse NM keyfile - we have a set of parsers and a set of generators. I really like the cleanness of this model, with the netplan core now basically being the `NetplanNetDefinition` definition, with different parsers generating it and different generators exporting it to other formats.

What do you think?

As for the code itself, still looking but so far it feels to be solid. We'd have to consult Steve regarding the `other` type. I'm not super happy about it, but then again I tried to figure out something that would be less troublesome but so far your `other` idea is the best.

src/nm-keyfile.c
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * NetworkManager writes the alias for '802-3-ethernet' (ethernet),
+ * '802-11-wireless' (wifi) and '802-11-wireless-security' (wifi-security)
+ * by default, so we only need to check for those. See:
+ * https://bugzilla.gnome.org/show_bug.cgi?id=696940
+ * https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/c36200a225aefb2a3919618e75682646899b82c0
+ */
+static const NetplanDefType
+type_from_str(const char* type_str)
```
sil2100, line 33: Maybe we should still support the default names just-in-case? Maybe one day we'd use this to parse some user-created NM configs, and maybe some users will use the canonical names instead of the aliases?

src/nm-keyfile.c
```diff
@@ -0,0 +1,251 @@
...
+    gchar *group_key = NULL;
+    gchar *value = NULL;
+    gsize klen = 0;
+    gsize glen = 0;
+
+    if (!*out_map)
+        *out_map = g_hash_table_new(g_str_hash, g_str_equal);
+    groups = g_key_file_get_groups(kf, &glen);
+    if (groups) {
+        for(unsigned i = 0; i < glen; ++i) {
```
sil2100, line 81:
```suggestion
        for (unsigned i = 0; i < glen; ++i) {
```

src/nm-keyfile.c
```diff
@@ -0,0 +1,251 @@
...
+            klen = 0;
+            keys = g_key_file_get_keys(kf, groups[i], &klen, NULL);
+            if (!keys) continue; // empty group
+            for(unsigned j = 0; j < klen; ++j) {
```
sil2100, line 85:
```suggestion
            for (unsigned j = 0; j < klen; ++j) {
```

tests/validate_docs.sh
```diff
@@ -41,6 +41,11 @@ for term in $(sed -n 's/[ ]\+{"\([a-z0-9-]\+\)", YAML_[A-Z]\+_NODE.*/\1/p' src/p
         continue
     fi
 
+    # 6. others is an undocumented fallback/passthrough feature
```
sil2100, line 44: Just an idea: maybe we should actually document this, but mention in the documentation that it's only for internal usage? Not sure what our approach to this is generally, but if it's undocumented then people might be all like 'WTF' when reading the NM-generated yaml neplan code.

sil2100 (changes requested):
Some minor stylistic code-style changes + ideas. Other then that it's all good.

src/nm-keyfile.c
slyon: Sure!

slyon:
> The concept I had in my head was to just add a new `netplan` backend to the set of nm, ovs, sriov and networkd backends, so providing the same set of `write_netplan_conf()` functions for it as for any backend - that would be basically what `netplan_render_netdef()` does already. In this case, NM would basically call something like `netplan_render_yaml_from_nm_keyfile`, but instead called something like `netplan_parse_nm_keyfile()` - doing basically the same thing as now but without actually saving the yaml file, just generating the `NetplanNetDefinitions`. The next call would be to just do the `write_netplan_conf()` on the parsed `nd` we got from the previous step.

Yes, I like that concept of keeping the same interfaces across the different backends! And just having `parsers` and `generators`.

> So as you can see, it's basically the same thing but less optimal as we're doing two steps instead of one - but in the end we still keep the same per-setting-backend model. This also extends the parser concept of netplan - previously we just had one parser, the yaml netplan one (in `parse.c`), but now that we learned to parse NM keyfile - we have a set of parsers and a set of generators. I really like the cleanness of this model, with the netplan core now basically being the `NetplanNetDefinition` definition, with different parsers generating it and different generators exporting it to other formats.
> 
> What do you think?

Yeah, using a more generic approach of parsers and generators will probably allow more flexibility in the future, I like that! The code above is mostly modularized in the required structure. I think I'd need to rename `serialize.c -> netplan.c` (being the new netplan/YAML backend – in line with nm.c, ovs.c, networkd.c, ...) and rename `nm-keyfile.c -> parse-nm.c` (being the new NM-Keyfile parser – in addition to the Netplan-YAML parser in parse.c). In addition, I need to juggle the function/API names around a bit, in order to match the interface names of the other parsers/generators.

> As for the code itself, still looking but so far it feels to be solid. We'd have to consult Steve regarding the `other` type. I'm not super happy about it, but then again I tried to figure out something that would be less troublesome but so far your `other` idea is the best.

Okay, let's check back with Steve about this. I don't like it too much either, but didn't find a better way to avoid failures on unsupported device types (like: vxlan, dummy, bluetooth, ...).

sil2100 (approved):
Ok, in that case let's remember about doing all the module renames and API changes at some point before getting it merged into main! Other than that, I think it's good to get it merged. We can iterate on the `other` bit with Steve on the nearest sync.

PR #184, opened 2021-01-19
NM: refactor keyfile generator, by using GLib's keyfile writer instead of custom writer

slyon:
## Description
This is the 3rd in a series of pull requests to enable the implementation of a YAML/netplan backend for NetworkManager.

In this PR netplan's current keyfile generation code (in `nm.c`) is refactored to make use of GLib's keyfile writer, instead of using a custom, home grown approach. This way the keyfile settings can easily be read, written and overwritten (e.g. by the keyfile passthrough/fallback settings in #183) and it should be more robust overall.

While on it: Clean up the usage of keyfile alias for `ethernet`, `wifi` and `wifi-security`, which NetworkManager writes by default since a long time, see: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/c36200a225aefb2a3919618e75682646899b82c0

## Checklist

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -179,19 +179,29 @@ addr_gen_mode_str(const NetplanAddrGenMode mode)
 }
 
 static void
-write_search_domains(const NetplanNetDefinition* def, GString *s)
+write_search_domains(const NetplanNetDefinition* def, const char* group, GKeyFile *kf)
 {
     if (def->search_domains) {
-        g_string_append(s, "dns-search=");
+        const gchar* list[def->search_domains->len];
         for (unsigned i = 0; i < def->search_domains->len; ++i)
-            g_string_append_printf(s, "%s;", g_array_index(def->search_domains, char*, i));
-        g_string_append(s, "\n");
+            list[i] = g_array_index(def->search_domains, char*, i);
+        g_key_file_set_string_list(kf, group, "dns-search", list, def->search_domains->len);
     }
 }
 
 static void
-write_routes(const NetplanNetDefinition* def, GString *s, int family)
+write_routes(const NetplanNetDefinition* def, GKeyFile *kf, int family)
 {
+    const gchar* group = NULL;
+    gchar* tmp_key = NULL;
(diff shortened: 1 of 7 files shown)


src/nm.c
```diff
@@ -515,297 +504,287 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
...
-
-        if (def->set_mac) {
-            g_string_append_printf(link_str, "cloned-mac-address=%s\n", def->set_mac);
-        }
-        if (def->mtubytes) {
-            g_string_append_printf(link_str, "mtu=%u\n", def->mtubytes);
+        const char* con_type = NULL;
+        switch (def->type) {
+            case NETPLAN_DEF_TYPE_WIFI:
+                con_type = "wifi";
```
sil2100, line 602: Just making sure, since I'm not that into NM right now: is this really `wifi` right now? Since I remember `802-11-wireless` and the docs also mention the same: https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless.html

Is `wifi` an alias or something?

sil2100 (approved):
Ok, I browsed through the changes and those seemed to make sense. I like the idea of using the Key-value parser for this, indeed makes it a bit cleaner.

But please see my inline question about one of the configuration settings that changed with this PR. If it's how it's indeed supposed to be (and the resulting configuration still works for wifi), please proceed with merging.

src/nm.c
slyon: Yes, I explicitly researched this as I wasn't sure and didn't change it in the first place. But I saw we had mixed definitions using 'wifi' and '802-11-wireless' (same for ethernet) and thought it makes sense to clean it up with this refactoring.

Both keywords are possible, but upstream NetworkManager switched over to using the short alias by default (wifi, wifi-security and ethernet) some 7 years ago. So I think we should do the same. See: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/c36200a225aefb2a3919618e75682646899b82c0

PR #185, opened 2021-01-24
fix wrong filter for restart wpa units

Epsilond1:
## Description
Step for reproduced:

```
sudo netplan try
Job for netplan-wpa-wlan0.service canceled.

An error occurred: Command '['systemctl', 'stop', 'systemd-networkd.service', 'netplan-wpa-*.service']' returned non-zero exit status 1.
```

In my shell:
```
sudo systemctl stop netplan-wpa-*.service
zsh: no matches found: netplan-wpa-*.service
``` 
-- returned non-zero exit code.

How does work?
```sudo systemctl stop netplan-wpa-wlan0.service``` -- exit code is 0.

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -152,7 +152,7 @@ class NetplanApply(utils.NetplanCommand):
                 utils.systemctl_daemon_reload()
             # Clean up any old netplan related OVS ports/bonds/bridges, if applicable
             NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
-            wpa_services = '"netplan-wpa-*.service"'
+            wpa_services = ['"netplan-wpa-*.service"']
             # Historically (up to v0.98) we had netplan-wpa@*.service files, in case of an
             # upgraded system, we need to make sure to stop those.
             if utils.systemctl_is_active('netplan-wpa@*.service'):
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -152,7 +152,7 @@ class NetplanApply(utils.NetplanCommand):
                 utils.systemctl_daemon_reload()
             # Clean up any old netplan related OVS ports/bonds/bridges, if applicable
             NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
-            wpa_services = ['netplan-wpa-{}.service'.format(iface) for iface in os.listdir('/sys/class/net/') if iface != 'lo']
+            wpa_services = '"netplan-wpa-*.service"'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: netplan/cli/commands/apply.py [hunk 1]
```diff
@@ -152,7 +152,7 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True):  # p
                 utils.systemctl_daemon_reload()
             # Clean up any old netplan related OVS ports/bonds/bridges, if applicable
             NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
-            wpa_services = ['netplan-wpa-*.service']
+            wpa_services = ['netplan-wpa-{}.service'.format(iface) for iface in os.listdir('/sys/class/net/') if iface != 'lo']
```
slyon, line 155:
```suggestion
            wpa_services = ['"netplan-wpa-*.service"']  # quote argument, so it won't be interpreted by the shell (zsh)
```

slyon (changes requested):
Hey @Epsilond1 thank you very much for contributing to Netplan and testing it in a ZSH environment.

The `zsh: no matches found` error seems to be a more generic problem, though, which we should not try to work around in netplan.

According to https://unix.stackexchange.com/questions/130958/scp-wildcard-not-working-in-zsh ZSH fails with this error if it cannot resolve the `netplan-wpa-*` glob, while bash passes it through as an argument to `systemctl`. Systemctl is correctly resolving the glob then.

So I think it would be better to make sure this parameter is passes through as an argument, instead of resolved by any shell. So I'd suggest to put the parameter in double-quotes instead of globbing via python, as there won't be an `netplan-wpa-IFACE.service` unit for each interface available in `/sys/class/net/`

Does this work for you?
```
systemctl stop "netplan-wpa-*.service"
```

If so, we should IMO add the double-quotes there, instead.

netplan/cli/commands/apply.py
Epsilond1: Yes, it is really working way.

slyon (approved):
Thank you very much! This looks good to me.

Let me run it through a round of integration tests, to make sure it does not interfere with any other part of the system before merging.

slyon (changes requested):
Unfortunately it doesn't seem to be that easy... Integration tests fail with:
```
Invalid unit name ""netplan-wpa-*.service"" escaped as "\x22netplan-wpa-*.service\x22" (maybe you should use systemd-escape?).
```

I think we need to find another way to pass through that parameter as a string, and avoid the shell/ZSH trying to expand the glob itself...

Cloud you have another look into that issue?

slyon:
Closing due to inactivity. Please re-open if somebody wants to pick up the work on this.

PR #186, opened 2021-01-26
Clarified `dhcp-identifier` option

lmlsna:
## Description
Added information about the other accepted scalar `duid` which can be specified or is assumed if the option is omitted entirely.

## Checklist


doc/netplan.md
```diff
@@ -285,9 +285,10 @@ Virtual devices
 
 ``dhcp-identifier`` (scalar)
 
-:   When set to 'mac'; pass that setting over to systemd-networkd to use the
-    device's MAC address as a unique identifier rather than a RFC4361-compliant
-    Client ID. This has no effect when NetworkManager is used as a renderer.
+:   (networkd backend only) Sets the source of DHCPv4 client identifier. If "mac"
+    is specified, the MAC address of the link is used. If this option is omitted,
+    or if "duid" is specified, networkd will generate an RFC4361-compliant client
```
slyon, line 290:
```suggestion
:   (networkd backend only) Sets the source of DHCPv4 client identifier. If ``mac``
    is specified, the MAC address of the link is used. If this option is omitted,
    or if ``duid`` is specified, networkd will generate an RFC4361-compliant client
```

slyon (approved):
Thank you very much for contributing to netplan and clarifying this documentation!

I'd only like to see the possible values emphasized as code-block, as we do at most of the other places in our reference (see inline suggestion). Once this is addressed, I'm +1 on merging!

lmlsna:
Should I squash these commits?

slyon:
Thanks a lot for the update @lmlsna ! No need to squash the commits, as I will squash-merge the full PR anyways.

LGTM. Merging.

PR #187, opened 2021-01-28
Prepare passthrough mode for NM backend and YAML serializer

slyon:
## Description
This is 4th in a series of pull requests. It enables parsing of the `passthrough` keyword inside the `networkmanager` backend settings, inside any netdef or wifi AP. Furthermore, it add a new `serialize.c/h` module, which can take a single netdef and render it into a YAML file. Testing is implemented by rendering a given YAML file into a new YAML file, and comparing the 1:1 match.

The YAML serializer does (by far) not yet support all supported netplan settings, but only the minimal set to enable the NetworkManager backend via passthrough mode.

The NM backend settings need to be available in every wifi AP definition (in addition to the netdef itself), as in the NetworkManager world, each wifi connection is a separate connection profile, while in the netplan world the wifi networks/AP are combined under a single device/netdef. But the settings passthrough (i.e. fallback mode) needs to be available for each individual NM connection profile, therefore we need them in the AP struct.

The `passthrough` setting is a mapping inside backend specific settings of key-value pairs, where the keys are separated by a dot in the format KEYFILE_GROUP.KEYFILE_KEY. Those values are passed through to the keyfile as is, and can be used as a fallback mechanism, where a specific feature is not yet implemented.
```
network:
  version: 2
  ethernets:
    renderer: NetworkManager
    networkmanager:
      uuid: ...
      passthrough:
        connection.type: vxlan
        ipv4.method: auto
        wireguard-peer.some-key.endpoint: 1.2.3.4
        ...
```

## Checklist


sil2100 (approved):
Ok, this feels good. I don't feel too strong about all the inline functions in serialize.h, but I suppose that's the best way to go with the YAML renderer. So code-wise this feels fine. I see it's incomplete as the renderer is missing a lot of code, but there's probably other PRs for the rest.

I'm still trying to get a good feel of the whole naming situation here. I think `serialize.*` seems good, but when thinking about this in the past I always thought we'd simply have a netplan settings backend. I don't feel strongly about that as well, but going this way I'd call `netplan_render_netdef` `write_netplan_conf()` to stay consistent with what we have with networkd, nm, ovs and sriov. Since I feel that's what basically this function is.

sil2100:
Ok, ignore my comment about the netplan backend for now - I see you went a bit of a different way here, looking at the part #5. Let's just continue this way and see how it ends up! Merging and slowly moving to #5.

PR #189, opened 2021-02-11
Improvements for the NetworkManager YAML backend integration

slyon:
## Description
This PR is 6th in a series of pull requests to prepare for the NetworkManager YAML backend. It builds upon #183 and contains a few improvements and fixes, enabling the NetworkManager test-suite (i.e. `make check`) to fully pass, especially the tests in the patched keyfile plugin (`src/settings/plugins/keyfile/tests/test-keyfile-settings`).

It implements the following functionality:
* Adding some basic support for the `modems` YAML schema, to allow matching of (physical) modem interfaces and enable the detection of GSM vs CDMA connections, which are distinct in NM while using the same definition in Netplan.
* Switching the `passthrough` key-value pairs to a DataList (instead of HashTable), to keep order of the keyfile elements, which is relevant in some cases (e.g. for `tc.qdiscs`)
* Avoid parsing and serialization of type `other` connections, as they might not contain the relevant handlers. Use full passthrough mode for those connections, to avoid parsing failures.
* Allow independent modification of Netdef ID & match.name, to enable changing the `connection.interface-name` from `nmcli`, while keeping the netdef ID equal to a previously existing ID. This enables overriding of existing netdef ID by 90-NM-... YAML file, keeping the same Netdef ID.
* Implement handling of keyfile UUIDs for each NM connection profile
* Implement handling of empty keyfile groups (i.e. relevant for `[bridge]`)

## Checklist

### DIFF
--- a/src/nm-keyfile.c
+++ b/src/nm-keyfile.c
@@ -177,14 +177,15 @@ netplan_render_yaml_from_nm_keyfile(GKeyFile* kf, const char* rootdir)
 
     /* Handle match: Netplan usually defines a connection per interface, while
      * NM connection profiles are usually applied to any interface of matching
-     * type (like wifi/ethernet/...). Therefore, we match the interface on '*'
-     * if not specified. */
-    nd->match.original_name = g_key_file_get_string(kf, "connection", "interface-name", NULL);
-    if (nd->match.original_name)
-        g_key_file_remove_key(kf, "connection", "interface-name", NULL);
-    else
-        nd->match.original_name = g_strdup("*");
-    nd->has_match = TRUE;
+     * type (like wifi/ethernet/...). */
+    if (nd->type < NETPLAN_DEF_TYPE_VIRTUAL) {
+        nd->match.original_name = g_key_file_get_string(kf, "connection", "interface-name", NULL);
+        if (nd->match.original_name)
+            _kf_clear_key(kf, "connection", "interface-name");
+        /* Set match, even if it is empty, so the NM renderer will not force
+         * the netdef ID as interface-name */
+        nd->has_match = TRUE;
+    }
 
     /* Modem parameters to detect GSM vs CDMA connections */
     nd->modem_params.auto_config = g_key_file_get_boolean(kf, "gsm", "auto-config", NULL);
--- a/src/nm-keyfile.c
+++ b/src/nm-keyfile.c
@@ -182,6 +182,16 @@ netplan_render_yaml_from_nm_keyfile(GKeyFile* kf, const char* rootdir)
         nd->match.original_name = g_strdup("*");
     nd->has_match = TRUE;
 
+    /* wake-on-lan, do not clear passthrough as we do not fully support this setting */
+    if (g_key_file_has_group(kf, "ethernet")) {
+        if (!g_key_file_has_key(kf, "ethernet", "wake-on-lan", NULL)) {
+            nd->wake_on_lan = TRUE; //NM's default is "1"
+        } else {
+            //XXX: fix delta between options in NM (0x1, 0x2, 0x4, ...) and netplan (bool)
+            nd->wake_on_lan = g_key_file_get_uint64(kf, "ethernet", "wake-on-lan", NULL) > 0;
+        }
+    }
+
     /* Special handling for WiFi "access-points:" mapping */
     if (nd->type == NETPLAN_DEF_TYPE_WIFI) {
         ap = g_new0(NetplanWifiAccessPoint, 1);
--- a/src/nm-keyfile.c
+++ b/src/nm-keyfile.c
@@ -81,7 +81,11 @@ read_passthrough(GKeyFile* kf, GHashTable** out_map)
         for (unsigned i = 0; i < glen; ++i) {
             klen = 0;
             keys = g_key_file_get_keys(kf, groups[i], &klen, NULL);
-            if (!keys) continue; // empty group
+            if (klen == 0) {
+                /* empty group */
+                g_datalist_set_data_full(list, g_strconcat(groups[i], ".", NETPLAN_NM_EMPTY_GROUP, NULL), g_strdup(""), g_free);
+                continue;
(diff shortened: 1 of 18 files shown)


src/nm-keyfile.c
```diff
@@ -138,32 +152,76 @@ netplan_render_yaml_from_nm_keyfile(GKeyFile* kf, const char* rootdir)
...
+    if (nd->type < NETPLAN_DEF_TYPE_VIRTUAL) {
+        nd->match.original_name = g_key_file_get_string(kf, "connection", "interface-name", NULL);
+        if (nd->match.original_name)
+            _kf_clear_key(kf, "connection", "interface-name");
+        /* Set match, even if it is empty, so the NM renderer will not force
+         * the netdef ID as interface-name */
+        nd->has_match = TRUE;
+    }
+
+    /* Modem parameters to detect GSM vs CDMA connections */
```
sil2100, line 194: hm, I don't think I understand this comment here. Below I only see handling of the GSM part - does this mean CDMA is to be handled via the passthrough? Is this what the comment implies?

src/parse.c
```diff
@@ -240,6 +240,8 @@ netplan_netdef_new(const char* id, NetplanDefType type, NetplanBackend backend)
     /* OpenVSwitch defaults */
     initialize_ovs_settings(&cur_netdef->ovs_settings);
 
+    if (!netdefs)
+        netdefs = g_hash_table_new(g_str_hash, g_str_equal);
```
sil2100, line 244: Wondering if since we now allocate the `netdefs` table in `netplan_netdef_new()`, maybe we could get rid of it from `netplan_parse_yaml()`? Since I don't think we put anything into `netdefs` before at least once calling `netplan_netdef_new()`.

sil2100 (commented):
Another chunk of solid code! Nothing per-se that directly needs fixing, but I had a few questions inline. After those are resolved, I guess we can merge.

sil2100:
Note to self: need to take a look at the new NM spec again to re-familiarize myself with the agreed approach, since in the meantime I feel as if I got confused a bit.

src/nm-keyfile.c
```diff
@@ -63,9 +63,22 @@ ap_type_from_str(const char* type_str)
     return NETPLAN_WIFI_MODE_OTHER;
 }
 
+static gboolean
+_kf_clear_key(GKeyFile* kf, const gchar* group, const gchar* key)
+{
+    gsize len = 1;
+    gboolean ret = FALSE;
+    ret = g_key_file_remove_key(kf, group, key, NULL);
+    g_strfreev(g_key_file_get_keys(kf, group, &len, NULL));
```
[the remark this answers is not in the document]
slyon: Yeah.. that's ugly. But unfortunately I did not find any other way how to know if this was the last key removed form a specific group. We could try to use non-public API (from https://github.com/GNOME/glib/blob/master/glib/gkeyfile.c#L499), but I don't think that would be any better...

src/nm-keyfile.c
slyon: I've update the comment to explain the situation a bit better:
```
    /* Modem parameters
     * NM differentiates between GSM and CDMA connections, while netplan
     * combines them as "modems". We need to parse a basic set of parameters
     * to enable the generator (in nm.c) to detect GSM vs CDMA connections,
     * using its modem_is_gsm() util. */
```

src/parse.c
slyon: Yes, I was wondering about this as well. Although, we do not put anything into `netdefs` before, there are a few places where we expect a pre-deployed hashmap for reading before. I've now figured out all of those spots and handled them accordingly.

src/nm-keyfile.c
```diff
@@ -138,32 +152,76 @@ netplan_render_yaml_from_nm_keyfile(GKeyFile* kf, const char* rootdir)
     }
     nd_type = type_from_str(type);
 
+    /* Use previously existing netdef IDs, if available, to override connections
+     * Else: generate a "NM-<UUID>" ID */
+    if (netdef_id)
+        nd_id = g_strdup(netdef_id);
+    else
+        nd_id = g_strconcat("NM-", uuid, NULL);
```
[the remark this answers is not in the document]
slyon: The problem with this is, that NM does not have an understanding of a "Netplan Netdef ID". In Netplan this can be an pretty arbitrary string/identifier, but usually it is the name of the interface. Furthermore, NM does not set a specific `interface-name` by default, but rather creates connections per connection type (ethernet, wifi, vxlan, ...), which it will then apply to any matching interface.

So when a new connection is created via NM (i.e. `nmcli con add type ethernet`) we're in a situation where we neither have a interface name, nor any pre-existing Netplan Netdef ID. In this case we fallback to a `NM-<UUID>` netdef id.

For consistency, the YAML filename will always be `90-NM-<UUID>.yaml`, as this way its origin can easily be detected by NM and/or netplan and it is always the same, as every NM connection profile will have a UUID.

The NM keyfile filename on the other hand, is generated by `nm.c` and follows the structure: `netplan-<NETDEF_ID>[-<SSID>].nmconnection`. So this can be something like `netplan-eth0.nmconnection` or `netplan-NM-SOME_UUID-SOME_SSID.nmconnection`. When a existing connection is modified via NM (i.e. `nmcli con mod netplan-eth0 connection.id myconnection`), it will extract the previous netdef id from the keyfile filename and pass it to this function, so that the newly created netdef can use the same ID, and overwrite the previous connection from the YAML, to avoid rendering the same connection twice under different names/IDs.

The naming of Netdef/YAML/Keyfile is quite complex but cannot be unified, due to missing information  (like the netdef ID in NM). But I hope this clarifies the situation a bit!

slyon:
@sil2100 Thank you very much for your review once again! I think I addressed your concerns and tried to explain the other situations inline. Let me know if there are any other questions or feel free to merge otherwise.

sil2100 (approved):
Looking good now!

PR #191, opened 2021-02-16
NM Integration: API update & Parser/Generator split

slyon:
## Description
This is 7th in a series of pull requests to provide the API needed for implementation of a NetworkManager YAML backend. It builds upon #189 

Changes contained in this PR:
* Split the code/modules into a YAML/netplan generator (`netplan.o` in addition to `nm.o`, `networkd.o`, `openvswitch.o`, ...)
* Split the code/modules into a keyfile/NM parser (`parse-nm.o` in addition to our YAML/netplan parser in `parse.o`)
* Rename `others` top-level device type/stanza to `nm-devices` to make it more explicit and print a warning when using it
* Move `netplan_get_id_from_nm_filename` back to `utils.o`, as it does not fit into the `parse-nm.o` module anymore (it is no parser specific functionality). Apply a small improvement to detect netplan_ids in testing scenario, where `rootdir` is provided.
* Unify the new netplan generator module API with the other generators: 
`void write_netplan_conf(const NetplanNetDefinition* def, const char* rootdir)`
* Unify the new keyfile parser module API with the other parser: 
`gboolean netplan_parse_keyfile(const char* filename, GError** error)`

## Checklist


sil2100 (approved):
Ok, I like this. I don't see anything that would need fixing, let's proceed with merging.

PR #194, opened 2021-02-24
Added TTL option for tunnels (LP: #1846783)

slyon:
## Description (Re-open of #139)
Some protocols set the TTL field of the packet to 1; when passing through the tunnel, the packet is discarded. To solve the problem, the tunnel has the TTL option, but it was not in netplan.

According to https://bugs.launchpad.net/netplan/+bug/1846783 this is required for IPIP/SIT/GRE tunnels.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/netplan/+bug/1846783

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -1087,6 +1087,10 @@ more general information about tunnels.
 
 :   Defines the address of the remote endpoint of the tunnel.
 
+``ttl`` (scalar)
+
+:   Defines the TTL of the tunnel.
+
 ``key``  (scalar or mapping)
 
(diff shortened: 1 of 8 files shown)


sil2100 (approved):
Ok, this one is pretty straightforward. And even though this changes the schema, it's straightforward enough that I don't think we need to wait for an ACK from Steve. Let's proceed.

QuAzI:
So how can I change TTL on WiFi connection with netplan now?

PR #199, opened 2021-02-28
Prevent duplicate ARPIPTargets in NetDev files (LP: #1915837)

kivancsunkar:
…(LP: #1915837)

https://bugs.launchpad.net/netplan/+bug/1915837

## Description

Using arp-ip-targets as below is generating 10-netplan-bond0.netdev file which contains ARPIPTargets property with duplicated values;

> arp-ip-targets:
>         - 10.10.10.10
>         - 20.20.20.20

> ARPIPTargets=10.10.10.10 20.20.20.20 10.10.10.10 20.20.20.20

Duplicate values in ARPIPTargets prevents systemd-networkd to process the netdev file due to a validation error in Ubuntu 18.04LTS.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://bugs.launchpad.net/netplan/+bug/1915837

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1624,6 +1624,18 @@ handle_ip_rules(yaml_document_t* doc, yaml_node_t* node, const void* _, GError**
 static gboolean
 handle_arp_ip_targets(yaml_document_t* doc, yaml_node_t* node, const void* _, GError** error)
 {
+    if (!cur_netdef->bond_params.arp_ip_targets) {
+        cur_netdef->bond_params.arp_ip_targets = g_array_new(FALSE, FALSE, sizeof(char *));
+    }
+
+    /* Avoid adding the same arp_ip_targets in a 2nd parsing pass by comparing
+     * the array size to the YAML sequence size. Skip if they are equal. */
+    guint item_count = node->data.sequence.items.top - node->data.sequence.items.start;
+    if (cur_netdef->bond_params.arp_ip_targets->len == item_count) {
+        g_debug("%s: all arp ip targets have already been added", cur_netdef->id);
+        return TRUE;
+    }
+
     for (yaml_node_item_t *i = node->data.sequence.items.start; i < node->data.sequence.items.top; i++) {
         g_autofree char* addr = NULL;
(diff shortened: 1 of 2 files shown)


slyon (changes requested):
Hello @kivancsunkar thank you very much for working on this and for your contribution to the Netplan.io project!

This PR looks good to me from a code point-of-view. But it introduces a reduction of test coverage (`make check-coverage`), as you can see here: https://app.codecov.io/gh/canonical/netplan/compare/199/diff (lines 1635-1635 are not covered by any test). Which means there is no test checking the case of multiple ARPIPTargets in NetDev files with more than one parsing pass.

Could you please add a new test case (e.g. in `tests/generator/test_common.py`) triggering this case, so we can have it covered and tested in our testsuite? This would most probably be a similar test case to `test_eth_address_option_multi_pass` from https://github.com/canonical/netplan/blob/master/tests/generator/test_common.py#L360

I'm looking forward to your update!
Thanks, Lukas

kivancsunkar:
Hi @slyon,

I tried my best to add a test case to the test suite for this case and then improved it to trigger a multiple pass on the YAML configuration. As far as I can tell from the local "make check-coverage" result (after I installed python3-nose package on my test environment), it was a successful test case as intended. However, the final commit haven't triggered a Codecov report yet.

Please let me know if you see room for improvement.

Thanks,
Kivanc

slyon:
Yes, this is looking good! I re-triggered the Github Action to run the coverage check again, something must have gone wrong there. It now shows a diff coverage of 100%, congrats!

Your test seems rather big, though. I think if we'd get rid of the `bond1`, `eno2` and `eno50` (and maybe even `eno49`?) definitions completely (but keeping `bonds` definition before `ethernets`, to trigger multi pass parsing), it should still run the same code path, but the test would be much smaller. Could you please improve this?

Afterwards I'm +1 on merging it!

kivancsunkar:
I updated the test case according to your feedback, thanks 👍

slyon (approved):
Thank you very much for your contribution :-)

LGTM!

PR #200, opened 2021-03-16
networkd: restart/reconfigure

slyon:
## Description
Replace the calls of `systemctl stop/start systemd-networkd.service` with `networkctl reload && networkctl reconfigure ...` during the execution of `netplan apply`. The devices to be reconfigured are currently all interfaces managed by systemd-networkd, as read from `networkctl`. This makes the whole restart/apply process quicker and avoids network interruptions for netplan's networkd backend.

The reload & reconfigure commands are available as of **systemd v244**.

In addition to the Apply-CLI changes, corresponding methods in `utils.py` are updated/refactored and the unit-test coverage is being improved.

## Checklist

### DIFF
--- a/netplan/cli/utils.py
+++ b/netplan/cli/utils.py
@@ -100,21 +100,12 @@ def nm_interfaces(paths, devices):
     return interfaces
 
 
-def systemctl_network_manager(action, sync=False):  # pragma: nocover (covered in autopkgtest)
-    service_name = NM_SERVICE_NAME
-
-    command = ['systemctl', action]
-    if not sync:
-        command.append('--no-block')
-
+def systemctl_network_manager(action, sync=False):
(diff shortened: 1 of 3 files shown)


netplan/cli/utils.py
```diff
@@ -100,50 +100,56 @@ def nm_interfaces(paths, devices):
...
+def systemctl(action, services, sync=False):
+    if len(services) >= 1:
+        command = ['systemctl', action]
 
-    subprocess.check_call(command)
+        if not sync:
+            command.append('--no-block')
 
+        for service in services:
+            command.append(service)
```
sil2100, line 119:
```suggestion
        command.extend(services)
```

sil2100 (approved):
Ok, this looks smooth. Thanks! (just one inline enhancement proposed)

PR #203, opened 2021-03-29
Fix parsing of OVS values with colons (LP: #1913906)

pdmack:
## Description
Bug fix for https://bugs.launchpad.net/netplan/+bug/1918197 and https://bugs.launchpad.net/netplan/+bug/1913906

Fixes the parse error for dict values that include ":"

`make check-coverage` errors out with less than 100% but so does the tip of master.

No new or changed keys.

## Checklist


slyon (approved):

This code is clean and sleek. It passes the 100% `make check-coverage` test for me locally (and also inside the Github Action CI), so that's looking good as well. I've added a small update to the integration test in `tests/integration/ovs.py` to check for this case, which also passes.

Thanks again. LGTM, ready for merging!

PR #204, opened 2021-04-07
Improve integration tests interface-ready waiting logic (LP: #1922126)

slyon:
## Description
This is a big refactoring of the integration test suite, which should make it much faster and more stable in different testing environments (e.g. armhf LXD containers).

The main change is in `tests/integration/base.py` and changes the waiting logic to decide when an given interface is readily configured. Previously, we were relying on `/lib/systemd/systemd-networkd-wait-online` to tell us magically when the networking is fully configured. This does not properly work for interfaces controlled by NetworkManager and is only watching the "general" networking state (i.e. whenever the first interface is up and running).

Waiting only on an arbitrary "first" network interface to be ready leads to races as we use asserts on many individual interfaces, which might not be ready at that time. So with this change the `generate_and_settle(self, wait_interfaces=None):` method got a new `wait_interfaces` argument, where the developer needs to specify each interface to be tested in a given test case. Furthermore, the developer can wait for a specified state (i.e. DHCP4/6 address assigned), to account for waiting on asynchronous background tasks, e.g.:

```
self.generate_and_settle([self.dev_e_client, 'eth43', self.state_dhcp4('mybond')])
```

All existing integration tests have been adopted to specify their interfaces-under-test in `generate_and_settle()`. Additionally, some logging was added to always print the name of the currently waited on interface and one dot per second (up to 60s), to see how long it takes to become ready (or times out), especially on slower architectures.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1922126

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -289,7 +289,7 @@ class IntegrationTestsBase(unittest.TestCase):
     def assert_iface_up(self, iface, expected_ip_a=None, unexpected_ip_a=None):
         '''Assert that client interface is up'''
 
-        out = self.assert_iface(iface, expected_ip_a=None, unexpected_ip_a=None)
+        out = self.assert_iface(iface, expected_ip_a, unexpected_ip_a)
         if 'bond' not in iface:
             self.assertIn('state UP', out)
 
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -293,7 +293,7 @@ class IntegrationTestsBase(unittest.TestCase):
         if 'bond' not in iface:
             self.assertIn('state UP', out)
 
-    def generate_and_settle(self):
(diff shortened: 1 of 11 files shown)


sil2100 (approved):
Ah, I actually forgot to approve this one earlier. So yeah, this looks good and I like this wait-logic much more than what we had previously.

sil2100:
Just curious: codecov seems to say there's less coverage this time, is that true or is it just getting confused?

slyon:
\o/ Thanks for the review and approval! The `make check-coverage` command shows a 100% coverage, so that must be a CodeCov hiccup. I'll be doing a (clean) rebase and merge.

PR #206, opened 2021-04-16
Fix ABI regression LP: #1922898 FR-1273

slyon:
## Description
In libnetplan v0.102 we introduced a regression, by breaking the ABI compatibility with netplan.io v0.101 (https://pad.lv/1922898)
This happened by introducing two new settings to the `NetplanNetDefinition` struct (`char* filename` and `guint tunnel.ttl`), modifying the memory layout, thus breaking applications linked against older versions of `libnetplan0`.

For `char* filename`, we can fix the problem by moving the new field to the end of the `NetplanNetDefinition` struct, thus not modifying the existing data layout, but only appending to it for the newer version.

For `guint netdef->tunnels.ttl` we cannot easily fix the problem easily, as this is a struct inside a struct and there seems to be only one byte left (at least on x86_64) before we hit the memory boundary, while this new `guint` field needs 4 bytes. So we rename that field as `tunnel_ttl` and append it to the end of the `net_definition` struct.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: https://pad.lv/1922898

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -1087,7 +1087,7 @@ more general information about tunnels.
 
 :   Defines the address of the remote endpoint of the tunnel.
 
-``ttl`` (scalar) – since **0.102**
+``ttl`` (scalar) – since **0.103**
 
 :   Defines the TTL of the tunnel.
 
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -1087,6 +1087,10 @@ more general information about tunnels.
 
 :   Defines the address of the remote endpoint of the tunnel.
 
+``ttl`` (scalar) – since **0.102**
+
+:   Defines the TTL of the tunnel.
+
 ``key``  (scalar or mapping)
 
 :   Define keys to use for the tunnel. The key can be a number or a dotted
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -1087,10 +1087,6 @@ more general information about tunnels.
 
 :   Defines the address of the remote endpoint of the tunnel.
 
-``ttl`` (scalar) – since **0.102**
-
-:   Defines the TTL of the tunnel.
-
 ``key``  (scalar or mapping)
 
 :   Define keys to use for the tunnel. The key can be a number or a dotted
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -142,6 +142,8 @@ write_tunnel_params(GString* s, const NetplanNetDefinition* def)
         g_string_append_printf(params, "Mode=%s\n", tunnel_mode_to_string(def->tunnel.mode));
     g_string_append_printf(params, "Local=%s\n", def->tunnel.local_ip);
     g_string_append_printf(params, "Remote=%s\n", def->tunnel.remote_ip);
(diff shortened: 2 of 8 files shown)


slyon:
Local autopkgtests PASS on Hirsute.
Local autopkgtests PASS on Focal, **using libnetplan 0.102 (+this patch) & netplan.io 0.101** (i.e. the regression case)

Bileto validation for Hirsute: https://bileto.ubuntu.com/#/ticket/4530
Bileto validation for Focal: https://bileto.ubuntu.com/#/ticket/4315

Four failures can be observed with the local "regression case" autopkgtest. This is expected as `advertised-receive-window` and `congestion-window` are new features of v0.102 which can be parsed by libnetplan 0.102 but the relevant NM/systemd configs are not written by the netplan.io 0.101 generator, as it does not know about those features. See the full logs below:
```
FAIL: test_per_route_advertised_receive_window (__main__.TestNetworkManager)
FAIL: test_per_route_congestion_window (__main__.TestNetworkManager)
FAIL: test_per_route_advertised_receive_window (__main__.TestNetworkd)
FAIL: test_per_route_congestion_window (__main__.TestNetworkd)

autopkgtest [18:01:11]: test cloud-init:  - - - - - - - - - - results - - - - - - - - - -
cloud-init           PASS
autopkgtest [18:01:12]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              FAIL non-zero exit status 1
vlans                PASS
wifi                 PASS
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
```
See the full logs here: https://paste.ubuntu.com/p/zhD8wnrGxw/

sil2100 (commented):
Ok, this seems to be properly reverting the ttl bits and I think this is adequate to what we did in hirsute, right? I suppose we still want to merge this?

That being said, I'm actually thinking we could potentially still get ttl working in a slightly 'dirty' way, but I guess we'll have to go this route anyway. What I'm thinking about is:
How about we just add do the main `net_definition` struct at the end a `guint tunnel_ttl` and do `nd->tunnel_ttl` instead everywhere in the code? This way we won't regress functionality re what we already had. Who knows, we might have to do it this way starting forward as it might not be possible to extend existing structures because of the same issue.

What do you think?

slyon:
> How about we just add do the main `net_definition` struct at the end a `guint tunnel_ttl` and do `nd->tunnel_ttl` instead everywhere in the code? This way we won't regress functionality re what we already had. Who knows, we might have to do it this way starting forward as it might not be possible to extend existing structures because of the same issue.
> 
> What do you think?

Yes, that should work. It's not pretty, but I guess we have to go that way (always append to the main `net_definition` struct) for now. That is until we have proper API isolation and _get() & _set() functions for each individual field. I can re-revert the TTL change and fix that up.

slyon:
The autopkgtests are looking good on Impish and Focal, also for the regression case where libnetplan is newer than the netplan generator. Merging.

PR #207, opened 2021-04-26
cli:set: Fix netplan-set on Core20

slyon:
## Description
Using os.replace() leads to "Invalid cross-device link" errors on Core20 setups, due to having different mounts.
Use shutil.copy() & os.remove() instead.

## Checklist

### DIFF
--- a/netplan/cli/commands/set.py
+++ b/netplan/cli/commands/set.py
@@ -160,7 +160,7 @@ class NetplanSet(utils.NetplanCommand):
             # Validate the newly created file, by parsing it via libnetplan
             utils.netplan_parse(tmpp)
             # Valid, move it to final destination
-            shutil.copy(tmpp, absp)
+            shutil.copy2(tmpp, absp)
             os.remove(tmpp)
         elif os.path.isfile(absp):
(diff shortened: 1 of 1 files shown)


netplan/cli/commands/set.py
```diff
@@ -159,7 +160,8 @@ def write_file(self, set_tree, name, rootdir='/'):
             # Validate the newly created file, by parsing it via libnetplan
             utils.netplan_parse(tmpp)
             # Valid, move it to final destination
-            os.replace(tmpp, absp)
+            shutil.copy(tmpp, absp)
```
xnox, line 163: wouldn't this loose permissions, ownership, extended attributes, etc? and is that intentional? or should an attempt be made to preserve acls? in which case `shutil.copy2()` should be used.

slyon: Good catch, thank you! Although, it shouldn't make a big difference, as that file (tmpp) is newly created by that method/process/user. The same method/process/user that would create/copy the file to the new location.
Still, preserving any metadata possible sounds like a good plan!

sil2100 (approved):
Good catch!

PR #208, opened 2021-05-03
dbus:cli: implement io.netplan.Netplan.Generate()

slyon:
## Description
Proxy the 'netplan generate' CLI via DBus, if called from within a snap,
to work around the strict AppArmor confinement.

see: https://github.com/snapcore/snapd/pull/10212

## Checklist

### DIFF
--- a/netplan/cli/commands/generate.py
+++ b/netplan/cli/commands/generate.py
@@ -18,8 +18,10 @@
 '''netplan generate command line'''
 
 import logging
+import os
 import sys
 import subprocess
+import shutil
 
 import netplan.cli.utils as utils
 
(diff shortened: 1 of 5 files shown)


anonymouse64 (approved):
lgtm, this matches the apply d-bus code we originally contributed

netplan/cli/commands/generate.py
```diff
@@ -44,6 +46,35 @@ def run(self):
...
+
+            if res != 0:
+                if res == 130:
+                    raise PermissionError(
+                        "failed to communicate with dbus service")
+                elif res == 1:
+                    raise RuntimeError(
+                        "failed to communicate with dbus service")
+            else:
+                return
```
sil2100, line 76: Since I don't see it straight away, but do we have test coverage for this conditional case as well (so res == 0)?

sil2100 (approved):
Looking good in overall. One question regarding coverage, but besides that this seems all correct.

netplan/cli/commands/generate.py
slyon: Yes. This is the default case, which is being checked in `test_netplan_generate_in_snap_calls_busctl` (https://github.com/canonical/netplan/pull/208/files#diff-2ef3425f71270b61c723558e26aceaa54c5eda80217aa9b216253b146b0754b4R146)

Also `make check-coverage` shows a 100% coverage for C and Python code (the CodeCov service is a bit off here, as it cannot execute all test cases correctly):
```
Overall coverage rate:
  lines......: 100.0% (3639 of 3639 lines)
  functions..: 100.0% (231 of 231 functions)
TOTAL                                     4899      0   100%
```

sil2100: Ah indeed! Was braindead for a moment there. hmm, so since I'm still a bit woozy, what about cases where the error code is != 0 and != 1 and != 130?

slyon: You're right, that case isn't handled... I was too quick with merging. You can find a followup PR here: https://github.com/canonical/netplan/pull/209

PR #210, opened 2021-05-06
netplan: support multi-line set

mvo5:
This adds support to write multi-line configurations into netplan.

## Description
The snapd netplan configuration api supports multi-line sets instead of the dotted key=value pairs.

This draft PR adds support to this to netplan via "netnplan set -" and then reads the multiline config
yaml "patch" via stdin.

This is a draft because I'm not super happy about the implementation, especially the error handling
is a bit flimsy and corner cases like huge stdout/stderr writes from netplan will make things hang.
OTOH it should be okay unless netplan really misbehaves.

I guess one question is if g_file_new_tmp/g_file_set_contents and a new "netplan set-file" or "set"
with a filename is not the easier interface to use instead of using stdin to pass the multi-line data
(with all the annoyances this brings to correctness when it comes to C). 

Anyway, hope this is helpful and I love hearing your thoughts. Test coverage is not 100% anymore
because I did not quite figure out how to test the "write()" failure in apply_multiline_netplan_set()

Closes: https://bugs.launchpad.net/netplan/+bug/1926486

## Checklist


slyon:
@mvo5 in the LP bug you suggested passing single line JSON would be a good approach.

Single line JSON is already supported (below the "network" key), like this:

`$ busctl call io.netplan.Netplan /io/netplan/Netplan/config/2HG220 io.netplan.Netplan.Config Set ss "network={ethernets:{eth42:{dhcp4: true, dhcp6: false}}}" ""`

Would that work for you?

mvo5:
Yeah, let's put this on hold for now, I will try the json route first

PR #211, opened 2021-05-20
Implement YAML generator (FR-1046)

slyon:
## Description
This is introducing a new YAML (netplan v2) rendere, enabling (lib-)netplan to write its own YAML config format from the currently parsed, internal data structures or a single, given `NetplanNetDefiniton` struct.

A big test coverage is achieved by extending the generator's `TestBase` class (from `generator/base.py`) to parse and re-generate all YAML passed to the `generate()` method and comparing the (normalized) original YAML input with the re-generated and normalized YAML output of this new renderer. That way the new YAML renderer is automatically tested for all existing and future test cases of the other renderers.

## Checklist


sil2100 (approved):
Looks crisp! Extensive testing, like the idea to validate existing generated yaml in tests. I just hope that non-standard yaml output is tested as well, like when users omit some quotes that the netplan generator otherwise always adds!
Anyway, this is good to go. Awesome work as always!

slyon:
> Looks crisp! Extensive testing, like the idea to validate existing generated yaml in tests. I just hope that non-standard yaml output is tested as well, like when users omit some quotes that the netplan generator otherwise always adds!
> Anyway, this is good to go. Awesome work as always!

Cool! Thank you very much for the (big) review!
I accepted your one-line cleanup/fix and rebased on top of `master`.
This should be good for merging now!

PR #212, opened 2021-05-26
Implement basic Keyfile parser (FR-1046)

slyon:
## Description
This PR builds upon #211 to implement a basic NetworkManager Keyfile parser, by providing the other part of the Keyfile -> YAML conversion.

I did some refactoring here, to move `tests/test_serialize.py` and `tests/test_nm_backend.py` into `tests/test_libnetplan.py` and `tests/parser/test_keyfile.py`, splitting it up to have all the library functions/utils tested in `test_libnetplan.py` and all the keyfile parsing logic tested in `test_keyfile.py`, using the new `tests/parser/base.py` class to provide Keyfile -> YAML conversion testing utils. In the new `parser/base.py` I'm also checking the reverse conversion (i.e. automatically doing a Keyfile -> YAML -> Keyfile conversion) to make sure we don't lose information on the way.

The primary keyfile parsing/conversion logic is implemented in `src/parse-nm.c` by reading different (and currently supported) settings form a NetworkManager keyfile, using GLib's `g_key_file_get_...` reader and setting up a `NetplanNetDefinition` struct in memory where the (converted and adapted) values can be stored for netplan to handle them.

Also, I added several test-cases to `parser/test_keyfile.py` to test the Keyfile->YAML conversion for supported settings. All unsupported settings will still be keep in the `passthrough` dict. That's each setting which doesn't get cleared using `_kf_clear_key(...)` during the parsing.

## Checklist


src/parse-nm.c
```diff
@@ -64,17 +65,306 @@ ap_type_from_str(const char* type_str)
...
+        if (!unhandled_data)
+            _kf_clear_key(kf, group, key);
+        g_free(key);
+        g_free(kf_value);
+    }
+}
+
+static void
+parse_dhcp_overrides(GKeyFile* kf, const gchar* group, NetplanDHCPOverrides* dataptr)
+{
```
sil2100, line 274:
```suggestion
{
    g_assert(dataptr);
```

For consistency.

src/parse-nm.c
```diff
@@ -64,17 +65,306 @@ ap_type_from_str(const char* type_str)
...
+static void
+parse_search_domains(GKeyFile* kf, const gchar* group, GArray** domains_arr)
+{
+    g_assert(domains_arr);
+    gsize len = 0;
+    gchar **split = g_key_file_get_string_list(kf, group, "dns-search", &len, NULL);
+    if (split) {
+        if (len == 0) {
+            _kf_clear_key(kf, group, "dns-search");
+            return;
```
sil2100, line 293: Ok, noob question: when uh, split is `!= NULL` but `len == 0`, does that mean something has been dynamically allocated? Do we still have to do `g_strfreev(split)` or there's nothing to be deallocated? This is a very unlikely case though I assume.

src/parse-nm.c
```diff
@@ -64,17 +65,306 @@ ap_type_from_str(const char* type_str)
...
+    }
+}
+
+static void
+parse_nameservers(GKeyFile* kf, const gchar* group, GArray** nameserver_arr)
+{
+    g_assert(nameserver_arr);
+    gchar **split = g_key_file_get_string_list(kf, group, "dns", NULL, NULL);
+    if (split) {
+        if (!*nameserver_arr)
```
sil2100, line 312: Related to the above: here we don't actually get or check the length of the resulting list - is this simply a different situation here and not being possible to be `len == 0`?

src/parse-nm.c
```diff
@@ -64,17 +65,306 @@ ap_type_from_str(const char* type_str)
...
+    handle_generic_str(kf, "802-1x", "ca-cert", &auth->ca_certificate);
+    handle_generic_str(kf, "802-1x", "client-cert", &auth->client_certificate);
+    handle_generic_str(kf, "802-1x", "private-key", &auth->client_key);
+    handle_generic_str(kf, "802-1x", "private-key-password", &auth->client_key_password);
+    handle_generic_str(kf, "802-1x", "phase2-auth", &auth->phase2_auth);
+}
+
+static void
+parse_bond_arp_ip_targets(GKeyFile* kf, GArray **targets_arr)
+{
```
sil2100, line 355:
```suggestion
{
    g_assert(targets_arr);
```

Also for consistency, I suppose.

src/parse-nm.c
```diff
@@ -197,42 +496,159 @@ netplan_parse_keyfile(const char* filename, GError** error)
         nd->has_match = TRUE;
     }
 
+    /* DHCPv4/v6 */
+    set_true_on_match(kf, "ipv4", "method", "auto", &nd->dhcp4);
+    set_true_on_match(kf, "ipv6", "method", "auto", &nd->dhcp6);
+    parse_dhcp_overrides(kf, "ipv4", &nd->dhcp4_overrides);
+    parse_dhcp_overrides(kf, "ipv6", &nd->dhcp6_overrides);
+
+    /* Manuall IPv4/6 addresses */
```
sil2100, line 505:
```suggestion
    /* Manual IPv4/6 addresses */
```

src/parse-nm.c
```diff
@@ -197,42 +496,159 @@ netplan_parse_keyfile(const char* filename, GError** error)
...
+            nd->wowlan = g_key_file_get_uint64(kf, "wifi", "wake-on-wlan", NULL);
+            _kf_clear_key(kf, "wifi", "wake-on-wlan");
+        } else {
+            nd->wowlan = NETPLAN_WIFI_WOWLAN_DEFAULT;
+        }
+
+        handle_common(kf, nd, "wifi");
     }
 
+    /* Cleanup some implicit keys */
```
sil2100, line 592: Good catch! I would miss this myself!

src/parse-nm.c
```diff
@@ -250,8 +666,52 @@ netplan_parse_keyfile(const char* filename, GError** error)
...
 
-        ap->hidden = g_key_file_get_boolean(kf, "wifi", "hidden", NULL);
-        _kf_clear_key(kf, "wifi", "hidden");
+        tmp_str = g_key_file_get_string(kf, "ipv4", "method", NULL);
+        if (tmp_str && g_strcmp0(tmp_str, "shared") == 0) {
+            ap->mode = NETPLAN_WIFI_MODE_AP;
+            _kf_clear_key(kf, "ipv4", "method");
+        }
+        g_free(tmp_str);
+
```
sil2100, line 675: Maybe a proposition regarding `netplan_parse_keyfile()` - but not for this PR, oh no, since this one is already quite big! But the proposition is to maybe divide this function into smaller ones. It's nothing bad as-is right now, just I guess most good-practice guidelines suggest for the functions to be relatively small. Guess somehow logically dividing the contents to 2-3 functions that would be called from inside could make it a bit better. Then again, this is only a suggestion for the future, since we need to think if this won't make the code even more complex than it is right now.

tests/parser/base.py
```diff
@@ -0,0 +1,169 @@
...
+            sys.stderr.flush()
+            os.dup2(tmp.fileno(), stderr_fd)
+            yield tmp
+        finally:
+            sys.stderr.flush()
+            os.dup2(stderr_copy, stderr_fd)
+            os.close(stderr_copy)
+
+
+class TestBase(unittest.TestCase):
```
sil2100, line 61: Maybe changing the name of this base class to be more representative of what the base is meant for? I mean, currently this is in `test/parser`, so this is more of like testing of various parser functionality. Which is fine. But the `base.py`'s `TestBase` class is typically for testing the keyfile parser. So possibilities: changing `TestBase` to something like `TestKeyfileBase` or changing the `generate()` method to `generate_from_keyfile()` or maybe renaming the `test/parser` directory? Not sure what is best!

sil2100 (approved):
+1. Left a few minor inline comments, suggestions and questions, but generally the whole concept and execution is as I imagined it.

I love how it all fits into the general parser + generator design we have in netplan, with libnetplan becoming a real two-way interpreter of multiple networking formats. I also really like how you kept similar naming between the existing `parser.c` and `parser-nm.c`, with all the static `handle_generic_*` helper functions. It all nicely clicks together! The testing story is also nicely handled.

So yeah, please take a look at the inline comments that I left (there's some minor suggestions for like typos I noticed), but then feel free to just merge it when you feel it's ready.

src/parse-nm.c
```diff
@@ -64,17 +65,306 @@ ap_type_from_str(const char* type_str)
...
+    g_assert(dataptr);
+    *dataptr = g_key_file_get_boolean(kf, group, key, NULL);
+    _kf_clear_key(kf, group, key);
+}
+
+static void
+handle_generic_str(GKeyFile* kf, const gchar* group, const gchar* key, char** dataptr)
+{
+    g_assert(dataptr);
+    g_assert(!*dataptr);
```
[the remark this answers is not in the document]
slyon: If this value is already set inside the NetplanNetDefinition, we do not want to override it, as this is usually a bug (and if it is only leaking memory, as the previous netdef value wouldn't be free'ed).

Overriding of the same config option in NetworkManager is still supported, as that is already handled by GLib's `g_key_file_get_string()` API, which will only always return the last (overwritten) config option. So we will only see a single value inside libnetplan anyways!

src/parse-nm.c
slyon: From my understanding if split `!= NULL` and `len == 0` it means we're getting a pointer, pointing to `NULL` (i.e. the last element of an – empty – NULL-terminated list). So all we have is a pointer that will go out of scope at the end of the function, no dynamically allocated data that we need to free.

Well... we could put a `g_strfreev(split)` in there, just in case, but IMO it's not needed.

Still we want to make sure to clear the "dns-search=" line (which NM generates by default) from the keyfile, so it doesn't end up in the passthrough.

src/parse-nm.c
slyon: Yeah it's a bit different, as NM generates a "dns-search=" line by default that we want to clear. It does not generate a "dns=" line by default. I restricted myself to only implicitly clearing the default values from the keyfile, but no (empty) data that was put in there manually (probably on purpose).

src/parse-nm.c
slyon: ACK! :+1:

slyon:
Sweet! Thank you very much for your great review.
I addressed most of the concerns and did a (clean) rebase on top of `origin/master`.
I will have a closer look again tomorrow, before merging it.

PR #213, opened 2021-06-08
Handle enum element override

kraj:
NETPLAN_DEF_TYPE_VIRTUAL and NETPLAN_DEF_TYPE_BRIDGE point
to same value in enum, however here they are assigned individually
which results in overriding the initialization of the objects

Fixes
src/netplan.h:85:33: error: initializer overrides prior initialization of this subobject [-Werror,-Winitializer-overrides]
    [NETPLAN_DEF_TYPE_BRIDGE] = "bridges",
                                ^~~~~~~~~

Signed-off-by: Khem Raj <raj.khem@gmail.com>

## Checklist


slyon (approved):
Good catch, this looks good to me! Thanks @kraj

PR #214, opened 2021-06-11
Implicitly consider devices with activation-mode set as optional

sil2100:
## Description

When the networkd backend is used, whenever an interface defines `activation-mode` it also needs to be implicitly marked as `optional`. Both of the available `activation-mode` states end up with the interface not being brought up by networkd on boot, and this can cause systemd to wait indefinitely for the interface to go online during wait-online. Currently systemd doesn't make this `NeededForOnline=no` assumption for us, so we need to do it ourselves.

I tried re-using the existing code for marking as optional not to duplicate too much code, so it's basically just moving stuff around.

## Checklist


slyon (approved):
This is a rather straight forward change and looks good to me. Thanks Łukasz!

PR #215, opened 2021-06-24
Add support for networkd ConfigureWithoutCarrier option on interfaces

n-cc:
## Description

The [ConfigureWithoutCarrier networkd option](https://www.freedesktop.org/software/systemd/man/systemd.network.html#ConfigureWithoutCarrier=) allows networkd to configure a specific link even if it has no carrier. This is useful for services such as isc-dhcp-server, which requires that all managed interfaces are up and assigned IP addresses regardless of their physical status.

I chose to put this key under `COMMON_LINK_HANDLERS` rather than `PHYSICAL_LINK_HANDLERS` as networkd does not differentiate between physical and virtual links WRT the `ConfigureWithoutCarrier` option, as far as I can tell. Better names are welcome.

## Checklist

- [x] Retains 100% code coverage (`make check-coverage`) *(`make check-coverage` reports the same coverage on master)


src/parse.h
```diff
@@ -295,6 +295,9 @@ struct net_definition {
     NetplanNetDefinition* vlan_link;
     gboolean has_vlans;
 
+    /* carrier */
+    gboolean configure_without_carrier;
```
slyon, line 299: This needs to move to the very end of the `net_definition` struct, to avoid breaking the ABI/memory layout on upgrades.

slyon:
Thank you very much for this PR! I didn't do a full review yet, but overall it is looking good from a code POV.

Though, as this contains a change of the YAML schema ("passthrough" of networkd's ConfigureWithoutCarrier= setting), I'd like to get an opinion from @vorlonofportland about if and how we want to support this use case before proceeding.

@n-cc Do you know if a similar feature is available and/or needed to support this with netplan's NetworkManager backend as well?

n-cc:
> @n-cc Do you know if a similar feature is available and/or needed to support this with netplan's NetworkManager backend as well?

I don't have much experience with NetworkManager, but `ignore-carrier` seems to be the equivalent option (manpage [here](https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html)). Adding the following section to `/etc/NetworkManager/NetworkManager.conf` allowed my physical interface to gain a static IP (configured separately under the `system-connections` directory) despite the physical link being down:

```
[device]
match-device=interface-name:enp0s31f6
managed=1
ignore-carrier=true
```

However, it doesn't seem that netplan's NetworkManager renderer currently templates the `[device]` section anywhere.

Looking at some of the common tests, it seems like there are situations where NetworkManager is used as the "main" renderer and networkd used for just specific ethernet interfaces; in these situations, my changes are sufficient to allow an interface to be assigned an IP without a physical link, so this feature can at least coexist with NetworkManager:

```
# cat /etc/netplan/01-network-manager-all.yaml 
# Let NetworkManager manage all devices on this system
network:
  version: 2
  renderer: NetworkManager
  ethernets:
    enp0s31f6:
      renderer: networkd
      addresses: [10.0.0.2/24]
      dhcp4: false
      dhcp-identifier: mac
      optional: false
      gateway4: 10.0.0.1
      configure-without-carrier: true
# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp0s31f6: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel state DOWN group default qlen 1000
    link/ether 30:5a:3a:83:8c:56 brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.2/24 brd 10.0.0.255 scope global enp0s31f6
       valid_lft forever preferred_lft forever
#
```

doc/netplan.md
```diff
@@ -277,6 +277,11 @@ Virtual devices
     Example to enable all link-local addresses: ``link-local: [ ipv4, ipv6 ]``
     Example to disable all link-local addresses: ``link-local: [ ]``
 
+``configure-without-carrier`` (bool)
```
vorlonofportland, line 280: my preference here is for the 'ignore-carrier' syntax, I think 'configure-without-carrier' is unnecessarily verbose.

vorlonofportland (changes requested):

doc/netplan.md
slyon, line 280:
```suggestion
``ignore-carrier`` (bool) – since **0.104*
```

src/parse.c
```diff
@@ -2235,6 +2235,7 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"activation-mode", YAML_SCALAR_NODE, handle_activation_mode, NULL, netdef_offset(activation_mode)}, \
     {"addresses", YAML_SEQUENCE_NODE, handle_addresses},                                      \
     {"critical", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(critical)},        \
+    {"configure-without-carrier", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(configure_without_carrier)}, \
```
slyon, line 2238:
```suggestion
    {"ignore-carrier", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(ignore_carrier)}, \
```

src/parse.h
```diff
@@ -401,6 +401,9 @@ struct net_definition {
   
     /* netplan-feature: activation-mode */
     char* activation_mode;  
+
+    /* carrier */
+    gboolean configure_without_carrier;
```
slyon, line 406:
```suggestion
    gboolean ignore_carrier;
```

slyon (approved):
Thank you Steve for your opinion on the `ignore-carrier` wording. Let's go with that.

So I think this PR is fine, as soon as the setting is renamed to "ignore-carrier".

Additionally, we should consider adopting it to the NetworkManager backend as well. Thanks for your investigation on this @n-cc ! We should be able to generated drop-in configs for NetworkManager in `/run/NetworkManager/conf.d/XX-netplan-NETDEF.conf` to contain the `[device]` section you mentioned. Would you be willing to look into this?

n-cc:
Thanks for the help getting this merged! This will hopefully benefit a lot of use-cases that require IP addresses set on inactive links.

> We should be able to generated drop-in configs for NetworkManager in /run/NetworkManager/conf.d/XX-netplan-NETDEF.conf to contain the [device] section you mentioned. Would you be willing to look into this (as part of an additional PR)?

I can give it a shot at some point, but if I recall correctly, the change to get it implemented was a bit more involved, as the `[device]` section is not currently templated by netplan in any NetworkManager config files, whereas networkd support was essentially a one-line change.

PR #216, opened 2021-07-07
Gateway fields deprecation and default routing support (FR-728) (LP: #1756590)

schopin-pro:
## Description

This patchset deprecates the `gateway4/6` fields, advising the user to use a proper route instead. The rationale is that it groups all routing matters in the same place, making things more consistent. Along with this deprecation comes the support for the "default" keyword in the route `to` field, and an expansion of the gateway validation step to encompass all default routes.

### Examples

```yaml
network:
  ethernets:
    en1:
      addresses:
      - 1.2.3.4/8
      - fc00:123:4567:89ab:cdef::1234/64
      routes:
      - to: default
        via: 1.1.1.1
      - to: default
        via: "fc00:123:4567:89cb:cdef::1"
```

Addresses LP: #1756590

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. https://pad.lv/1756590

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1119,7 +1119,7 @@ handle_gateway4(yaml_document_t* doc, yaml_node_t* node, const void* _, GError**
         return yaml_error(node, error, "invalid IPv4 address '%s'", scalar(node));
     set_str_if_null(cur_netdef->gateway4, scalar(node));
     g_warning("`gateway4` has been deprecated, use default routes instead.\n"
-            "See the 'Default routes' section of the documentation for more details.");
+              "See the 'Default routes' section of the documentation for more details.");
     return TRUE;
 }
 
@@ -1130,7 +1130,7 @@ handle_gateway6(yaml_document_t* doc, yaml_node_t* node, const void* _, GError**
         return yaml_error(node, error, "invalid IPv6 address '%s'", scalar(node));
     set_str_if_null(cur_netdef->gateway6, scalar(node));
     g_warning("`gateway6` has been deprecated, use default routes instead.\n"
-            "See the 'Default routes' section of the documentation for more details.");
+              "See the 'Default routes' section of the documentation for more details.");
     return TRUE;
 }
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -1655,6 +1655,7 @@ handle_routes(yaml_document_t* doc, yaml_node_t* node, const void* _, GError** e
         cur_route->scope = g_strdup("global");
         cur_route->family = G_MAXUINT; /* 0 is a valid family ID */
         cur_route->metric = NETPLAN_METRIC_UNSPEC; /* 0 is a valid metric */
+        cur_route->table = NETPLAN_ROUTE_TABLE_UNSPEC;
         g_debug("%s: adding new route", cur_netdef->id);
 
         if (!process_mapping(doc, entry, routes_handlers, NULL, error))
@@ -2701,10 +2702,14 @@ GHashTable *
 netplan_finish_parse(GError** error)
 {
     if (netdefs) {
+        GError *recoverable = NULL;
         g_debug("We have some netdefs, pass them through a final round of validation");
-        if (!validate_gateway_consistency(netdefs, error))
-            g_warning("More than one global gateway specified, the routing is ambiguous! "
-                      "Please set up multiple routing tables and use `routing-policy` instead.");
+        if (!validate_default_route_consistency(netdefs, &recoverable)) {
+            g_warning("Problem encountered while validating default route consistency."
+                      "Please set up multiple routing tables and use `routing-policy` instead.\n"
+                      "Error: %s", (recoverable) ? recoverable->message : "");
+            g_clear_error(&recoverable);
+        }
         g_hash_table_foreach(netdefs, finish_iterator, error);
     }
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -1118,6 +1118,8 @@ handle_gateway4(yaml_document_t* doc, yaml_node_t* node, const void* _, GError**
     if (!is_ip4_address(scalar(node)))
         return yaml_error(node, error, "invalid IPv4 address '%s'", scalar(node));
     set_str_if_null(cur_netdef->gateway4, scalar(node));
+    g_warning("`gateway4` has been deprecated, use default routes instead.\n"
+            "See the 'Default routes' section of the documentation for more details.");
     return TRUE;
 }
 
@@ -1127,6 +1129,8 @@ handle_gateway6(yaml_document_t* doc, yaml_node_t* node, const void* _, GError**
     if (!is_ip6_address(scalar(node)))
         return yaml_error(node, error, "invalid IPv6 address '%s'", scalar(node));
     set_str_if_null(cur_netdef->gateway6, scalar(node));
+    g_warning("`gateway6` has been deprecated, use default routes instead.\n"
+            "See the 'Default routes' section of the documentation for more details.");
     return TRUE;
 }
 
--- a/src/parse.c
+++ b/src/parse.c
@@ -1464,6 +1464,16 @@ handle_routes_ip(yaml_document_t* doc, yaml_node_t* node, const void* data, GErr
 }
 
 static gboolean
+handle_routes_destination(yaml_document_t *doc, yaml_node_t *node, const void *data, GError **error)
+{
+    const char *addr = scalar(node);
+    if (g_strcmp0(addr, "default") != 0)
+        return handle_routes_ip(doc, node, route_offset(to), error);
+    set_str_if_null(cur_route->to, addr);
+    return TRUE;
+}
+
+static gboolean
 handle_ip_rule_ip(yaml_document_t* doc, yaml_node_t* node, const void* data, GError** error)
 {
     guint offset = GPOINTER_TO_UINT(data);
@@ -1611,7 +1621,7 @@ static const mapping_entry_handler routes_handlers[] = {
     {"on-link", YAML_SCALAR_NODE, handle_routes_bool, NULL, route_offset(onlink)},
     {"scope", YAML_SCALAR_NODE, handle_routes_scope},
     {"table", YAML_SCALAR_NODE, handle_routes_guint, NULL, route_offset(table)},
-    {"to", YAML_SCALAR_NODE, handle_routes_ip, NULL, route_offset(to)},
+    {"to", YAML_SCALAR_NODE, handle_routes_destination},
     {"type", YAML_SCALAR_NODE, handle_routes_type},
     {"via", YAML_SCALAR_NODE, handle_routes_ip, NULL, route_offset(via)},
     {"metric", YAML_SCALAR_NODE, handle_routes_guint, NULL, route_offset(metric)},
--- a/tests/generator/test_common.py
+++ b/tests/generator/test_common.py
@@ -555,14 +555,46 @@ RouteMetric=100
 UseMTU=true
 '''})
 
-    def test_gateway(self):
+    def test_gateway4(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      addresses: ["192.168.14.2/24"]
+      gateway4: 192.168.14.1''')
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+Address=192.168.14.2/24
+Gateway=192.168.14.1
+'''})
+
+    def test_gateway6(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      addresses: ["2001:FFfe::1/64"]
+      gateway6: 2001:FFfe::2''')
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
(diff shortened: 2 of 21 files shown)


src/parse.c
```diff
@@ -1105,12 +1105,76 @@ handle_addresses(yaml_document_t* doc, yaml_node_t* node, const void* _, GError*
     return handle_generic_addresses(doc, node, TRUE, &(cur_netdef->ip4_addresses), &(cur_netdef->ip6_addresses), error);
 }
 
+const static char *GW_MULTIPLE_IP_ERROR = "gateway: there must be at most one of each IP type";
```
slyon, line 1108: I think we can move this into `handle_single_gateway()`, to keep it all together. It doesn't need to be a global variable, does it?

Also, please replace "IP type" => "IP address family" here as well.

src/parse.c
```diff
@@ -1105,12 +1105,76 @@ handle_addresses(yaml_document_t* doc, yaml_node_t* node, const void* _, GError*
...
+
+struct gw_parse_result {
```
slyon, line 1110: We should probably prefix this with "_" (`_gw_parse_result`) to mark it as: "for internal use only" and maybe move it into the `parse.h` header.

src/parse.c
```diff
@@ -1105,12 +1105,76 @@ handle_addresses(yaml_document_t* doc, yaml_node_t* node, const void* _, GError*
...
+    }
+
+    if (res.gw4) {
+        if (cur_netdef->gateway4)
+            g_assert_cmpstr(cur_netdef->gateway4, ==, res.gw4);
+        else
+            cur_netdef->gateway4 = g_strdup(res.gw4);
+    }
+
+    if (res.gw6) {
```
slyon, line 1161: I wonder if we need the `res.gw4/6` helper variable at all... I see how it nicely avoids some multi-pass parsing issues, as res is initialized as `{NULL, NULL}` for each new parsing pass. But then down here we still need to check if it was assigned already (to an equal value) in a previous parsing pass.

I think we should be able to write directly into `cur_netdef->gateway4/6` from inside `handle_single_gateway()`, comparing if the value already exists and is equal (=> skip). I guess this would not throw an error for cases like this: `gateway: [1.2.3.4, 1.2.3.4]` as this would look like being a 2nd parsing pass. But this is obviously wrong in the YAML and makes no difference, really (setting the same gateway twice). I wonder if we should make this trade-off in order to save passing around the extra `_gw_parse_result` struct. What do you think?

src/parse.c
```diff
@@ -1105,12 +1105,76 @@ handle_addresses(yaml_document_t* doc, yaml_node_t* node, const void* _, GError*
...
+        if (cur_netdef->gateway6)
+            g_assert_cmpstr(cur_netdef->gateway6, ==, res.gw6);
+        else
+            cur_netdef->gateway6 = g_strdup(res.gw6);
+    }
+
```
slyon, line 1167: Just for clarity:The `res.gw4/6` values are still the same from inside the YAML `node`, right? That's why we don't need to `g_free()` them down here?

slyon (commented):
Thank you very much for this great PR! I really like the overall design of your changes and how you integrated it nicely with the existing test-suite, especially the new YAML (re-)generator.

I've left some comments inline, mostly small wording or formatting things.
One bigger change I'd like to ask for is the handling of "multi-pass" parsing, that's a special case where an interface definition (netdef) might be parsed multiple times if it is defined as a slave interface (e.g. of a bridge) but the bridge itself is defined first in the YAML. I think we should setup a `gateway_multi_pass` test in `test_common.py`, similar to: https://github.com/canonical/netplan/blob/master/tests/generator/test_common.py#L408 instead of testing that inside the `test_tunnels.py` wireguard tests (which do multi-pass parsing as well).

Thank you and let me know what you think!

src/parse.c
schopin-pro: I used this approach initially, although I didn't think of strcmp'ing directly to the gateway4/6, which of course meant that the wireguard tests were failing.

However, in retrospect, I still prefer using the _gw_parse_result local structure to store the intermediate data while we're still validating the content of the field. This makes the failure case much easier to reason about. As a bonus, it catches an obvious configuration mistake which, while harmless in itself might prompt the user to look at their code a bit more carefully ;)

slyon:
Ah, please also mention the corresponding LP bug in your PR description and title, so we can find it, and put it in our release notes afterwards: [LP: #1756590](https://bugs.launchpad.net/netplan/+bug/1756590)

tests/generator/base.py
```diff
@@ -72,11 +72,11 @@
 UDEV_MAC_RULE = 'SUBSYSTEM=="net", ACTION=="add", DRIVERS=="%s", ATTR{address}=="%s", NAME="%s"\n'
 UDEV_NO_MAC_RULE = 'SUBSYSTEM=="net", ACTION=="add", DRIVERS=="%s", NAME="%s"\n'
 UDEV_SRIOV_RULE = 'ACTION=="add", SUBSYSTEM=="net", ATTRS{sriov_totalvfs}=="?*", RUN+="/usr/sbin/netplan apply --sriov-only"\n'
-ND_WITHIPGW = '[Match]\nName=%s\n\n[Network]\nLinkLocalAddressing=ipv6\nAddress=%s\nAddress=%s\nGateway=%s\n\
+ND_WITHIPGW = '[Match]\nName=%s\n\n[Network]\nLinkLocalAddressing=ipv6\nAddress=%s\nAddress=%s\nGateway=%s\nGateway=%s\n\
 ConfigureWithoutCarrier=yes\n'
 NM_WG = '[connection]\nid=netplan-wg0\ntype=wireguard\ninterface-name=wg0\n\n[wireguard]\nprivate-key=%s\nlisten-port=%s\n%s\
 \n\n[ipv4]\nmethod=manual\naddress1=15.15.15.15/24\ngateway=20.20.20.21\n\n[ipv6]\nmethod=manual\naddress1=\
-2001:de:ad:be:ef:ca:fe:1/128\n'
+2001:de:ad:be:ef:ca:fe:1/128\ngateway=2001:de:ad:be:ef:ca:fe:2\n'
```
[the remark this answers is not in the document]
schopin-pro: This was because of the mandatory coverage thing. If #218 is merged beforehand, I could simply convert the manual NULL-checking to use the new macro, which would have been battle-tested elsewhere anyway, rendering these changes unnecessary.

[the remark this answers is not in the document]
schopin-pro: In a future version (probably tomorrow) I'll put a proper unit test for the multipass situation in test_common.py as suggested in the review cover letter.

tests/generator/test_common.py
```diff
@@ -1153,35 +1221,6 @@ def test_eth_ipv6_privacy(self):
 [ipv6]
 method=auto
 ip6-privacy=2
-'''})
-
-    def test_gateway(self):
```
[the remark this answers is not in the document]
schopin-pro: Good catch, this was an oversight! Fixed.

tests/parser/test_keyfile.py
```diff
@@ -286,8 +286,9 @@ def test_keyfile_method_manual(self):
         - foo.local
         - bar.remote
         - bar.local
-      gateway4: 6.6.6.6
-      gateway6: 6:6::6
+      gateway:
+      - "6.6.6.6"
+      - "6:6::6"
```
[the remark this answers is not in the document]
schopin-pro: It turns out this is compared to the text output of the generator, which uses the long form.

tests/generator/base.py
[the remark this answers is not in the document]
schopin-pro: Addressed. Thanks for the heads-up, I had actually misunderstood the reasons behind the multi-pass: I thought it was needed for the definition of a Wireguard interface, which retrospectively was a bit weird :grin:.

schopin-pro:
Thank you for the thorough review. There is only one point remaining regarding the implementation. I'd rather keep the current code, but I can change it if you feel strongly about it :)

src/parse.c
slyon: ACK. Properly handling the (obvious) error case is a valid reason to keep that struct around, IMO.

slyon (approved):
Thank you for addressing all my comments so quickly.
This LGTM now :eyes: 

Ready for merging!

src/parse.c
```diff
@@ -1112,12 +1112,66 @@ handle_addresses(yaml_document_t* doc, yaml_node_t* node, const void* _, GError*
...
+handle_single_gateway(yaml_node_t* node, GError** error, struct _gw_parse_result *out)
+{
+    const char *addr;
+
+    assert_type(node, YAML_SCALAR_NODE);
+
+    addr = scalar(node);
+    if (is_ip4_address(addr)) {
+        if (out->gw4)
+            return yaml_error(node, error, "gateway: there must be at most one of each IP family");
```
sil2100, line 1130: This is good, but I wonder if it's maybe even possible to be more clear about the error here. Since I think at first we might be getting a lot of those!

sil2100 (approved):
I think this is good and I think we can merge it as-is.

src/parse.c
schopin-pro: Thanks for the feedback! I tried to improve the message, let me know if you have something more in mind.

slyon (commented):
Thank you Simon for your comments on the specification and the re-work of this PR!

This is looking very good! I especially like that you've integrated checking (and testing!) the legacy `gateway4/6` fields into the new `validate_default_route_consistency()` function.

Apart from some tiny inline comments, I have two remarks, which I'd like you to adopt before we merge this PR:
* Please update the PR description (that will become the commit message), to reflect the latest changes
* As this is a bigger change, I would like to see an integration test, validating that it is working as expected on a running system (autopkgtest)
   * This should be placed in the `_CommonTests()` class in `tests/integration/routing.py` and could look similar to the `test_routes_v4()` test (but checking default routes for IP4 & IP6).

slyon:
Thanks for addressing my comments! I've pushed a small fixup to remove some DHCP references (in comments) where they don't need to be in `test_routes_default()` and can confirm that the routing integration tests are working as expected:

```
$ autopkgtest -U ../netplan.io_0.102-99wip1_amd64.changes --test-name=routing -- lxd autopkgtest/ubuntu/impish/amd64
autopkgtest [15:40:05]: starting date: 2021-07-26
autopkgtest [15:40:05]: version 5.16ubuntu0.1
[...]
autopkgtest [15:41:17]: test routing: python3 tests/integration/run.py --test=routing
autopkgtest [15:41:17]: test routing: [-----------------------
test_per_route_advertised_receive_window (__main__.TestNetworkManager) ... eth42 .ok
test_per_route_congestion_window (__main__.TestNetworkManager) ... eth42 .ok
test_per_route_mtu (__main__.TestNetworkManager) ... eth42 .ok
test_route_from (__main__.TestNetworkManager) ... eth42 ok
test_route_on_link (__main__.TestNetworkManager) ... eth42 ..ok
test_route_table (__main__.TestNetworkManager) ... eth42 .ok
test_routes_default (__main__.TestNetworkManager) ... eth42 .ok
test_routes_v4 (__main__.TestNetworkManager) ... eth42 .ok
test_routes_v4_with_dhcp (__main__.TestNetworkManager) ... skipped 'fails due to networkd bug setting routes with dhcp'
test_routes_v6 (__main__.TestNetworkManager) ... eth42 ..ok
test_link_route_v4 (__main__.TestNetworkd) ... eth42 .............ok
test_per_route_advertised_receive_window (__main__.TestNetworkd) ... eth42 ..............ok
test_per_route_congestion_window (__main__.TestNetworkd) ... eth42 .............ok
test_per_route_mtu (__main__.TestNetworkd) ... eth42 ..............ok
test_route_from (__main__.TestNetworkd) ... eth42 .............ok
test_route_on_link (__main__.TestNetworkd) ... eth42 ..............ok
test_route_table (__main__.TestNetworkd) ... eth42 ..............ok
test_route_type_blackhole (__main__.TestNetworkd) ... skipped 'networkd does not handle non-unicast routes correctly yet (Invalid argument)'
test_route_with_policy (__main__.TestNetworkd) ... eth42 ..............ok
test_routes_default (__main__.TestNetworkd) ... eth42 .............ok
test_routes_v4 (__main__.TestNetworkd) ... eth42 ..............ok
test_routes_v4_with_dhcp (__main__.TestNetworkd) ... skipped 'fails due to networkd bug setting routes with dhcp'
test_routes_v6 (__main__.TestNetworkd) ... eth42 ..............ok

Ran 23 tests in 192.097s

OK (skipped=3)
autopkgtest [15:44:30]: test routing: -----------------------]
autopkgtest [15:44:31]: test routing:  - - - - - - - - - - results - - - - - - - - - -
routing              PASS
autopkgtest [15:44:31]: @@@@@@@@@@@@@@@@@@@@ summary
routing              PASS
```

raharper:
Is there a feature flag or some other way other than version for netplan generation programs to know if the netplan in use will complain about gateway4/6 ?   cloud-init project generates netplan and currently looks at netplan info for features to detect how it should construct a netplan config.  For example, we check for ipv6-mtu in features to handle the previous use of 'mtu' vs 'ipv6-mtu'

slyon:
Good point @raharper. We do not have a feature flag for this in v0.103, but I've just added the `default-routes` feature flag for the future. In the meantime `gateway4`/`gateway6` can still be used as-is.

PR #217, opened 2021-07-08
parser: warn user on multiple global gateway (LP: #1901836) (FR-728)

schopin-pro:
## Description

It doesn't make sense for the user to have multiple global gateways. This patch introduces a consistency check at the end of the parsing stage, and only emits a warning in order not to break current configurations.

Addresses LP #1901836

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -2684,8 +2684,8 @@ netplan_finish_parse(GError** error)
     if (netdefs) {
         g_debug("We have some netdefs, pass them through a final round of validation");
         if (!validate_gateway_consistency(netdefs, error))
-            g_warning("More than one global gateway specified, the routing is ambiguous!"
-                    "Please set up multiple routing tables and use `routing-policy` instead.");
+            g_warning("More than one global gateway specified, the routing is ambiguous! "
+                      "Please set up multiple routing tables and use `routing-policy` instead.");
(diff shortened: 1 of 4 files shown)


schopin-pro:
Note that #216 and this PR are in slight conflict, as merging one of them will lead to the tests failing on the other. To solve this is only a matter of changing the keywords from gateway4/6 to gateway in the tests introduced in this PR.

slyon:
Please mention [LP: #1901836](https://bugs.launchpad.net/netplan/+bug/1901836) in the PR description and title.

slyon (commented):
Thank you for another great PR! Contrary to the PR description, we do actually support multiple routing tables (`table:` setting) – maybe you could update that accordingly.

> This patch introduces a consistency check at the end of the parsing stage, and only emits a
warning in order not to break current configurations.

Yes that's good. We do not want to break existing configurations in the field. Although, arguably configuration defining multiple gateways are already broken (in a different way).

So I like your changes: Clean functions and straight forward testing.
I just want to ask for two small changes see inline comments):
* Logical order of execution of the `validate_gateway_consistency()` function
* Verbosity of the error message, please feel free to extend/modify the example I gave below

slyon (approved):
Thank you Simon, LGTM! :eyes:

I've cleaned up some whitespace in the warning message log and I think it's ready for merging, independently of #216

PR #218, opened 2021-07-08
parse: new macro set_str_if_null to avoid duplicating existing fields

schopin-pro:
## Description

When we do multi-pass parsing, we sometimes duplicate string without
freeing the previous one. This macro allows us to avoid said
duplication.

The commit uses the macro in a couple of places, but there are a lot of
other sites where it might be useful.

## Checklist


slyon (approved):
Thanks, that's a nice helper util!
LGTM :+1:

PR #219, opened 2021-07-20
parse: handle_routes: clean up the route on error

schopin-pro:
This patch reworks the single route processing to only append the parsed
route to the interface when it has been fully validated, and use a
single error path otherwise. This allows us to locally control the
cleanup and freeing of the allocated route.

The previous version could indeed leak if process_mapping returned an
error (and that's assuming that the interface cleanup process would free
up the routes already assigned, which I haven't checked).

## Checklist


slyon (approved):
Thanks! That looks good to me and passes all the tests.

PR #220, opened 2021-07-20
parse: routes: avoid creating dangling pointers on error

schopin-pro:
## Description

Parsing a yaml route looking like this:

```yaml
routes:
- to: 162.168.0.0/16
- to: invalid value
```

would result in having a dangling pointer in cur_route->to, as its
content would have been freed when starting to parse the invalid value,
without the pointer itself having been cleared. By deferring the free() to
right before the pointer update, we ensure that (barring interrupts or
multithreading, neither of which are supported AFAIK) the pointer remain
valid.

Note that to be thread-safe, we'd need to do an atomic swap and only
then free the variable, but that's out of scope ATM.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1446,7 +1446,6 @@ handle_routes_ip(yaml_document_t* doc, yaml_node_t* node, const void* data, GErr
     guint offset = GPOINTER_TO_UINT(data);
     int family = get_ip_family(scalar(node));
     char** dest = (char**) ((void*) cur_route + offset);
-    g_free(*dest);
 
     if (family < 0)
         return yaml_error(node, error, "invalid IP family '%d'", family);
@@ -1454,6 +1453,7 @@ handle_routes_ip(yaml_document_t* doc, yaml_node_t* node, const void* data, GErr
     if (!check_and_set_family(family, &cur_route->family))
         return yaml_error(node, error, "IP family mismatch in route to %s", scalar(node));
 
+    g_free(*dest);
     *dest = g_strdup(scalar(node));
 
     return TRUE;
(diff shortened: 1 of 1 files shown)


slyon (approved):
Nice find!
I'm generally +1 on this, but I wonder if we could do even better by using something like `g_clear_pointer(dest, g_free);` instead of the original `g_free(*dest);` (at the top of the functions), i.e. free'ing the value and resetting the pointer to `NULL` for the time being, before it is set to the new value.

That way we could avoid the dangling pointers and at the same time avoid keeping a pointer to an outdated/overwritten value (i.e. the first `to: 162.168.0.0/16`). That's because when `handle_routes_ip` is called for the 2nd time, the first value is basically outdated/overwritten and should not be kept around, as that could lead to confusion during debugging.

What do you think?

schopin-pro:
The (slight) problem I see with your proposal is that it relies heavily on the assumption that the previous value is in any case obsolete. This property is not local to the function but comes rather from knowing all its call sites, which all bubble up errors instead of treating them locally.

However, we could easily imagine a caller that would do something like this

```c
GError tmperr;
if (!handle_routes_ip(doc, node, dest, &tmperr)) {
    g_warn("Invalid IP, dismissing entry");
    log_error_as_warning(&tmperr);
}
```

This is a contrived example, and I'm fairly confident that in this particular instance both solutions would be equally future-proof, but in general I much prefer the pattern of delaying side-effects as later as possible, as it tends to make the code more robust and easier to reason about.

schopin-pro:
However, thank you for the review and the mention of `g_clear_pointer`, I didn't know of this function :)

slyon:
Yeah, I see that if the function is being used for checking the pending value it could lead to problems if we free to early...
The function is not currently used in this way (and not really intended for this purpose) so it is mostly a theoretical problem. As I was just trying to improve another theoretical problem with my `g_clear_pointer` suggestion as well, let's just go on with this changeset as is!

PR #221, opened 2021-07-23
cli: Fix failure with "set-name" and "bridges"

mkj:
## Description
The code path when a config has `set-name` as well as `bridges` doesn't seem to work at present, failure below. This change seems to fix it. 

## Checklist

Config:
```yaml
network:
  version: 2
#  renderer: networkd
#  networkd doesn't handle ethernet well...
  renderer: NetworkManager
  ethernets:
    devnet:
      renderer: networkd
      match:
        name: enx00aa11223344
    enx00bb11223344:
      addresses: [ 10.20.0.1/24 ]
      renderer: networkd
    wlan:
      match:
        macaddress: 00:99:11:22:33:44
      set-name: "wlan0"

  bridges:
    # ... and NetworkManager doesn't handle bridges well
    renderer: networkd
    br0:
      addresses: [ "10.9.0.1/16" ]
      interfaces: [ "devnet" ]
```

```
netplan try
  File "/usr/share/netplan/netplan/cli/commands/try_command.py", line 84, in command_try
    NetplanApply().command_apply(run_generate=True, sync=True, exit_on_error=False)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 191, in command_apply
    changes = NetplanApply.process_link_changes(devices, config_manager)
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 281, in process_link_changes
    if NetplanApply.is_composite_member(composite_interfaces, phy):
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 252, in is_composite_member
    members = settings.get('interfaces', [])
AttributeError: 'str' object has no attribute 'get'
```

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -258,6 +258,8 @@ class NetplanApply(utils.NetplanCommand):
         """
         for composite in composites:
             for _, settings in composite.items():
+                if not type(settings) is dict:
+                    continue
                 members = settings.get('interfaces', [])
                 for iface in members:
                     if iface == phy:
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -251,7 +251,7 @@ class NetplanApply(utils.NetplanCommand):
             utils.systemctl_network_manager('start', sync=sync)
 
     @staticmethod
-    def is_composite_member(composites, phy):  # pragma: nocover (covered in autopkgtest)
+    def is_composite_member(composites, phy):
         """
         Is this physical interface a member of a 'composite' virtual
         interface? (bond, bridge)
--- a/tests/test_cli_units.py
+++ b/tests/test_cli_units.py
@@ -1,0 +1,41 @@
+#!/usr/bin/python3
+# Blackbox tests of netplan CLI. These are run during "make check" and don't
+# touch the system configuration at all.
+#
+# Copyright (C) 2021 Canonical, Ltd.
+# Author: Lukas Märdian <slyon@ubuntu.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


netplan/cli/commands/apply.py
```diff
@@ -251,17 +251,16 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True):  # p
...
     @staticmethod
-    def is_composite_member(composites, phy):  # pragma: nocover (covered in autopkgtest)
+    def is_composite_member(composites, phy):
         """
         Is this physical interface a member of a 'composite' virtual
         interface? (bond, bridge)
         """
         for composite in composites:
-            for _, settings in composite.items():
-                members = settings.get('interfaces', [])
```
slyon, line 261: I think we should rather check something like this before `members = settings.get('interfaces', [])`:
```
if not type(settings) is dict:
    continue
```

slyon (changes requested):
Thank you very much @mkj for your contribution!
This is/was an untested code-path, which is bad (from Netplan's side!). I've added a new testing class and corresponding unit tests for this method, to make sure it does not fail in the future. And pushed this unittest to your branch!

The unit test also shows that your implementation does not seem to work correctly, it always returns `False`, I've left an inline comment of what should be an easy fix for this issue.

What do you think?

netplan/cli/commands/apply.py
mkj: Yes that is a much more sensible fix, I had mistaken what the arguments were. Thanks

slyon: Could you update that accordingly? As I do not want to do a self-review & merge of this change.

mkj: I've pushed that change now.

slyon (approved):
Thank you for fixing up the comments I had.
This is ready for merging!

PR #223, opened 2021-08-20
tests:base:ethernets: Improve stability of autopkgtests

slyon:
Make sure NetworkManager's `unmanaged-devices` are kept in check so that NM does not take control over some network interfaces that are controlled by the test system, such as the routers veth42, veth42, fake wifi AP, ...

Replaces the "autopkgtest-fixes.patch" Distro patch.

Changing the location of the denylist from /run to /etc because the changes in /run would be overwritten by our test system's /etc/NetworkManager/conf.d/90-test-ignore.conf, because /etc/ drop-in configs are read after /run/ drop-ins and the /etc/ drop-in uses a hard overwrite of `unmanaged-devices=...` in order to reset the system's default in /usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf (guess we could also downgrade the test systems static /etc/ drop-in to /run/ which should have the same effect).

So with this change we would have the following order:
1) base system: (system config, to be overwritten)
/usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf
2) setUpClass: (static test system config, i.e. management interface eth0, en*)
/etc/NetworkManager/conf.d/90-test-ignore.conf
3) create_devices: (dynamic test interfaces, like eth42, eth43, wlan1, ...)
/etc/NetworkManager/conf.d/99-test-denylist.conf

> [...] you can add additional .conf files to the /etc/NetworkManager/conf.d directory. These will be read in order, with later files overriding earlier ones. Packages might install further configuration snippets to /usr/lib/NetworkManager/conf.d. This directory is parsed first, even before NetworkManager.conf. Scripts can also put per-boot configuration into /run/NetworkManager/conf.d. This directory is parsed second, also before NetworkManager.conf. The loading of a file /run/NetworkManager/conf.d/name.conf can be prevented by adding a file /etc/NetworkManager/conf.d/name.conf. Likewise, a file /usr/lib/NetworkManager/conf.d/name.conf can be shadowed by putting a file of the same name to either /etc/NetworkManager/conf.d or /run/NetworkManager/conf.d.

https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html

## Checklist

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -67,20 +67,19 @@ class IntegrationTestsBase(unittest.TestCase):
         with open('/etc/systemd/network/20-wired.network', 'w') as f:
             f.write('[Match]\nName=eth0 en*\n\n[Network]\nDHCP=yes\nKeepConfiguration=yes')
 
-        # ensure NM doesn't interfere with our test backend (fake eth endpoints & mgmt network)
+        # force-reset NM's unmanaged-devices list (using "=" instead of "+=")
+        # from /usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf,
+        # to stop it from trampling over our test & mgmt interfaces.
+        # https://pad.lv/1615044
         os.makedirs('/etc/NetworkManager/conf.d', exist_ok=True)
-        with open('/etc/NetworkManager/conf.d/99-test-ignore.conf', 'w') as f:
-            f.write('''[keyfile]
-unmanaged-devices+=interface-name:eth0,interface-name:en*,interface-name:veth42,interface-name:veth43''')
+        with open('/etc/NetworkManager/conf.d/90-test-ignore.conf', 'w') as f:
+            f.write('[keyfile]\nunmanaged-devices=interface-name:en*,eth0,nptestsrv')
         subprocess.check_call(['netplan', 'apply'])
         subprocess.call(['/lib/systemd/systemd-networkd-wait-online', '--quiet', '--timeout=30'])
 
     @classmethod
     def tearDownClass(klass):
-        try:
-            os.remove('/run/NetworkManager/conf.d/test-denylist.conf')
-        except FileNotFoundError:
-            pass
+        pass
 
     def tearDown(self):
         subprocess.call(['systemctl', 'stop', 'NetworkManager', 'systemd-networkd', 'netplan-wpa-*',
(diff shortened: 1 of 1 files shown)


tests/integration/base.py
```diff
@@ -444,8 +445,8 @@ def create_devices(klass):
         klass.dev_w_client = devs[1]
 
         # don't let NM trample over our fake AP
-        with open('/run/NetworkManager/conf.d/test-blacklist.conf', 'w') as f:
-            f.write('[main]\nplugins=keyfile\n[keyfile]\nunmanaged-devices+=nptestsrv,%s\n' % klass.dev_w_ap)
+        with open('/etc/NetworkManager/conf.d/99-test-denylist.conf', 'w') as f:
```
schopin-pro, line 448: Out of curiosity, why the change from /run to /etc ?

slyon:
This PR should be integrated with https://git.launchpad.net/~ubuntu-core-dev/netplan/+git/ubuntu/commit/?id=abdfe5deb62ed312dec56a67ae4dae42a348fbf1

tests/integration/base.py
slyon: This is because the changes in /run would be overwritten by our test system's /etc/NetworkManager/conf.d/90-test-ignore.conf, because /etc/ drop-in configs are read after /run/ drop-ins and the /etc/ drop-in uses a hard overwrite of `unmanaged-devices=...` in order to reset the system's default in /usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf (guess we could also downgrade the test systems static /etc/ drop-in to /run/ which should have the same effect).

So with this change we would have the following order:
1) base system: (system config, to be overwritten)
/usr/lib/NetworkManager/conf.d/10-globally-managed-devices.conf
2) setUpClass: (static test system config, i.e. management interface eth0, en*)
/etc/NetworkManager/conf.d/90-test-ignore.conf
3) create_devices: (dynamic test interfaces, like eth42, eth43, wlan1, ...)
/etc/NetworkManager/conf.d/99-test-denylist.conf

> [...] you can add additional .conf files to the /etc/NetworkManager/conf.d directory. These will be read in order, with later files overriding earlier ones. Packages might install further configuration snippets to /usr/lib/NetworkManager/conf.d. This directory is parsed first, even before NetworkManager.conf. Scripts can also put per-boot configuration into /run/NetworkManager/conf.d. This directory is parsed second, also before NetworkManager.conf. The loading of a file /run/NetworkManager/conf.d/name.conf can be prevented by adding a file /etc/NetworkManager/conf.d/name.conf. Likewise, a file /usr/lib/NetworkManager/conf.d/name.conf can be shadowed by putting a file of the same name to either /etc/NetworkManager/conf.d or /run/NetworkManager/conf.d.

https://developer.gnome.org/NetworkManager/stable/NetworkManager.conf.html

slyon:
I conducted some PPA autopkgtest testing to confirm those fixes. It's looking good and I'll merge it.

* https://autopkgtest.ubuntu.com/results/autopkgtest-lunar-slyon-testing/lunar/amd64/n/netplan.io/20230208_132926_35ddf@/log.gz
* https://autopkgtest.ubuntu.com/results/autopkgtest-lunar-slyon-testing/lunar/armhf/n/netplan.io/20230208_115551_087de@/log.gz
* https://autopkgtest.ubuntu.com/results/autopkgtest-lunar-slyon-testing/lunar/ppc64el/n/netplan.io/20230208_115910_124b3@/log.gz
* https://autopkgtest.ubuntu.com/results/autopkgtest-lunar-slyon-testing/lunar/s390x/n/netplan.io/20230208_130047_f7986@/log.gz

PR #224, opened 2021-08-20
Add `.editorconfig` for convenience

nicolasbock:
The `Convenientions` section in `CONTRIBUTING` emphasizes an
indentation of 4 spaces which is not always the default in the
developer's environment. The EditorConfig project aims to maintain
consistent styles across various IDEs and editors. Some editors
natively support EditorConfig [1] and for some editors a plugin to
support EditorConfig can be downloaded [2].

[1] https://editorconfig.org/#pre-installed
[2] https://editorconfig.org/#download

Signed-off-by: Nicolas Bock <nicolas.bock@canonical.com>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.{c,h,py}]
+indent_style = space
+indent_size = 4
+
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2


.editorconfig
```diff
@@ -0,0 +1,5 @@
+root = true
+
+[*.{c,h,py}]
+indent_style = space
+indent_size = 4
```
schopin-pro, line 5:
```suggestion
indent_size = 4

[*.yml]
indent_style = space
indent_size = 4
```

I had to add this in my local .editorconfig to make my editor behave correctly, I figure it might be worth to have?

nicolasbock: Thanks @schopin-pro ! I added your suggestion. I changed the `indent_size` to `2` though since that seems to be the convention currently.

xnox:
```
  yes: please
```

xnox (approved):

PR #226, opened 2021-08-30
clarify requirements for a link local route

kjetilho:
## Description
This gives a better hint on how to add a link local route.  Ie., if the user tries to add a local route (without a gateway (`via`)), they will get an error message that a global route needs `via`, and hopefully understand they should use `scope: local`.  Also, the docs spell this out.

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -585,7 +585,8 @@ eth1:
 ``routes`` (mapping)
 
 :    The ``routes`` block defines standard static routes for an interface.
-     At least ``to`` and ``via`` must be specified.
+     At least ``to`` and ``via`` must be specified.  Routes with scope
+     ``link`` does not need to specify ``via``.
 
      For ``from``, ``to``, and ``via``, both IPv4 and IPv6 addresses are
      recognized, and must be in the form ``addr/prefixlen`` or ``addr``.
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -585,8 +585,8 @@ eth1:
 ``routes`` (mapping)
 
 :    The ``routes`` block defines standard static routes for an interface.
-     At least ``to`` and ``via`` must be specified.  Routes with scope
-     ``link`` does not need to specify ``via``.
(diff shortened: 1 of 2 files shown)


doc/netplan.md
```diff
@@ -585,7 +585,8 @@ eth1:
 ``routes`` (mapping)
 
 :    The ``routes`` block defines standard static routes for an interface.
-     At least ``to`` and ``via`` must be specified.
+     At least ``to`` and ``via`` must be specified.  Routes with scope
+     ``link`` does not need to specify ``via``.
```
schopin-pro, line 589: I feel like keeping the first sentence as-is might be deserving, as someone might overlook the latter sentence contradicting it. What do you think of this alternative?

```suggestion
     At least ``to`` must be specified.  Unless the scope is ``link``, ``via`` is
     also required.
```

schopin-pro (approved):
LGTM, I recall having a discussion on IRC on this particular point last week, which indicates a clear need ;-)

doc/netplan.md
kjetilho: (took a while until I understood "deserving" should be "deceiving" :smile: )

Yes. I think your suggestion is better.  May still be slightly misleading (and this was introduced by me), since the keyword controlling this is not `scope: link`, but `on-link: true`.

```suggestion
     At least ``to`` must be specified.  Unless route is ``on-link``, ``via`` is
     also required.
```

schopin-pro: Actually, I meant "disserving" but my brains and my fingers apparently disagreed ;-).
Your latest suggestion is indeed much better than mine, but I'll have to check if `scope: link` doesn't have the same effect, as I believe it should?

kjetilho: I'm sorry, I confused myself.  Indeed it is `scope: link` which is correct.  `on-link: true` does not work.

slyon (approved):
Thank you for clarifying this bit of our documentation! LGTM as well!

PR #228, opened 2021-09-21
NetDefinition: ownership and cleanup (FR-786)

schopin-pro:
## Description

This PR is mostly about cleaning up the NetplanNetDefinition objects, and when to do it. There are a couple of precursor commits that either made the transition easier or fixed some bugs that were uncovered when actually trying to clean up things ;).

There are no API or ABI change, although there's one observable change with the `dhcp_identifier` which now defaults to `NULL` instead of an owned string `"duid"`. The rationales are in the commit that makes the change.

As usual, each commit log contains some context information related to each change, and the commits are split up with the explicit intent of making it easier to review, one commit at a time.

## Checklist

### DIFF
--- a/src/netdef.c
+++ b/src/netdef.c
@@ -1,0 +1,316 @@
+/*
+ * Copyright (C) 2021 Canonical, Ltd.
+ * Author: Simon Chopin <simon.chopin@canonical.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; version 3.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* This module contains functions to deal with the NetplanNetDefinition objects.
+ * Notably, accessors and destructors.
+ */
+
+#include <glib.h>
+#include "parse.h"
+
+#define FREE_AND_NULLIFY(ptr) { g_free(ptr); ptr = NULL; }
+
+/* Helper function to free a GArray after applying a destructor to its
+ * elements. Note that in the most trivial case (g_free) we should probably
+ * have used a GPtrArray directly... */
+static void
+free_garray_with_destructor(GArray** array, void (destructor)(void *))
+{
+    if (*array) {
+        for (size_t i = 0; i < (*array)->len; ++i) {
+            void* ptr = g_array_index(*array, char*, i);
+            destructor(ptr);
+        }
+        g_array_free(*array, TRUE);
+        *array = NULL;
+    }
+}
+
+/* Helper function to free a GHashTable after applying a simple destructor to its
+ * elements. */
+static void
+free_hashtable_with_destructor(GHashTable** hash, void (destructor)(void *)) {
+    if (*hash) {
+        GHashTableIter iter;
+        gpointer key, value;
+        g_hash_table_iter_init(&iter, *hash);
+        while (g_hash_table_iter_next(&iter, &key, &value))
+            destructor(value);
+        g_hash_table_destroy(*hash);
+        *hash = NULL;
+    }
+}
+
+static void
+free_address_options(void* ptr)
+{
+    NetplanAddressOptions* opts = ptr;
+    g_free(opts->address);
+    g_free(opts->label);
+    g_free(opts->lifetime);
+    g_free(opts);
+}
+
+static void
+free_route(void* ptr)
+{
+    NetplanIPRoute* route = ptr;
+    g_free(route->scope);
+    g_free(route->type);
+    g_free(route->to);
+    g_free(route->from);
+    g_free(route->via);
+    g_free(route);
+}
+
+static void
+free_ip_rules(void* ptr)
+{
+    NetplanIPRule* rule = ptr;
+    g_free(rule->to);
+    g_free(rule->from);
+    g_free(rule);
+}
+
+static void
+free_wireguard_peer(void* ptr)
+{
+    NetplanWireguardPeer* wg = ptr;
+    g_free(wg->endpoint);
+    g_free(wg->preshared_key);
+    g_free(wg->public_key);
+    free_garray_with_destructor(&wg->allowed_ips, g_free);
+    g_free(wg);
+}
+
+static void
+reset_auth_settings(NetplanAuthenticationSettings* auth)
+{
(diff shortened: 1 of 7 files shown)


schopin-pro:
V2: Split off the new reset code into its own file, add some more comments, fix some style issues (`type *val` vs `type* val`, I tend to write the former, the codebase uses the latter)

src/netdef.c
```diff
@@ -0,0 +1,307 @@
...
+    netdef->linklocal.ipv4 = FALSE;
+    netdef->linklocal.ipv6 = TRUE;
+
+    FREE_AND_NULLIFY(netdef->bridge);
+    FREE_AND_NULLIFY(netdef->bond);
+
+    FREE_AND_NULLIFY(netdef->peer);
+
+    netdef->vlan_id = G_MAXUINT; /* 0 is a valid ID */
+    netdef->vlan_link = NULL;
```
slyon, line 228: I think we're missing `has_vlans = FALSE` here.

src/netdef.c
```diff
@@ -0,0 +1,307 @@
...
+    FREE_AND_NULLIFY(netdef->bond_params.up_delay);
+    FREE_AND_NULLIFY(netdef->bond_params.down_delay);
+    FREE_AND_NULLIFY(netdef->bond_params.fail_over_mac_policy);
+    FREE_AND_NULLIFY(netdef->bond_params.primary_reselect_policy);
+    FREE_AND_NULLIFY(netdef->bond_params.learn_interval);
+    FREE_AND_NULLIFY(netdef->bond_params.primary_slave);
+    memset(&netdef->bond_params, 0, sizeof(netdef->bond_params));
+
+    FREE_AND_NULLIFY(netdef->modem_params.apn);
+    netdef->modem_params.auto_config = FALSE;
```
slyon, line 267: That's fine, but why don't we call `memset(..., 0, ...)` , to keep it in line with the cleanup of the other structs?

src/netdef.c
```diff
@@ -0,0 +1,307 @@
...
+    netdef->custom_bridging = FALSE;
+
+    netdef->tunnel.mode = NETPLAN_TUNNEL_MODE_UNKNOWN;
+    FREE_AND_NULLIFY(netdef->tunnel.local_ip);
+    FREE_AND_NULLIFY(netdef->tunnel.remote_ip);
+    FREE_AND_NULLIFY(netdef->tunnel.input_key);
+    FREE_AND_NULLIFY(netdef->tunnel.output_key);
+    FREE_AND_NULLIFY(netdef->tunnel.private_key);
+    netdef->tunnel.fwmark = 0;
+    netdef->tunnel.port = 0;
```
slyon, line 291: Same here. This is fine, but we could be using `memset()` to keep it in line with the cleanup of the other structs. Then move the `netdef->tunnel.mode = NETPLAN_TUNNEL_MODE_UNKNOWN` after `memset()`.

slyon (changes requested):
Simon, thank you very much for another excellent PR!
The cleanup of this big memory leak was long overdue.

I am overall +1 on your changes, but left a few inline comments that I'd like to see addressed before we merge this.

src/netdef.c
schopin-pro: I'm still undecided between `memset()`-ing the substructs vs initializing every component to 0, hence the lack of coherence.

I'll go with `memset()` for now to get this out the doors, but in the future I think I'd rather name the structs and use dedicated `reset_` functions with explicit defaults for each field. It wouldn't add that much more noise anyway :grin:

schopin-pro (commented):
Thanks for the thorough review! There were quite a few nice spots in there, expect a V3 later today with the mentioned changes.

schopin-pro:
V3: Addressed comments in review, and tweaked slightly the first patch to document `passthrough` a bit more.

src/netdef.c
```diff
@@ -0,0 +1,316 @@
...
+    switch (backend) {
+        case NETPLAN_BACKEND_NETWORKD:
+            FREE_AND_NULLIFY(settings->networkd.unit);
+            break;
+        case NETPLAN_BACKEND_NM:
+            FREE_AND_NULLIFY(settings->nm.name);
+            FREE_AND_NULLIFY(settings->nm.uuid);
+            FREE_AND_NULLIFY(settings->nm.stable_id);
+            FREE_AND_NULLIFY(settings->nm.device);
+            g_datalist_clear(&settings->nm.passthrough);
```
slyon, line 162:
```suggestion
            g_datalist_clear(&settings->nm.passthrough);
            settings->nm.passthrough = NULL;
```

src/netdef.c
```diff
@@ -0,0 +1,307 @@
...
+    switch (backend) {
+        case NETPLAN_BACKEND_NETWORKD:
+            FREE_AND_NULLIFY(settings->networkd.unit);
+            break;
+        case NETPLAN_BACKEND_NM:
+            FREE_AND_NULLIFY(settings->nm.name);
+            FREE_AND_NULLIFY(settings->nm.uuid);
+            FREE_AND_NULLIFY(settings->nm.stable_id);
+            FREE_AND_NULLIFY(settings->nm.device);
+            FREE_AND_NULLIFY(settings->nm.passthrough);
```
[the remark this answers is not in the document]
slyon: haha, indeed. that `GData*` type isn't documented very well. And yes, `g_datalist_clear()` seems to be the correct tool for the job here.
But we should probably Nullify the `settings->nm.passthrough` field afterwards, to avoid a dangling pointer.

src/netdef.c
slyon: Yeah.. I was leaning towards explicitly initializing every component to 0 at first as well. But OTOH I also liked the forward-compatibility of `memset()` to initialize everything, including future fields, to 0, similar to how `g_new0()` creates this data structure in the first place.
I'd be fine with both, I guess :-)

src/netdef.c
schopin-pro: Yup, it sets the value to NULL all by itself :) https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gdataset.c#L264

slyon: Alright :+1: nevermind!

slyon:
Ok all unit- and integration-tests are looking good, let's move on with this!

```
$ time autopkgtest ../netplan.io_0.103-0wip99_amd64.changes -U -- qemu /data/autopkgtest-impish-amd64.img
autopkgtest [09:19:58]: starting date: 2021-09-23
autopkgtest [09:19:58]: version 5.16ubuntu1
[...]
autopkgtest [10:20:50]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              PASS
vlans                PASS
wifi                 FLAKY non-zero exit status 1
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
qemu-system-x86_64: terminating on signal 15 from pid 51051 (/usr/bin/python3)

real	60m52,951s
user	36m2,680s
sys	2m23,091s
```

PR #229, opened 2021-09-24
tests: normalize multiple keys on a single dict

schopin-pro:
## Description

The YAML normalizer has some special cases for some given keys. However,
the code block where those are dealt with used `elif` everywhere, which
meant that if a single dict had multiple keys needing to be dealt with,
only the first one would be fixed, and the rest would continue on, still
broken.

This bug is latent in the current codebase, but it showed up in a
patchset I'm working on, where by the game of butterfly effects the
rendered configuration would insist on having proper renderer for each
netdef. This key needs normalizing, but in the test cases where the
netdefs have other problematic properties, for instance the GSM config
in `test_cdma_config`, those pesky renderer lines would still show up
and screw up the test.

## Checklist

### DIFF
--- a/tests/generator/base.py
+++ b/tests/generator/base.py
@@ -174,35 +174,35 @@ class NetplanV2Normalizer():
             if 'password' in keys and ':auth' not in full_key:
                 data['auth'] = {'key-management': 'psk', 'password': data['password']}
                 del data['password']
-            elif 'auth' in keys and data['auth'] == {}:
+            if 'auth' in keys and data['auth'] == {}:
                 data['auth'] = {'key-management': 'none'}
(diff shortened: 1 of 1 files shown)


slyon (approved):
Looks good! Just a small inline suggestion about the `pycheckstyle` formatting that we should improve.

Also, we do have the `skip_generated_yaml_validation` parameter to `generate()` if there are tests that really do not want to play nicely with the normalizer (just FYI).

schopin-pro:
Thanks for the formatting suggestion!

That parameter wouldn't have been right for this, but thanks for the tip, I'll keep it in mind :)

slyon:
Thanks for cleaning it up. Let's merge this!

PR #230, opened 2021-09-24
lib: unify how constant names are exposed and used

schopin-pro:
## Description

This patch moves all the various enum-to-string arrays into a single
compilation unit instead of exposing them directly into headers, and
exposes the data through simple functions.

This approach has multiple advantages:

* If we want to expose one or several of those conversion functions to
  the ABI, we now can add new constants without breaking the ABI, as the
  functions can deal with unknown constants in a sensible manner.
  Directly exposing the arrays in the headers as was previously done was
  a recipy for disaster, as the array is baked into the compiled client code,
  and will not be resized if we decide to add new values to the enums
  in new versions of the library.
* It's marginally faster to compile, as the compiler doesn't have to
  copy around multiple times the same arrays, only to deduplicate them
  at link time (if it even does that).

V2:
* Drop the np_ prefix in favor of the classical netplan_ prefix
* Whitespace fixes
* Copyright headers
* Remove a couple of defines that sneaked up where they didn't belong

## Checklist


src/names.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
slyon, line 1: I guess we should add a copyright header/comment here.

src/names.c
```diff
@@ -0,0 +1,89 @@
...
+
+/* Non-static as we need it for ABI compatibility, see at the end of the file */
+const char* const np_backend_to_str[NETPLAN_BACKEND_MAX_] = {
+        [NETPLAN_BACKEND_NONE] = "none",
+        [NETPLAN_BACKEND_NETWORKD] = "networkd",
+        [NETPLAN_BACKEND_NM] = "NetworkManager",
+        [NETPLAN_BACKEND_OVS] = "OpenVSwitch",
+};
+
+static const char* const np_wifi_mode_to_str[NETPLAN_WIFI_MODE_MAX_] = {
```
slyon, line 14:
```suggestion
static const char* const
netplan1_wifi_mode_to_str[NETPLAN_WIFI_MODE_MAX_] = {
```

Hmm... I don't really like the `np_*` API naming too much, as I think it is too generic. How about using a `netplan1_*` prefix? As we are on libnetplan v0 right now. And this could be the first functions of our libnetplan v1 API. It would even allow to keep backwards compatibility using `libnetplan1_X` and `libnetplan2_X` if we want/need to break API in the future.
What do you think?

Also, starting the symbol name in a new line is a good practice/code-style IMO. We're not fully consistent with that in the current codebase, but let's try to stick to that convention for new definitions. (That applies to all functions in this file)

src/names.c
```diff
@@ -0,0 +1,89 @@
...
+
+#include "names.h"
+#include "parse.h"
+
+/* Non-static as we need it for ABI compatibility, see at the end of the file */
+const char* const np_backend_to_str[NETPLAN_BACKEND_MAX_] = {
+        [NETPLAN_BACKEND_NONE] = "none",
+        [NETPLAN_BACKEND_NETWORKD] = "networkd",
+        [NETPLAN_BACKEND_NM] = "NetworkManager",
+        [NETPLAN_BACKEND_OVS] = "OpenVSwitch",
```
slyon, line 11:
```suggestion
    [NETPLAN_BACKEND_NONE] = "none",
    [NETPLAN_BACKEND_NETWORKD] = "networkd",
    [NETPLAN_BACKEND_NM] = "NetworkManager",
    [NETPLAN_BACKEND_OVS] = "OpenVSwitch",
```
Let's fix up this whitespace, while on it :)

src/names.c
```diff
@@ -0,0 +1,89 @@
+#include <glib.h>
```
slyon, line 1: Let's add a copyright header/comment here.

src/parse.c
```diff
@@ -29,9 +29,13 @@
 
 #include "parse.h"
 #include "util.h"
+#include "names.h"
 #include "error.h"
 #include "validation.h"
 
+#define NETPLAN_VERSION_MIN	2
+#define NETPLAN_VERSION_MAX	3
```
slyon, line 37: We do we add these definitions here and keeping them in `parse.h` at the same time?

slyon (changes requested):
Thank you Simon, I really like the consolidation of all `*_name` functions and moving them into a single compilation unit!

I do not really like the `np_*` API naming, though, as I think it is too generic. What do you think about `netplan1_` (for the pending v1 API)?

I left some other smaller inline comments about formatting and missing copyright headers.

src/names.c
schopin-pro: ACK for the symbol name at the start of the line.

After a quick verbal exchange with @slyon and @vorlonofportland we'll stick with `netplan_` as the prefix of our public API, as it is the more elegant solution from the user's PoV, even if it makes my life a bit harder ;-). I'll update this PR (and all the others...) shortly.

src/names.c
schopin-pro: At some point I'll cook up a clang-format file, it'll save us both some time :)

src/parse.c
schopin-pro: These shouldn't be there, and are gone in V2 :)

schopin-pro:
Posted a V2, with the following changelog: 

V2:
* Drop the np_ prefix in favor of the classical netplan_ prefix
* Whitespace fixes
* Copyright headers
* Remove a couple of defines that sneaked up where they didn't belong

src/names.c
```diff
@@ -0,0 +1,114 @@
...
+NAME_FUNCTION(def_type, NetplanDefType);
+NAME_FUNCTION(auth_key_management_type, NetplanAuthKeyManagementType);
+NAME_FUNCTION(auth_eap_method, NetplanAuthEAPMethod);
+NAME_FUNCTION(tunnel_mode, NetplanTunnelMode);
+NAME_FUNCTION(addr_gen_mode, NetplanAddrGenMode);
+NAME_FUNCTION(wifi_mode, NetplanWifiMode);
+
+/* ABI compatibility definitions */
+
+const char *
```
slyon, line 110:
```suggestion
const char*
```
Just a tiny nitpick...

slyon:
Thanks for the V2 updates. I only left a tiny formatting nitpick comment. Could you also rebase on top of the latest `main` branch?
We're then ready to merge this first part of our new API! Kudos!

slyon (approved):
Thanks for the V2 updates. I only left a tiny formatting nitpick comment. Could you also rebase on top of the latest main branch?
We're then ready to merge this first part of our new API! Kudos!

slyon:
Actually... I just did that myself. Let's merge this!

PR #234, opened 2021-09-30
parser: fix some error handling code

schopin-pro:
## Description

This PR rework some error handling paths, to achieve two results:

* Avoid as much as possible putting data in a netdef before it has been validated. This makes recoverable error handling much easier, and also facilitates debugging: reducing side effects means enhanced local reasoning, resulting in less headaches ;)
* Be consistent regarding `GError**` acceptable values. Most of the code assumes that actual error signalling is done via the value returned on the stack, either via `FALSE` or `NULL`, but a couple of places would use the GError mechanisms instead. This meant that providing `GError** error = NULL` would mostly work, but for those places where the code would glance over some failures as it wouldn't be able to detect them.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1107,6 +1107,7 @@ handle_wifi_access_points(yaml_document_t* doc, yaml_node_t* node, const void* d
 {
     for (yaml_node_pair_t* entry = node->data.mapping.pairs.start; entry < node->data.mapping.pairs.top; entry++) {
         yaml_node_t* key, *value;
+        gboolean ret = TRUE;
 
         key = yaml_document_get_node(doc, entry->key);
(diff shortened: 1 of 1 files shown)


src/parse.c
```diff
@@ -2677,19 +2675,24 @@ netplan_finish_parse(GError** error)
...
         }
-        g_hash_table_foreach(netdefs, finish_iterator, error);
-    }
-
-    if (error && *error)
-        return NULL;
+        g_hash_table_iter_init (&iter, netdefs);
 
+        while (g_hash_table_iter_next (&iter, &key, &value))
+        {
```
slyon, line 2690:
```suggestion
        while (g_hash_table_iter_next (&iter, &key, &value)) {
```
Tiny code-style nitpick...

PR #236, opened 2021-10-02
netplan: set: make it possible to unset a whole devtype subtree (LP: #1942930) (FR-1685)

schopin-pro:
## Description

Previously, trying to unset a whole devtype subtree, such as `netplan set network.ethernets=null` would fail. This PR fixes it by detecting this special case and manually deleting each netdef assigned to the subtype.

This approach is far from being performant, but it has the merit of working without needing too much new code.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: LP#1942930

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -334,3 +334,47 @@ get_global_network(int ip_family)
     else
         return "::/0";
 }
+
+struct netdef_pertype_iterator {
+    NetplanDefType type;
+    GHashTableIter iter;
+};
+
+NETPLAN_INTERNAL struct netdef_pertype_iterator*
+_netplan_iter_defs_per_devtype_init(const char *devtype)
+{
+    NetplanDefType type = netplan_def_type_from_name(devtype);
+    struct netdef_pertype_iterator *iter = g_malloc0(sizeof(*iter));
+    iter->type = type;
+    if (netdefs)
+        g_hash_table_iter_init(&iter->iter, netdefs);
+    return iter;
+}
+
+NETPLAN_INTERNAL NetplanNetDefinition*
+_netplan_iter_defs_per_devtype_next(struct netdef_pertype_iterator* it)
+{
+    gpointer key, value;
+
+    if (!netdefs)
+        return NULL;
+
+    while (g_hash_table_iter_next(&it->iter, &key, &value)) {
+        NetplanNetDefinition* netdef = value;
+        if (netdef->type == it->type)
+            return netdef;
+    }
+    return NULL;
+}
+
+NETPLAN_INTERNAL void
+_netplan_iter_defs_per_devtype_free(struct netdef_pertype_iterator* it)
+{
+    g_free(it);
+}
+
+NETPLAN_INTERNAL const char*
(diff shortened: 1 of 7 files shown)


schopin-pro:
To whoever wants to backport this onto older releases, the first commit can be dropped, the second will mostly likely need some form of rewrite as it builds upon very recent work (#230). AFAICT the last two commits shouldn't pose any problem.

src/util.c
```diff
@@ -334,3 +334,47 @@ get_global_network(int ip_family)
...
+}
+
+NETPLAN_INTERNAL void
+_netplan_iter_defs_per_devtype_free(struct netdef_pertype_iterator* it)
+{
+    g_free(it);
+}
+
+NETPLAN_INTERNAL const char*
+_netplan_netdef_id(NetplanNetDefinition* nd)
```
slyon, line 377: This is a rather generic function, that will probably be needed later on. Should we define it to be part of the `NETPLAN_PUBLIC` API and call it `netplan_netdef_get_id`?
We could still do this later on (once it's actually needed), as this is currently for internal use only.

tests/test_cli_get_set.py
```diff
@@ -195,6 +193,21 @@ def test_set_delete(self):
...
+    def test_set_delete_subtree(self):
+        with open(self.path, 'w') as f:
+            f.write('''network:\n  version: 2\n  renderer: NetworkManager
+  ethernets:
+    eth0: {addresses: [1.2.3.4/24]}''')
+        self._set(['network.ethernets=null'])
+        self.assertTrue(os.path.isfile(self.path))
+        with open(self.path, 'r') as f:
+            out = f.read()
+            # print(out, flush=True) # debugging
```
slyon, line 205: This seems to be a copy & pasted left over from my debugging attempts a while ago. We can probably drop it, but it doesn't hurt in here either.

slyon (approved):
Thank you Simon, I like how you're using libnetplan's internal data structure here, instead of relying on the deprecated `configmanager.py`.

The performance isn't good, but still I like that you implemented it in a way that can be backported (relatively) easy, without relying on all the new API code. Clearing a full devtype subtree isn't on the critical path, so this is fine IMO.

+1 from my side! I left two small inline comments, but nothing that would block merging of this PR.

src/util.c
schopin-pro: I indeed deliberately chose a symbol name that would be explicitly "unstable" (starting with `_`) because even though it is a generic function, I wasn't sure about the naming and didn't want to start shaving this particular yak in this PR ;).

When we settle on a proper scheme for the public accessors (I'm reserved about your proposal ;-) ) we can just alias this one or drop it completely if we decide on lockstep upgrade of `netplan.io` and `libnetplanX`.

tests/test_cli_get_set.py
schopin-pro: Removed, and I credited you for the tests, as is proper :)

slyon: Ha! I didn't even remember that dropped that test code on LP the other day xD

PR #237, opened 2021-10-06
feat: mdns configuration

eaglesemanation:
Fix for [\#1830507](https://bugs.launchpad.net/netplan/+bug/1830507)

## Description
Adds mdns boolean flag for all connection types, which enables multicast DNS support on specific link

## Checklist

- [x] Retains 100% code coverage (`make check-coverage`). (Actually 53%, same as before changes)

### DIFF
--- a/doc/example-config
+++ b/doc/example-config
@@ -10,6 +10,7 @@ network:
         macaddress: 00:11:22:33:44:55
       wakeonlan: true
       dhcp4: true
+      mdns: true
       addresses:
         - 192.168.14.2/24
         - "2001:1::1/64"
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -428,6 +428,11 @@ similar to ``gateway*``, and ``search:`` is a list of search domains.
               search: [lab, home]
               addresses: [8.8.8.8, "FEDC::1"]
 
+``mdns`` (bool)
+
+:   Enables multicast DNS support. That allows IP address resolution by hostname
+in networks without dedicated DNS server.
+
 ``macaddress`` (scalar)
 
 :   Set the device's MAC address. The MAC address must be in the form
@@ -910,7 +915,7 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
           the ``band`` property is also set.
 
      ``hidden`` (bool) – since **0.100**
-     :    Set to ``true`` to change the SSID scan technique for connecting to 
+     :    Set to ``true`` to change the SSID scan technique for connecting to
           hidden WiFi networks. Note this may have slower performance compared
           to ``false`` (the default) when connecting to publicly broadcast
           SSIDs.
--- a/src/parse.c
+++ b/src/parse.c
@@ -2252,6 +2252,7 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"macaddress", YAML_SCALAR_NODE, handle_netdef_mac, NULL, netdef_offset(set_mac)},        \
     {"mtu", YAML_SCALAR_NODE, handle_netdef_guint, NULL, netdef_offset(mtubytes)},            \
     {"nameservers", YAML_MAPPING_NODE, NULL, nameservers_handlers},                           \
+    {"mdns", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(mdns)},                \
     {"optional", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(optional)},        \
     {"optional-addresses", YAML_SEQUENCE_NODE, handle_optional_addresses},                    \
     {"renderer", YAML_SCALAR_NODE, handle_netdef_renderer},                                   \
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -687,6 +687,8 @@ write_network_file(const NetplanNetDefinition* def, const char* rootdir, const c
     if (def->ip6_nameservers)
         for (unsigned i = 0; i < def->ip6_nameservers->len; ++i)
             g_string_append_printf(network, "DNS=%s\n", g_array_index(def->ip6_nameservers, char*, i));
+    if (def->mdns)
+        g_string_append_printf(network, "MulticastDNS=yes\n");
     if (def->search_domains) {
         g_string_append_printf(network, "Domains=%s", g_array_index(def->search_domains, char*, 0));
         for (unsigned i = 1; i < def->search_domains->len; ++i)
--- a/src/nm.c
+++ b/src/nm.c
@@ -608,6 +608,8 @@ write_nm_conf_access_point(NetplanNetDefinition* def, const char* rootdir, const
         uuid_unparse(def->uuid, uuidstr);
         g_key_file_set_string(kf, "connection", "uuid", uuidstr);
     }
+    if (def->mdns)
+        g_key_file_set_string(kf, "connection", "mdns", "yes");
 
     if (def->activation_mode) {
         /* XXX: For now NetworkManager only supports the "manual" activation
--- a/src/types.h
+++ b/src/types.h
@@ -200,6 +200,7 @@ struct net_definition {
     char* gateway6;
     GArray* ip4_nameservers;
     GArray* ip6_nameservers;
+    gboolean mdns;
     GArray* search_domains;
     GArray* routes;
     GArray* ip_rules;


doc/netplan.md
```diff
@@ -428,6 +428,11 @@ similar to ``gateway*``, and ``search:`` is a list of search domains.
               search: [lab, home]
               addresses: [8.8.8.8, "FEDC::1"]
 
+``mdns`` (bool)
```
slyon, line 431:
```suggestion
``mdns`` (bool) – since **0.104**
```
Please mention the version that this feature will be introduced in.

doc/example-config
```diff
@@ -10,6 +10,7 @@ network:
         macaddress: 00:11:22:33:44:55
       wakeonlan: true
       dhcp4: true
+      mdns: true
```
slyon, line 13: Thank you for providing an example! Could you please additionally add a test case for NetworkManager and a tests case for networkd in `tests/generator/test_common.py` in order to keep the coverage at 100%?

slyon (changes requested):
Thank you very much @eaglesemanation for contributing to netplan and implementing this long standing feature request! Your work is much appreciated!

Overall the changes look good to me, but I think we need to change the type of this data to an `enum` instead of a `boolean`, to reflect its different states and fully support this feature.

Additionally, I'd like to ask you to add test cases around the new code, in order to keep a good code coverage. See:
https://app.codecov.io/gh/canonical/netplan/compare/237/diff#diff-c3JjL25ldHdvcmtkLmM=
https://app.codecov.io/gh/canonical/netplan/compare/237/diff#diff-c3JjL25tLmM=

Other than this I only left some smaller inline comments.

src/parse.c
```diff
@@ -2252,6 +2252,7 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"macaddress", YAML_SCALAR_NODE, handle_netdef_mac, NULL, netdef_offset(set_mac)},        \
     {"mtu", YAML_SCALAR_NODE, handle_netdef_guint, NULL, netdef_offset(mtubytes)},            \
     {"nameservers", YAML_MAPPING_NODE, NULL, nameservers_handlers},                           \
+    {"mdns", YAML_SCALAR_NODE, handle_netdef_bool, NULL, netdef_offset(mdns)},                \
```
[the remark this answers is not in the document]
slyon: Yes, that would be a nice solution IMO! Using `handle_generic_bool` inside your custom handler function, checking for `TRUE/FALSE` and then setting the enum accordingly.

tests/generator/test_common.py
```diff
@@ -685,6 +685,48 @@ def test_nameserver(self):
...
+LinkLocalAddressing=ipv6
+MulticastDNS=yes
+''',
+                              'eth2.network': '''[Match]
+Name=eth2
+
+[Network]
+LinkLocalAddressing=ipv6
+MulticastDNS=no
+'''})
```
slyon, line 720:
```suggestion
        self.assert_networkd({'eth0.network': (ND_EMPTY % ('eth0', 'ipv6'))
                              .replace('ConfigureWithoutCarrier=yes', 'MulticastDNS=resolve'),
                              'eth1.network': (ND_EMPTY % ('eth1', 'ipv6'))
                              .replace('ConfigureWithoutCarrier=yes', 'MulticastDNS=yes'),
                              'eth2.network': (ND_EMPTY % ('eth2', 'ipv6'))
                              .replace('ConfigureWithoutCarrier=yes', 'MulticastDNS=no')})
```
Trying to shrink this a bit...

tests/generator/test_common.py
```diff
@@ -1308,6 +1350,65 @@ def test_nameserver(self):
...
+        self.generate('''network:
+  version: 2
+  renderer: NetworkManager
+  ethernets:
+    eth0:
+      mdns: resolve
+    eth1:
+      mdns: true
+    eth2:
+      mdns: false
```
slyon, line 1367:
```suggestion
    eth0: {mdns: resolve}
    eth1: {mdns: true}
    eth2: {mdns: false}
```
See above.

src/parse.c
```diff
@@ -2252,6 +2269,7 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"macaddress", YAML_SCALAR_NODE, handle_netdef_mac, NULL, netdef_offset(set_mac)},        \
     {"mtu", YAML_SCALAR_NODE, handle_netdef_guint, NULL, netdef_offset(mtubytes)},            \
     {"nameservers", YAML_MAPPING_NODE, NULL, nameservers_handlers},                           \
+    {"mdns", YAML_SCALAR_NODE, handle_mdns, NULL, netdef_offset(mdns)},                       \
```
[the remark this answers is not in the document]
vorlonofportland: Thanks, I do like it better spelled out as 'multicast-dns'.

My greater concern is whether this is something netplan should be doing at all.  Yes, systemd-networkd CAN do mdns, but is this the right component on an Ubuntu system (the primary, though not exclusive, target for netplan) to handle mnds?  We also have libnss-mdns for resolution, installed by default on the desktop; which pulls in avahi-daemon, which binds to the mdns port by default.  So, by configuring mdns through systemd-networkd, are we saying that we should stop shipping libnss-mdns and avahi-daemon by default?  (Note, systemd-networkd's service broadcast support is significantly less featureful than avahi's.)  Or, if we should not replace avahi with systemd-networkd's functionality, should this netplan option be used to drive configuration of avahi+nss_mdns instead?

These are larger architectural questions than just the schema change, and should be resolved first.

eaglesemanation:
I don't think I'm capable of implementing some sort of avahi configuration in any meaningful amount of time. Unlike what I did it would actually require to write my own code, instead of copy pasting and modifying a bit. 

But I think if Ubuntu is advertising netplan as their main way of configuring networking - then it should be capable of doing so, and mDNS is clearly a part of that. So maybe there is need to asses if Ubuntu even needs any features that avahi provides over systemd-resolved. I wasn't able to find any docs that would describe full feature set of each software to compare them, but according to some random Reddit post the only difference is lack of discovery functionality in resolved (please correct me if I'm wrong). If that's the case and there is no software in Ubuntu that relies on that functionality - I would propose removing avahi+nss_mdns, because less software == better :)

src/parse.c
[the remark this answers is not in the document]
jblachly: @vorlonofportland one must separate the goals of resolution and response. No one who is clamoring for netplan to support a `mdns` flag is suggesting AFAICT to remove avahi (which as you observe, is generally superior for purpose).

The problem is this: lots of software apparently does not use libc resolution (for which mdns can be enabled in `/etc/nsswitch.conf`), but instead uses systemd-resolved.  In this case, mdns resolution works for everything EXCEPT that which uses systemd-resolved (127.0.0.53:53) , which is maddening.

jblachly:
> > But I think if Ubuntu is advertising netplan as their main way of configuring networking - then it should be capable of doing so, and mDNS is clearly a part of that. So maybe there is need to asses if Ubuntu even needs any features that avahi provides over systemd-resolved.
> 
> I agree with your point here. But I guess that needs some bigger investigation/discussion about removing (or rather replacing) avahi+nss_mdns

There is no need to remove avahi+nss_mdns.  One can simply set the systemd-resolved mDNS support to "resolve" (not "yes" or true).

There is a horrifically maddening mix of software, yes, but I wouldn't throw the baby out with the bathwater, and avahi-daemon is generally more featureful. However, as I pointed out in a comment above, there is apparently a lot of software that doesn't use libc resolution (and hence mdns even if included in `/etc/nsswitch.conf`), but instead connects directly to 127.0.0.53:53, and that SUBSET of software can't make use of mdns, which is maddening.

https://unix.stackexchange.com/questions/459991/how-to-configure-systemd-resolved-for-mdns-multicast-dns-on-local-network

slyon:
FTR: "systemd's resolved only handles mDNS hostname lookups (something.local) and not generic service discovery. [...] systemd supports registering a SRV and TXT record for a systemd service (e.g. a _ipps-system._tcp service registration for cupsd, if it supported the IPP System Service), but not "sub-services" under that systemd service (like the individual printer registrations to _ipp._tcp and _ipps._tcp), nor does it support DNS-SD sub-types (like _print._sub._ipp._tcp for IPP Everywhere), flagship naming (_printer._tcp is used to reserve service instance names for network printers)." https://github.com/apple/cups/issues/5452#issuecomment-625217203

I like the idea of advertising the `resolve` value and printing a warning (about concurrent use of avahi & systemd-resolved) if `yes` is being used...

moorchegue:
Any chance for this to get merged?

Paradroid888:
Just upgraded to 2204, lost mDNS and arrived here after hours of digging. Respectfully disagree with you @vorlonofportland, how does exposing a flag to turn an existing feature on, for users who know what they want to use, cause wider architectural issues? Isn't it just a choice?

Any chance of reconsidering this PR please?

jonhnet:
Another bid for getting this done. It's a bottleneck to automating the installation of an SBC. I get to set up a netplan on the config partition when (automatically) burning an SD/EMMC image. At some point, I need to boot the SBC and talk to it to run the rest of the install. Absent mdns, that's a difficult-to-automate proposition.

PR #238, opened 2021-10-06
parse-nm: Handle missing gateway in keyfile routes, keep dns-search fallback

slyon:
## Description
NM assumes a `scope: link` route if the gateway is empty or unspecified (i.e. "0.0.0.0"/"::") in keyfile.
E.g. the route is only valid on the local network:
`ip route add NETWORK dev DEVICE [metric METRIC]`
see https://github.com/NetworkManager/NetworkManager/blob/main/src/libnm-core-impl/nm-keyfile.c#L520

netplan cannot differentiate between `ipv4.dns-search` and `ipv6.dns-search` so keep it in the passthrough/fallback list as an override.

## Checklist

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -220,7 +220,6 @@ write_routes(const NetplanNetDefinition* def, GKeyFile *kf, int family)
         for (unsigned i = 0, j = 1; i < def->routes->len; ++i) {
             const NetplanIPRoute *cur_route = g_array_index(def->routes, NetplanIPRoute*, i);
             const char *destination;
-            const char *via;
 
             if (cur_route->family != family)
                 continue;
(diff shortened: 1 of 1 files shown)


schopin-pro:
Just to be clear, this solves 2 separate issues, right?

schopin-pro (dismissed):
LGTM, although I'm not sure how I could easily test those changes :)

src/parse-nm.c
```diff
@@ -225,6 +225,11 @@ parse_routes(GKeyFile* kf, const gchar* group, GArray** routes_arr)
         /* Append gateway/via IP */
         if (split[0] && split[1])
             route->via = g_strdup(split[1]); //no need to free, will stay in netdef
+        else {
+            /* NM assumes a "default" route if this is not defined in the keyfile.
+             * See nm-settings-ip-config.c -> nm_ip_route_set_next_hop()/canonicalize_ip(). */
+            route->via = g_strdup("default");
```
schopin-pro, line 231: As mentioned in a private chat, I have doubts regarding this. The concept of "default" to designate a gateway doesn't make sense to me, and we suspect this is actually used to specify local routes

src/parse-nm.c
```diff
@@ -225,6 +225,14 @@ parse_routes(GKeyFile* kf, const gchar* group, GArray** routes_arr)
         /* Append gateway/via IP */
         if (split[0] && split[1])
             route->via = g_strdup(split[1]); //no need to free, will stay in netdef
+        else {
+            /* If the gateway (via) is unspecified, it means that this route is
+             * only valid on the local network (see nm-keyfile.c ->
+             * read_one_ip_address_or_route()), e.g.:
+             * ip route add NETWORK dev DEV [metric METRIC] */
+            route->via = g_strdup(get_unspecified_address(route->family));
```
schopin-pro, line 233: Shouldn't this be 
```suggestion
            route->via = NULL;
            g_free(route->scope);
            route->scope = g_strdup("link");
```

slyon:
ACK, fair point!
NetworkManager is kind of special in this regard, as it does not have an understanding of "scope" in it's keyfile format, but it auto-detects a route's scope depending on destination IP and gateway fields. If gateway/via is empty or unspecified (e.g. ""/"0.0.0.0"/"::") it assumes a "link" or "host" scope, while it assumes a "global" scope if the gateway is valid. Up to now netplan has been ignoring this fact and just errored out if the scope was not "global".

But you're absolutely right that this implementation detail of NM should not leak into the netplan YAML schema by specifying an invalid "via" field. But rather we want to keep that local to the NM generator backend. So I've added another commit that reworks the scope logic of the NM backend, to accept all values ("global"/"link"/"host") and write the keyfile accordingly (proper gateway/via field or unspecified address if scope = "link" or "host"), tricking NM to do the right thing.

src/nm.c
```diff
@@ -234,20 +235,23 @@ write_routes(const NetplanNetDefinition* def, GKeyFile *kf, int family)
...
+             * For IPv4 addresses, NetworkManager determines the scope of addresses on its own
+             * ("link"/"host" for addresses without gateway, "global" for addresses with next-hop).
+             * Routes without gateway are represented by the unspecified address in keyfile. */
+            if (g_strcmp0(cur_route->scope, "global") == 0)
+                via = cur_route->via;
+            else {
+                g_debug("%s: Overriding 'via: %s' as NetworkManager does not support "
+                        "setting a route's scope directly, but will auto-detect them.",
+                        def->id, get_unspecified_address(cur_route->family));
+                via = get_unspecified_address(cur_route->family);
```
schopin-pro, line 248: That's not what I read there https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/libnm-core-impl/nm-keyfile.c#L506

Wouldn't we want 
```suggestion
                via = "";
```
instead ?

schopin-pro (approved):

PR #239, opened 2021-10-11
libnetplan: expose coherent generator APIs

schopin-pro:
## Description

New APIs for the generator APIs of libnetplan, with proper error handling and support for the NetplanState feature.

API and ABI compatibility is retained for now. Depends on #232 and #234, as usual see the commits after the merge.

## Checklist


src/nm.h
```diff
@@ -19,11 +19,27 @@
...
+NETPLAN_INTERNAL gboolean
+netplan_state_finish_nm_write(
+        const NetplanState* np_state,
+        const char* rootdir,
+        GError** error);
+
+NETPLAN_INTERNAL gboolean
+netplan_nm_cleanup(const char* rootdir);
+
+
```
slyon, line 39: Add comment about deprecated API, instead of additional whitespace.

src/networkd.c
```diff
@@ -1053,27 +1069,28 @@ write_wpa_conf(const NetplanNetDefinition* def, const char* rootdir)
...
+    return TRUE;
 }
 
-/**
- * Generate networkd configuration in @rootdir/run/systemd/network/ from the
- * parsed #netdefs.
- * @rootdir: If not %NULL, generate configuration in this root directory
- *           (useful for testing).
- * Returns: TRUE if @def applies to networkd, FALSE otherwise.
- */
```
slyon: We should probably keep that comment in `networkd.c` (instead of abi_compat.c)

src/networkd.c
```diff
@@ -1137,29 +1156,3 @@ cleanup_networkd_conf(const char* rootdir)
      * upgraded system, we need to make sure to clean those up. */
     unlink_glob(rootdir, "/run/systemd/system/systemd-networkd.service.wants/netplan-wpa@*.service");
 }
-
-/**
- * Create enablement symlink for systemd-networkd.service.
- */
```
slyon, line 1143: Let's not drop this comment, but move it to `generate.c`, too.

PR #240, opened 2021-10-12
networkd:wpa_supplicant: driver fallback to nl80211 and/or wext (LP: #1814012)

kees:
## Description

Adds tests to check the generated content of the wpa_supplicant .service files.

Fixes non-wired wpa_supplicant .service definition to use correct driver fallback list:

The default behavior for wpa_supplicant under systemd is to try both
nl80211 and wext drivers[1]. However, netplan was not specifying the
the same configuration, so wext devices had no way to be configured[2].

[1] https://salsa.debian.org/debian/wpa/-/blob/debian/unstable/debian/patches/networkd-driver-fallback.patch
[2] https://bugs.launchpad.net/netplan/+bug/1814012

Signed-off-by: Kees Cook <kees@ubuntu.com>

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -46,11 +46,11 @@ wifi_append_freq(gpointer key, gpointer value, gpointer user_data)
 /**
  * append wowlan_triggers= string for wpa_supplicant.conf
  */
-static void
-append_wifi_wowlan_flags(NetplanWifiWowlanFlag flag, GString* str) {
+static gboolean
+append_wifi_wowlan_flags(NetplanWifiWowlanFlag flag, GString* str, GError** error) {
     if (flag & NETPLAN_WIFI_WOWLAN_TYPES[0].flag || flag >= NETPLAN_WIFI_WOWLAN_TCP) {
-        g_fprintf(stderr, "ERROR: unsupported wowlan_triggers mask: 0x%x\n", flag);
-        exit(1);
+        g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT, "ERROR: unsupported wowlan_triggers mask: 0x%x\n", flag);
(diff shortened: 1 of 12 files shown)


tests/generator/test_auth.py
```diff
@@ -225,6 +225,21 @@ def test_auth_wired(self):
...
+Description=WPA supplicant for netplan eth0
+DefaultDependencies=no
+Requires=sys-subsystem-net-devices-eth0.device
+After=sys-subsystem-net-devices-eth0.device
+Before=network.target
+Wants=network.target
+
+[Service]
+Type=simple
+ExecStart=/sbin/wpa_supplicant -c /run/netplan/wpa-eth0.conf -ieth0 -Dwired
```
slyon, line 240: We should probably move this into a template string (e.g. `SD_WPA` in `tests/generator/base.by`), to combine the common pattern (filling it only with the relevant interface name and `ExecStart=` line.

This could then be reused in `test_wifis.py` as well, so we do not repeat ourselves too much.

kees: Ah-ha, good idea. I will prepare an updated commit. Thanks!

src/networkd.c
```diff
@@ -975,8 +976,12 @@ write_wpa_unit(const NetplanNetDefinition* def, const char* rootdir)
 
     if (def->type != NETPLAN_DEF_TYPE_WIFI) {
         g_string_append(s, " -Dwired\n");
+    } else {
+        g_string_append(s, " -Dnl80211,wext\n");
     }
+    orig_umask = umask(022);
```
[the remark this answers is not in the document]
kees: I have split the umask change into a separate patch. I actually think all these patches should be backported, but I'll leave that to the SRU folks. :)

PR #241, opened 2021-10-25
Consolidate the YAML parsing into libnetplan: low-hanging fruits (FR-702)

schopin-pro:
## Description

Until now, we parsed the YAML configuration at different places in the program:

* libnetplan, where we have most of our business logic, and where we implement data verification
* the `netplan set` command, where we parse the YAML tree to do some syntactical merging of the trees without looking too much into the semantics
* `configmanager.py` + `sriov.py` where we have some business logic to be able to configure the SRIOV backend on the fly.
* `netplan migrate`, which isn't supposed to be used in normal circumstances

Additionally, we use PyYAML in `netplan info` to easily generate well-formed YAML data from a Python dict.

This PR removes the `migrate` and `info` use cases, the other cases will have their own PRs in order to keep review complexity somewhat manageable.

### Why?

There are a couple of advantages to this new approach:

* Single point of validation: whenever we want to parse some configuration, we'll be able to validate the new configuration in a consistent manner
* Smaller dependency tree: netplan is intended for all sorts of use-case, including some where disk size is a real constraint. python3-yaml has an unpacked size of 0.5 Mo, which can now be shaved off.
* Better APIs: this is more of a by-product, but doing this forced us to implement proper APIs for libnetplan ;-)

## Checklist

### DIFF
--- a/netplan/cli/commands/info.py
+++ b/netplan/cli/commands/info.py
@@ -62,7 +62,7 @@ class NetplanInfo(utils.NetplanCommand):
 
         elif self.version_format == 'yaml':
             print('''netplan.io:
-  website: "https://netplan.io"
-  features:''')
+  website: "{}"
+  features:'''.format(netplan_version['netplan.io']['website']))
             for feature in netplan._features.NETPLAN_FEATURE_FLAGS:
                 print('  - ' + feature)
--- a/netplan/cli/commands/info.py
+++ b/netplan/cli/commands/info.py
@@ -61,5 +61,8 @@ class NetplanInfo(utils.NetplanCommand):
             print(json.dumps(netplan_version, indent=2))
 
         elif self.version_format == 'yaml':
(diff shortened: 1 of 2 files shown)


schopin-pro:
I've reworked the code to use file descriptors where it made sense instead of giving out file names, and the code coverage should reach 100% (with some error paths excluded as really hard to trigger in the tests). I also added some documentation for the non-trivial new APIs.

Now all that remains (besides merging #233) is the reviewability issue.

schopin-pro:
This PR has just undergone a drastic diet to get things moving, and is now just the low hanging fruits, the main course will come in subsequents PRs. As such, I'm removing the draft status and move it up to review.

slyon (approved):
This is a very nice PR description! Thank you, LGTM.
I left a tiny nitpick inline, feel free to merge regardless.

PR #231, opened 2021-09-24
Implement YAML state tracking and use it in the DBus API and netplan-try (LP: #1943120) (FR-1745)

slyon:
## Description
Allow to pass an optional `--state` parameter to `netplan try/apply` describing a directory that contains a netplan configuration tree/state (i.e. `/{etc,run,lib}/netplan/*.yaml`).
Netplan will make use of this "old state" to calculate the delta of dropped interface definitions, like bridges/bonds/vlans/tunnels that have been configured before but are not part of the current YAML configuration anymore. It will then try to delete those virtual interfaces (via `ip link del dev IFACE`) if they still exist.

The same functionality is used to roll back a `netplan try` command that failed or was rejected.

Generally, the state needs to be provided manually. The DBus API (using io.netplan.Netplan.Config.Try/Apply) is an exception, as the previous state can be backed up automatically in this case.

## Checklist

- [x] Runs `make check` successfully.
- [x] Retains 100% code coverage (`make check-coverage`).
- [ ] New/changed keys in YAML format are documented.
- [ ] \(Optional\) Adds example YAML for new feature.
- [x] \(Optional\) Closes an open bug in Launchpad: LP#1943120

PR #243, opened 2021-11-03
cli:apply: initialize self.state (LP: #1949104)

slyon:
## Description
The `self.state` attribute needs to be available even if NetplanApply() is not executed via the command line (and thus the `run()` method is not executed).

Fixes a regression in #231

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: LP#1949104

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -44,6 +44,7 @@ class NetplanApply(utils.NetplanCommand):
                          leaf=True)
         self.sriov_only = False
         self.only_ovs_cleanup = False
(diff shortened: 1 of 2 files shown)


schopin-pro (changes requested):
LGTM, but it'd be even nicer with a test that triggers the fixed bug :)

slyon:
I agree. `netplan try` modifies the running system, so we cannot easily wrap this into an unit test. But I've created two new integration tests that execute `netplan try` and accept and reject the configuration accordingly, checking for the correct output.

schopin-pro (approved):

PR #244, opened 2021-11-10
Disable temporary address generation when renderer is NetworkManager (LP: #1948027) – change in behavior

peleroux:
…and ipv6-privacy is false

## Description

Fix for https://bugs.launchpad.net/bugs/1948027

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1948027

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -819,6 +819,8 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
             g_key_file_set_string(kf, "ipv6", "addr-gen-mode", addr_gen_mode_str(def->ip6_addr_gen_mode));
         if (def->ip6_privacy)
             g_key_file_set_integer(kf, "ipv6", "ip6-privacy", 2);
+        else
+            g_key_file_set_integer(kf, "ipv6", "ip6-privacy", 0);
         if (def->gateway6)
             g_key_file_set_string(kf, "ipv6", "gateway", def->gateway6);
         if (def->ip6_nameservers) {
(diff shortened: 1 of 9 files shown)


slyon:
Thank you very much for the contribution @peleroux !
This change seems to interfere with some of the existing unit tests in `make check` (see https://github.com/canonical/netplan/runs/4180730044?check_suite_focus=true) Could you please check those failures and handle them accordingly?

FTR: Relevant documentation: https://developer-old.gnome.org/NetworkManager/stable/settings-ipv6.html

peleroux:
Thank you for your answer.
Last commit should fix errors in make check.

slyon (approved):
Thank you Pierre-Emmanuel for this contribution and bug fix! (Sorry for me taking so long to get around reviewing this)

The fix is perfectly valid and brings the NM renderer implementation in line with the networkd backend and what is suggested by our netplan reference documentation:

> Enable IPv6 Privacy Extensions (RFC 4941) for the specified interface, and
> prefer temporary addresses. **Defaults to false** - no privacy extensions. There
> is currently no way to have a private address but prefer the public address.

networkd's `IPv6PrivacyExtensions=` setting defaults to **false** and does not generate privacy addresses: https://www.freedesktop.org/software/systemd/man/systemd.network.html#IPv6PrivacyExtensions=

NM OTOH has been using the fallback mechanism to check the NM's global `ipv6.ip6-privacy` setting or fall back to the kernel's setting in `/proc/sys/net/ipv6/conf/default/use_tempaddr` (that would be networkd's "kernel" config value). => Therefore it would fall back to generate privacy IPs anyways, regardless of the `ipv6-privacy: false` setting.

So this is a change in default behaviour for the NM renderer backend, but at the same time fixes a bug to bring it in line with what is suggested by the documentation (and actually be able to disable privacy IP generation).

I'll merge this, but we should make sure to mention the change in behaviour in the release notes.

PS: I had a tiny remark about your adopted tests in `test_keyfile.py`, we should mark the `ip6-privacy=0` setting as a default value instead of passing it through the keyfile parser. I fixed that up myself and pushed to your branch.

PR #245, opened 2021-11-24
dbus: wait for 'netplan try' to be ready (LP: #1949893)

slyon:
## Description
netplan-dbus now waits for the spawned 'netplan try' child to touch the `/run/netplan/netplan-try.ready` stamp file before it returns the DBus call. If no stamp file is detected within up to 5 sec it hits a timeout and returns an error.

Additionally, this PR places ephemeral netplan configs in `/run/netplan/config-XXXXXX`, to avoid potential exploits of tmpfiles, like https://lwn.net/Articles/250468/

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1949893

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -26,6 +26,7 @@ import glob
 import subprocess
 import shutil
 import netifaces
+import time
 
 import netplan.cli.utils as utils
 from netplan.configmanager import ConfigManager, ConfigurationError
@@ -262,6 +263,15 @@ class NetplanApply(utils.NetplanCommand):
             for iface in utils.nm_interfaces(restart_nm_glob, devices):
                 utils.ip_addr_flush(iface)
             utils.systemctl_network_manager('start', sync=sync)
+            if sync:
+                # wait up to 2 sec for 'STATE=connected (site/local-only)' or
+                # 'STATE=connected' to appear in 'nmcli general' STATE
+                env = dict(os.environ, LC_ALL='C')
+                cmd = ['nmcli', 'general', 'status']
+                for _ in range(20):
+                    if b'\nconnected' in subprocess.check_output(cmd, env=env):
+                        break
+                    time.sleep(0.1)
 
     @staticmethod
     def is_composite_member(composites, phy):
--- a/netplan/cli/commands/try_command.py
+++ b/netplan/cli/commands/try_command.py
@@ -33,6 +33,7 @@ import netplan.terminal
 # Keep a timeout long enough to allow the network to converge, 60 seconds may
 # be slightly short given some complex configs, i.e. if STP must reconverge.
 DEFAULT_INPUT_TIMEOUT = 120
+NETPLAN_TRY_STAMP = os.path.join('/', 'tmp', 'netplan-try.ready')
 
 
 class NetplanTry(utils.NetplanCommand):
@@ -53,6 +54,15 @@ class NetplanTry(utils.NetplanCommand):
         if not self._config_manager:
             self._config_manager = ConfigManager()
         return self._config_manager
+
+    @staticmethod
+    def clear_ready_stamp():
+        if os.path.isfile(NETPLAN_TRY_STAMP):
+            os.remove(NETPLAN_TRY_STAMP)
+
+    @staticmethod
+    def touch_ready_stamp():
+        open(NETPLAN_TRY_STAMP, 'w').close()
 
     def run(self):  # pragma: nocover (requires user input)
         self.parser.add_argument('--config-file',
@@ -86,6 +96,9 @@ class NetplanTry(utils.NetplanCommand):
 
             NetplanApply().command_apply(run_generate=True, sync=True, exit_on_error=False, state_dir=self.state)
 
+            # Touch stamp file, it is the signal (for netplan-dbus) that we're
+            # ready to accept any Accept/Reject input (like SIGUSR1 or SIGTERM)
+            NetplanTry.touch_ready_stamp()
             self.t.get_confirmation_input(timeout=self.timeout)
         except netplan.terminal.InputRejected:
             print("\nReverting.")
@@ -100,6 +113,7 @@ class NetplanTry(utils.NetplanCommand):
             if self.t:
                 self.t.reset(self.t_settings)
             self.cleanup()
+            NetplanTry.clear_ready_stamp()
 
     def backup(self):  # pragma: nocover (requires user input)
         backup_config_dir = False
--- a/netplan/cli/commands/try_command.py
+++ b/netplan/cli/commands/try_command.py
@@ -33,7 +33,6 @@ import netplan.terminal
 # Keep a timeout long enough to allow the network to converge, 60 seconds may
 # be slightly short given some complex configs, i.e. if STP must reconverge.
 DEFAULT_INPUT_TIMEOUT = 120
-NETPLAN_TRY_STAMP = os.path.join('/', 'tmp', 'netplan-try.ready')
 
 
 class NetplanTry(utils.NetplanCommand):
@@ -48,6 +47,8 @@ class NetplanTry(utils.NetplanCommand):
         self._config_manager = None
         self.t_settings = None
         self.t = None
+        self._rootdir = os.environ.get('DBUS_TEST_NETPLAN_ROOT', '/')
+        self._netplan_try_stamp = os.path.join(self._rootdir, 'run', 'netplan', 'netplan-try.ready')
 
     @property
     def config_manager(self):  # pragma: nocover (called by later commands)
@@ -55,14 +56,15 @@ class NetplanTry(utils.NetplanCommand):
             self._config_manager = ConfigManager()
         return self._config_manager
 
-    @staticmethod
-    def clear_ready_stamp():
-        if os.path.isfile(NETPLAN_TRY_STAMP):
-            os.remove(NETPLAN_TRY_STAMP)
+    def clear_ready_stamp(self):
+        if os.path.isfile(self._netplan_try_stamp):
+            os.remove(self._netplan_try_stamp)
+            return True
+        return False
 
-    @staticmethod
-    def touch_ready_stamp():
-        open(NETPLAN_TRY_STAMP, 'w').close()
+    def touch_ready_stamp(self):
+        os.makedirs(self._rootdir + '/run/netplan', mode=0o700, exist_ok=True)
+        open(self._netplan_try_stamp, 'w').close()
 
     def run(self):  # pragma: nocover (requires user input)
         self.parser.add_argument('--config-file',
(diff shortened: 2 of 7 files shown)


src/dbus.c
```diff
@@ -506,7 +508,35 @@ method_try(sd_bus_message *m, void *userdata, sd_bus_error *ret_error)
...
+     * ready to accept any Accept/Reject signal from us (e.g. SIGTERM/SIGUSR1). */
+    char buf[512];
+    const char* buf_p = (char*)&buf;
+    struct pollfd fd = { .fd = child_stdout, .events = POLLIN };
+    guint poll_timeout = 5000;
+    if (seconds > 0 && seconds < 5)
+        poll_timeout = seconds * 1000;
+    /* Timeout after up to 5 sec of waiting on stdout */
+    while (poll(&fd, 1, poll_timeout) > 0) {
+        ssize_t cnt = read(child_stdout, &buf, 512);
```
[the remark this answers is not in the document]
mvo5: (nitpick²) this will not detect short reads and if that happens thing will fail with the timeout  - which I think are probably not a problem here but wanted to mention it anyway.

mvo5 (approved):
Looks good, two nitpicks inside but given the constraints probably the best we can do :)

mvo5:
So I couldn't quite let go of this and I'm probably being a bit silly here. But using stdout for the indication that netplan try is ready bugged me a bit, it feels a bit inelegant (and if it ever gets translated both sides need to look for the right translations and there are short reads etc). 

So I was wondering if a small tweak could be used instead. AIUI the `netplan try` from dbus always passes a `--state` dir which cold be used to simply create a file there when things are ready, something like: https://paste.ubuntu.com/p/j54KHGh2HT/ (needs more cleanup as it does not undo the pipe for stdout for example from the original patch and I probably did get other things wrong, did not really test this. If we don't want to poll we could use g_file_monitor or something but probably overkill).

I was also thinking about to just pass a dbus path from the dbus method and let the cli call into this but it seems a bit too much. Anyway, just a thought.

netplan/cli/commands/apply.py
```diff
@@ -262,6 +262,9 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
             for iface in utils.nm_interfaces(restart_nm_glob, devices):
                 utils.ip_addr_flush(iface)
             utils.systemctl_network_manager('start', sync=sync)
+            if sync:
+                # give it some time to actually start up and create any new interfaces
+                subprocess.call(['nm-online', '-q', '-t', '5'])
```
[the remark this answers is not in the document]
slyon: Let me double check that. Maybe there is a better way to wait for NM to be fully ready (and have all its virtual interfaces created). Apparently it is not ready directly after `systemctl start NetworkManager.service` finished.

slyon:
Thank you @mvo5 for your comments! Indeed having a "stamp file" is a much more universal approach. I'm not sure why I didn't think about that earlier... I thought about passing signals from child to parent process (like we do from parent to child currently) and also about passing DBus signal directly from the Python code (but that would probably be overkill as we do not have proper DBus integration with the python code, currently and on the `netplan-dbus` daemon side it would have led to additional sd-bus interrupts and probably the need for async methods.)

Polling for a file is plain and simple solution.
I adopted your recommendation a bit, primarily by having a global `/tmp/netplan-try.ready` stamp file. As there can only every be one single `netplan try` call at a time (as it is touching system system config and (re-)starting services). This way we do not have to shuffle it around with the states and it can also be used in a non-DBus environment.

netplan/cli/commands/apply.py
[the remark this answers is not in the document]
slyon: Okay, I guess I found a better way than `nm-online`. When starting up the NetworkManager STATE (as found in "nmcli general") goes through different state, when reaching a "connected" state (can be "(site-only)" or "(local-only)") it has brought up all the local (virtual) network interfaces, like bridges.
```
$ LC_ALL=C nmcli g
STATE      CONNECTIVITY  WIFI-HW  WIFI     WWAN-HW  WWAN     
connected (site-only)  full          enabled  enabled  enabled  disabled 
```

netplan/cli/commands/try_command.py
```diff
@@ -47,13 +47,24 @@ def __init__(self):
         self._config_manager = None
         self.t_settings = None
         self.t = None
+        self._rootdir = '/'
```
mvo5, line 50: (nitpick) you could write this slightly more compact as:
```
self._rootdir = os.environ.get('DBUS_TEST_NETPLAN_ROOT', '/')
```
but of course it's debatable if it's also more readable - up to you of course :)

netplan/cli/commands/try_command.py
```diff
@@ -47,13 +47,24 @@ def __init__(self):
...
+        self._netplan_try_stamp = os.path.join(self._rootdir, 'run', 'netplan-try.ready')
 
     @property
     def config_manager(self):  # pragma: nocover (called by later commands)
         if not self._config_manager:
             self._config_manager = ConfigManager()
         return self._config_manager
 
+    def clear_ready_stamp(self):
+        if os.path.isfile(self._netplan_try_stamp):
```
mvo5, line 62: (ntpick^3) this is a potential TOCTOU issue, probably utterly irrelevant here, it could be a problem in a multi-threaded environment when clear_ready_stamp is called in two threads - which is not really an issue, just noticed it while reviewing and couldn't help but mention it.

netplan/cli/commands/try_command.py
```diff
@@ -47,13 +47,24 @@ def __init__(self):
         self._config_manager = None
         self.t_settings = None
         self.t = None
+        self._rootdir = '/'
+        if 'DBUS_TEST_NETPLAN_ROOT' in os.environ:
+            self._rootdir = os.environ['DBUS_TEST_NETPLAN_ROOT']
+        self._netplan_try_stamp = os.path.join(self._rootdir, 'run', 'netplan-try.ready')
```
mvo5, line 53: This is very nice!

tests/dbus/test_dbus.py
```diff
@@ -49,6 +50,7 @@ def setUp(self):
     eth0:
       dhcp4: true""")
         self.addCleanup(shutil.rmtree, self.tmp)
+        self.addCleanup(shutil.rmtree, '/tmp/netplan-config-BACKUP', True)
```
mvo5, line 53: Is this using the real /tmp during the tests? Would it make sense to also move this under /run or would that break something ?

tests/dbus/test_dbus.py
```diff
@@ -424,6 +425,10 @@ def test_netplan_dbus_config_apply(self):
         self.assertFalse(os.path.isdir(tmpdir))
 
     def test_netplan_dbus_config_try_cancel(self):
+        # touch self._netplan_try_stamp to signal that 'netplan try' is ready
+        # to take (Accept/Reject) input signals, before the timeout
+        self.addCleanup(os.remove, self._netplan_try_stamp)
```
mvo5, line 430: (nitpick) I see this cleanup in a bunch of places and wonder if it would make sense to move it into `setUpTest` (or if half the tests need it and half don't maybe split TestNetplanDBus into two where one does the cleanup in central place and one does not).

tests/test_cli_units.py
```diff
@@ -79,3 +91,12 @@ def test_clear_virtual_links_no_devices(self, mock):
...
             self.assertEqual(ctx.output, ['WARNING:root:Cannot clear virtual links: no network interfaces provided.'])
         mock.assert_not_called()
+
+    def test_netplan_try_ready_stamp(self):
+        stamp_file = os.path.join(self.tmproot, 'run', 'netplan-try.ready')
+        cmd = NetplanTry()
+        self.assertFalse(os.path.isfile(stamp_file))
+        cmd.touch_ready_stamp()
+        self.assertTrue(os.path.isfile(stamp_file))
+        cmd.clear_ready_stamp()
```
mvo5, line 101: (nitpick) maybe a test that runs clear_ready_stamp() again when the file does not exist to check that it behaves correctly in this case (i.e. check the expected error or silently ignore error, whatever is the right thing here).

mvo5 (approved):
Looks very nice! I have some comments/nitpicks/suggestions inline that are hopefully useful but nothing that would bock this. Thanks a lot for working on this!

tests/dbus/test_dbus.py
slyon: Moving everything into `<TEST-TMP>/run/netplan/*` which is cleaned up after every test-case, this is not needed anymore.

tests/dbus/test_dbus.py
slyon: Yes. That's ugly. Let's move everything to `<TEST-ROOT>/run/netplan/config-*`, that way test can also be run in parallel.

netplan/cli/commands/try_command.py
slyon: `netplan try` is not designed to run two instances or threads in parallel, if we'd want to enable this, we'd probably need a much bigger refactoring. So I'll leave this as is for now.

netplan/cli/commands/try_command.py
```diff
@@ -47,13 +47,26 @@ def __init__(self):
...
         return self._config_manager
 
+    def clear_ready_stamp(self):
+        if os.path.isfile(self._netplan_try_stamp):
+            os.remove(self._netplan_try_stamp)
+            return True
+        return False
+
+    def touch_ready_stamp(self):
+        os.makedirs(self._rootdir + '/run/netplan', mode=0o700, exist_ok=True)
```
mvo5, line 66: Nice

netplan/cli/commands/try_command.py
```diff
@@ -47,13 +47,26 @@ def __init__(self):
...
+        open(self._netplan_try_stamp, 'w').close()
+        return True
```
mvo5, line 68: (nitpick) Given that this never returns False I wonder if we need a return value here ? I see it's used in the tests but I think it's okay to just not have a return value here, the side-effect of touching a file is something that can be observed.

src/dbus.c
```diff
@@ -478,13 +480,16 @@ method_try(sd_bus_message *m, void *userdata, sd_bus_error *ret_error)
...
     if (d->config_id)
-        state = g_strdup_printf("--state=%s/netplan-config-%s", g_get_tmp_dir(), NETPLAN_GLOBAL_CONFIG);
+        state = g_strdup_printf("--state=%s/run/netplan/config-%s", NETPLAN_ROOT, NETPLAN_GLOBAL_CONFIG);
     gchar *argv[] = {SBINDIR "/" "netplan", "try", timeout, state, NULL};
 
     // for tests only: allow changing what netplan to run
     if (getenv("DBUS_TEST_NETPLAN_CMD") != 0)
        argv[0] = getenv("DBUS_TEST_NETPLAN_CMD");
 
+    /* Delete any left-over netplan-try.ready stamp file, if it exists */
```
mvo5, line 490: Nice, thanks for thinking about this!

mvo5 (approved):
Looks very nice, thank you! One tiny nitpick inline but feel free to ignore :)

netplan/cli/commands/try_command.py
slyon: ACK.

slyon:
Thank you so much @mvo5 for all your input here. I think this is ready for merging \o/

Let me run some more test on it before merging it early next week.

slyon:
Merged in e19441c32a35cc58770753dfa8e6783438db295b

PR #247, opened 2021-12-03
keyfile: do not try to write out unvalidated YAML (LP: #1952967)

slyon:
## Description
netplan's keyfile parser can handle certain device types and settings from NM's keyfile format. But we should try to avoid writing out partial YAML where some (essential) fields are missing, as they cannot (yet) be properly parsed.

We need to check that during testing and ideally also at runtime, in order to fall back to the full passthrough mode, using the generic `nm-devices` type.

FR-1898

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1952967

### DIFF
--- a/tests/parser/base.py
+++ b/tests/parser/base.py
@@ -19,6 +19,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from configparser import ConfigParser
+from netplan.cli.utils import _GError
 import os
 import sys
 import shutil
@@ -73,6 +74,7 @@ class TestKeyfileBase(unittest.TestCase):
 
     def generate_from_keyfile(self, keyfile, netdef_id=None, expect_fail=False, filename=None):
         '''Call libnetplan with given keyfile string as configuration'''
+        err = ctypes.POINTER(_GError)()
         # Autodetect default 'NM-<UUID>' netdef-id
         ssid = ''
         if not netdef_id:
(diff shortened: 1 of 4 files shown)


tests/parser/base.py
```diff
@@ -88,21 +90,33 @@ def generate_from_keyfile(self, keyfile, netdef_id=None, expect_fail=False, file
...
         with capture_stderr() as outf:
             if expect_fail:
-                self.assertFalse(lib.netplan_parse_keyfile(f.encode(), None))
+                self.assertFalse(lib.netplan_parse_keyfile(f.encode(), ctypes.byref(err)))
+                if err:
+                    return err.contents.message.decode('utf-8')
             else:
-                self.assertTrue(lib.netplan_parse_keyfile(f.encode(), None))
+                self.assertTrue(lib.netplan_parse_keyfile(f.encode(), ctypes.byref(err)))
+                if err:  # pragma: nocover (this case should never happen)
```
schopin-pro, line 109: suggestion (non-blocking): the comment isn't exactly true, this codepath should actually happen, but isn't relevant for code coverage

```suggestion
                if err:  # pragma: nocover (only happens if a test fails so irrelevant for coverage)
```

tests/parser/base.py
```diff
@@ -150,7 +164,9 @@ def assert_nm_regenerate(self, file_contents_map):
         argv = [exe_generate, '--root-dir', self.workdir.name]
         p = subprocess.Popen(argv, stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE, universal_newlines=True)
+        returncode = p.wait(5)
```
schopin-pro, line 167: praise: Good catch!

This seem like the sort of thing that could cause serious headaches!

schopin-pro (approved):
LGTM, although I wouldn't have expected this many different changes from the PR description. But then, who am I to talk :wink:

There's only a small nitpick, but you can disregard it and merge if you feel like it.

PR #248, opened 2021-12-16
Improve routing capabilities (LP: #1892272) (LP: #1805038)

slyon:
## Description
* Supporting additional route types: anycast, broadcast, local, multicast, nat, throw, xresolve (supported by systemd-networkd as of v243)
* Not assuming a type: unicast, scope: global, by default, but making the default scope depend on the route type:
  * local -> host
  * nat -> host
  * broadcast -> link
  * multicast -> link
  * anycast -> link
  * unicast without gateway/via -> link
  * unicast with gateway/via -> global
  * global otherwise

This only affects the systemd-networkd backend renderer as for NetworkManager only "unicast" routes are supported.

FR-1798
FR-1078
FR-1077

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1892272 & LP#1805038

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1607,7 +1607,7 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, const void* _, GError** err
         g_assert(npp->current.route == NULL);
         route = g_new0(NetplanIPRoute, 1);
         route->type = g_strdup("unicast");
-        route->scope = g_strdup("global");
+        route->scope = NULL;
         route->family = G_MAXUINT; /* 0 is a valid family ID */
         route->metric = NETPLAN_METRIC_UNSPEC; /* 0 is a valid metric */
         route->table = NETPLAN_ROUTE_TABLE_UNSPEC;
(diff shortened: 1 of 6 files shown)


src/parse.c
```diff
@@ -1608,6 +1618,21 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, const void* _, GError** err
         if (!process_mapping(npp, entry, routes_handlers, NULL, error))
             goto err;
 
+        /* Set the default scope, according to type */
+        if (!route->scope) {
+            if (   g_ascii_strcasecmp(route->type, "local") == 0
+                || g_ascii_strcasecmp(route->type, "nat") == 0)
+                route->scope = (g_strdup("host"));
+            /* Non-gatewayed unicast routes are scope:link, too */
```
slyon, line 1626: Should we rather print a warning, suggesting to add `scope: link` in this case, instead of doing it implicitly?

schopin-pro: I'd rather not. The default behavior seems logical to me and matches what `ip route` does by default. I'd rather have `netplan get` grow the capacity to display the full configuration (which should be easy after #252), which could help whoever forgot their gateway.

schopin-pro (approved):
LGTM, merging.

schopin-pro:
Well, I'll let you rebase and merge is more like it, but still, nice work :)

src/parse.c
```diff
@@ -1597,7 +1607,7 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, const void* _, GError** err
         g_assert(npp->current.route == NULL);
         route = g_new0(NetplanIPRoute, 1);
         route->type = g_strdup("unicast");
-        route->scope = g_strdup("global");
+        route->scope = NULL;
```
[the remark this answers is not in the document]
slyon: Good idea, I've added a section to `TODO`, adding a few things to do when bumping the soname, so we can look it up once we're ready.

PR #249, opened 2022-01-05
YAML consolidation prelude: OVS handling (FR-702)

schopin-pro:
## Description

A couple of patches related to OpenVSwitch handling. The consistency patch is needed to be able to ensure idempotency of the user input (within reason). The type name definition will be needed when we'll rewrite `sriov.py` as we'll need to identify those definitions.

## Checklist

### DIFF
--- a/src/names.c
+++ b/src/names.c
@@ -48,7 +48,7 @@ netplan_def_type_to_str[NETPLAN_DEF_TYPE_MAX_] = {
     [NETPLAN_DEF_TYPE_BOND] = "bonds",
     [NETPLAN_DEF_TYPE_VLAN] = "vlans",
     [NETPLAN_DEF_TYPE_TUNNEL] = "tunnels",
-    [NETPLAN_DEF_TYPE_PORT] = NULL,
+    [NETPLAN_DEF_TYPE_PORT] = "ovs-ports",
     [NETPLAN_DEF_TYPE_NM] = "nm-devices",
 };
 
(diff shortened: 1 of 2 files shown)


src/names.c
```diff
@@ -48,7 +48,7 @@ netplan_def_type_to_str[NETPLAN_DEF_TYPE_MAX_] = {
     [NETPLAN_DEF_TYPE_BOND] = "bonds",
     [NETPLAN_DEF_TYPE_VLAN] = "vlans",
     [NETPLAN_DEF_TYPE_TUNNEL] = "tunnels",
-    [NETPLAN_DEF_TYPE_PORT] = NULL,
+    [NETPLAN_DEF_TYPE_PORT] = "ovs-ports",
```
slyon, line 51: This is not about the backends/renderers (as described in the commit msg), but about the interface type definition, right?
We do not expose `ovs-ports` to the YAML schema, but this is kind of a special type to carry the relevant data around.

Would you mind changing that name a bit to reflect that this is something special?

```suggestion
    [NETPLAN_DEF_TYPE_PORT] = "_ovs-ports",
```

slyon (approved):
LGTM! It makes sense to handle the definitions in an ordered way. If you don't mind I'd like the `ovs-ports` name, though, as this is kind of an edge case and should be visible as such, IMO.

Thanks!

src/names.c
schopin-pro: I'll apply the patch, as, even though I'm not very fond of this new syntax, your reasons for adopting it are better than my pure "it looks ugly" approach ;-)

PR #251, opened 2022-01-05
YAML consolidation prelude: new YAML dump APIs (FR-702)

schopin-pro:
## Description

Introduce a couple of APIs that are needed for the various YAML consolidation efforts. Note that this is technically an API break, but the function that is removed has never been part of a release, so this *should* be OK.

## Checklist

### DIFF
--- a/src/netplan.c
+++ b/src/netplan.c
@@ -16,11 +16,17 @@
  */
 
 #include <glib.h>
+#include <glib/gstdio.h>
 #include <yaml.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
 
 #include "netplan.h"
 #include "parse.h"
+#include "types.h"
 #include "yaml-helpers.h"
+#include "util-internal.h"
 #include "names.h"
 
 gchar *tmp = NULL;
@@ -492,21 +498,6 @@ err_path: return FALSE; // LCOV_EXCL_LINE
 }
 
 static gboolean
-has_openvswitch(const NetplanOVSSettings* ovs, NetplanBackend backend, GHashTable *ovs_ports) {
-    return (ovs_ports && g_hash_table_size(ovs_ports) > 0)
-            || (ovs->external_ids && g_hash_table_size(ovs->external_ids) > 0)
-            || (ovs->other_config && g_hash_table_size(ovs->other_config) > 0)
-            || ovs->lacp
-            || ovs->fail_mode
-            || ovs->mcast_snooping
-            || ovs->rstp
-            || ovs->protocols
-            || (ovs->ssl.ca_certificate || ovs->ssl.client_certificate || ovs->ssl.client_key)
-            || (ovs->controller.connection_mode || ovs->controller.addresses)
-            || backend == NETPLAN_BACKEND_OVS;
-}
-
-static gboolean
 write_openvswitch(yaml_event_t* event, yaml_emitter_t* emitter, const NetplanOVSSettings* ovs, NetplanBackend backend, GHashTable *ovs_ports)
 {
     GHashTableIter iter;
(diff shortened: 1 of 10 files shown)


schopin-pro:
Draft as it depends on #250

src/util.c
```diff
@@ -119,6 +122,157 @@ int find_yaml_glob(const char* rootdir, glob_t* out_glob)
...
+            g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT, "Error emitting YAML: %s", emitter->problem);
+            return FALSE;
+            // LCOV_EXCL_STOP
+        }
+    } while (map_count || seq_count);
+    return TRUE;
+}
+
+static gboolean
+drill_into_yaml_subtree(yaml_parser_t *parser, yaml_emitter_t *emitter, char** key, GError** error) {
```
slyon, line 162: This function took me a while to grasp. It all makes sense, but I'd like to see a comment above briefly describing what this is supposed to do. Also stating that `char**keys` is not some kind of string to be returned to the caller (as is usual with this prototype), but rather a vector of the split up prefix.

tests/generator/base.py
```diff
@@ -277,8 +277,6 @@ def validate_generated_yaml(self, yaml_input):
             if os.path.isfile(output_path):
                 with open(output_path, 'r') as generated:
                     output_yaml = yaml.safe_load(generated.read())
-            else:
-                output_yaml = yaml.safe_load('')
```
slyon, line 281: This change seems kind of unrelated. Could you please clarify why it is part of this commit?

slyon (commented):
Thank you Simon! This is another great step in consolidating our YAML parsing story.

It is looking very good overall. I asked for some clarification and mentioned a few nitpicks in the inline comments. We should be good to merge this after those are resolved!

Also, I can confirm that the `netplan_state_write_yaml()` function isn't used anywhere and was never released. So we're good to "break" that API.

src/netplan.c
```diff
@@ -973,16 +964,51 @@ netplan_state_write_yaml(const NetplanState* np_state, const char* file_hint, co
...
+    }
+
+
+    while (iter) {
+        NetplanNetDefinition* netdef = iter->data;
+        to_write = g_list_append(to_write, netdef);
+        iter = iter->next;
+    }
+
+    return netplan_netdef_list_write_yaml(np_state, to_write, out_fd, error);
```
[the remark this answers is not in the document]
schopin-pro: Nice catch! I'm fairly sure that if this function leaks its list we'll have many other later on.

I'm wondering if we can't use a decorator to automate g_list_free when we lose the scope, though.

tests/generator/base.py
schopin-pro: I had to dig into my memory to remember why I changed this one, and I must have been tired when I decided that it would be the good solution: I removed those lines because they weren't hit by code coverage anymore. I didn't bother wondering why, I just removed them.

Now that I've actually thought about it, it means there's a subtle regression: `write_netplan_conf_full()` will now create an empty file if the global state is empty, whereas the previous version wouldn't touch the FS.

There are multiple options there:
1/ we simply don't care
2/ we approximate the previous behavior by removing the created file if it's empty
3/ we keep the exact behavior as before by creating a new function
```c
gboolean
netplan_state_nondefault_globals(NetplanState *np_state)
{
    return (np_state->backend != NETPLAN_BACKEND_NONE)
        || has_openvswitch(&np_state->ovs_settings, NETPLAN_BACKEND_NONE, NULL);
}
```
and using it in `write_netplan_conf_full` to do an early check 

```c
if (!netplan_state_nondefault_globals(&global_state) || !netplan_state_get_netdefs_size(&global_state))
    return;
```

Now that I've done the work, the latter approach is nicer :)

src/netplan.c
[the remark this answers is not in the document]
schopin-pro: Turns out we don't even need a new list here. Should be fixed in the new patchset.

tests/generator/base.py
schopin-pro: Addressed in the latest patchset.

src/util.c
schopin-pro: I added some comments, renamed the function to emit_yaml_subtree and the argument to yaml_path.

slyon: Thank you, I like that much better!

tests/generator/base.py
slyon: Empty files can cause issues (e.g. with snapd get/set, as addressed with the most recent netplan.io SRUs in Ubuntu), so thank you for going with option (3) and keeping status quo!

slyon (approved):
Thank you for addressing the comments, especially the `netplan_state_nondefault_globals()` function should avoid issues with our snapd integration so good idea going with option (3) there!

LGTM. Let's merge this.

PR #252, opened 2022-01-05
YAML consolidation: `netplan get` (FR-702)

schopin-pro:
## Description

Port the `netplan get` command to use the internal YAML generator, originally written for the NM-to-netplan system.
In order to make this work, we had to add a notion of "dirtyness" to the input data, as the user would want their explicit settings displayed even if they match the default value.

## Checklist

### DIFF
--- a/netplan/cli/commands/get.py
+++ b/netplan/cli/commands/get.py
@@ -17,15 +17,16 @@
 
 '''netplan get command line'''
 
-import yaml
+import sys
+import io
+import tempfile
 import re
 
 import netplan.cli.utils as utils
-from netplan.configmanager import ConfigManager
+import netplan.libnetplan as libnetplan
 
 
 class NetplanGet(utils.NetplanCommand):
-
     def __init__(self):
         super().__init__(command_id='get',
                          description='Get a setting by specifying a nested key like "ethernets.eth0.addresses", or "all"',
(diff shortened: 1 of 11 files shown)


schopin-pro:
Draft as it depends on #250 and #251 , and would also benefit from #249 being merged

tests/generator/base.py
```diff
@@ -274,6 +274,7 @@ def validate_generated_yaml(self, yaml_input):
                     input_yaml = yaml.safe_load('')
 
             # Read output of the YAML generator (if any)
+            output_yaml = None
```
slyon, line 277: This seems unrelated. It will be overridden in either the `if` or the `else` statement. It is best practice to have a variable clearly defined on top before we do any logic on it, so we can leave it in there.
But just to clarify: This has no other purpose than better readability/clarity of the code, right?

netplan/cli/commands/get.py
```diff
@@ -41,27 +42,37 @@ def run(self):
         self.parse_args()
         self.run_command()
 
+    def dump_state(self, key, np_state, output_file):
+        if key == 'all':
+            np_state.dump_yaml(output_file=output_file)
+            return
+
+        if not key.startswith('network'):
```
slyon, line 50: Hmm... I remember having a reason for doing that in the specific way it was done before; I can remember the details, though... Would you mind keeping it at this, to not change behaviour?
```suggestion
        if not key.startswith('network.') and not key == 'network':
```

I think one problem could be that if somebody tries to `netplan get networkINVALID`, this should return an error/empty tree. OTOH.. I think that might actually be handled by the `dump_yaml_subtree` function, as it won't find the `networkINVALID` prefix.

slyon (commented):
Thank you for another great PR, Simon!
The dirty tracking functionality is really nice and should help us cleaning up some parts of the test environment, too, where we're currently doing some "normalization" of default values.

I left some inline comments. But again, nothing critical.

tests/generator/base.py
schopin-pro: Oh, no, even better than that: it has no purpose. It's an accidental leftover of the previous version of #251 with its unfortunate side effect on empty files. I'll get rid of it.

netplan/cli/commands/get.py
schopin-pro: One reason would be the handling of documents with more than one top-level key, but since we go through the entire parsing process, those would get tossed out anyway.

Unless you give me an actual case where the new code is problematic, I'd rather keep it as is, as I've noticed that weird code with unclear rationales tend to calcify.

I've added a test case for the `networkINVALID` query, though :)

src/parse.c
```diff
@@ -2228,128 +2276,128 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
...
+    {"ipv6-mtu", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(ipv6_mtubytes)},  \
+    {"ipv6-privacy", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(ip6_privacy)}, \
+    {"link-local", YAML_SEQUENCE_NODE, {.generic=handle_link_local}},                                    \
+    {"macaddress", YAML_SCALAR_NODE, {.generic=handle_netdef_mac}, netdef_offset(set_mac)},        \
+    {"mtu", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(mtubytes)},            \
+    {"nameservers", YAML_MAPPING_NODE, {.map={FALSE, {.handlers=nameservers_handlers}}}},                           \
+    {"optional", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(optional)},        \
+    {"optional-addresses", YAML_SEQUENCE_NODE, {.generic=handle_optional_addresses}},                    \
+    {"renderer", YAML_SCALAR_NODE, {.generic=handle_netdef_renderer}},                                   \
+    {"routes", YAML_SEQUENCE_NODE, {.generic=handle_routes}},                                            \
```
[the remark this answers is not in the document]
schopin-pro: I ended up removing them here, as the fields are way too varying for it to render cleanly, but I've kept them in the other instance, since all fields have the same type, thus same approximate length.

src/parse.c
```diff
@@ -187,16 +187,25 @@ netplan_netdef_new(NetplanParser *npp, const char* id, NetplanDefType type, Netp
...
-    /* handler for the value of this key */
-    node_handler handler;
-    /* if type == YAML_MAPPING_NODE and handler is NULL, use process_mapping()
-     * on this handler map as handler */
-    const struct mapping_entry_handler_s* map_handlers;
+    union {
+        node_handler generic;
+        custom_map_handler variable;
+        struct {
+            gboolean custom_handling;
```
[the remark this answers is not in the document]
schopin-pro: Turns out it could be implied :)

schopin-pro:
Tests run on my local LXD, as GH ACtions seems to be having troubles today:

```
autopkgtest [11:31:23]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              PASS
vlans                PASS
wifi                 PASS
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
```

netplan/cli/commands/get.py
slyon: That's fair, I guess! All our tests are passing. Should we hit a regression, we'll fix it up and add a new test case for it.

slyon (approved):
Thank you for the V2 improvements and for running the integration tests locally!

LGTM :eyes:

PR #253, opened 2022-01-10
Enable eswitch mode setting on SmartNICs

slyon:
## Description
Certain types of network cards/SmartNICs allow toggling their low-level (`devlink`) embedded switch mode PCI setting in order to utilize the full hardware offloading capabilities. With this PR we're exposing this functionality to the netplan YAML schema:
* `embedded-switch-mode: [switchdev|legacy]`
  * choose the devlink setting for the embedded switch (eswitch)
  * sticks to the hardware's preset by default
* `delay-virtual-functions-rebind: [false|true]`
  * False by default: SR-IOV virtual functions are re-bind to their driver right after setting the eswitch mode
  * Can be delayed to a later stage during the boot process; useful when OVS/OVN bonding (VF LAG) is to be used

Example:
```
network:
  version: 2
  ethernets:
    engreen:
      embedded-switch-mode: "switchdev"
      delay-virtual-functions-rebind: true
    enblue:
      match:
        driver: "mlx5_core"
      set-name: enblue
      embedded-switch-mode: "legacy"
      virtual-function-count: 4
    sriov_vf0:
      link: engreen
```

This PR builds upon the new libnetplan YAML parser (https://github.com/canonical/netplan/pull/250) for accessing the new settings.

## Checklist

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -814,6 +814,23 @@ Example:
 
      **Requires feature: sriov**
 
+``embedded-switch-mode`` (scalar) – since **0.104**
+
+:    (SR-IOV devices only) Change the operational mode of the embedded switch
+     of a supported SmartNIC PCI device (e.g. Mellanox ConnectX-5). Possible
+     values are ``switchdev`` or ``legacy``, if unspecified the vendor's
+     default configuration is used.
+
+     **Requires feature: eswitch-mode**
+
+``delay-virtual-functions-rebind`` (bool) – since **0.104**
+
+:    (SR-IOV devices only) Delay rebinding of SR-IOV virtual functions to its
+     driver after changing the embedded-switch-mode setting to a later stage.
+     Can be enabled when bonding/VF LAG is in use. Defaults to ``false``.
+
+     **Requires feature: eswitch-mode**
+
 ## Properties for device type ``modems:``
 GSM/CDMA modem configuration is only supported for the ``NetworkManager``
 backend. ``systemd-networkd`` does not support modems.
--- a/netplan/cli/sriov.py
+++ b/netplan/cli/sriov.py
@@ -502,10 +502,6 @@ def apply_sriov_config(config_manager, rootdir='/'):
             # this only works for SR-IOV VF interfaces
             link = settings.get('link')
             vlan_id = settings.get('id')
-            if not vlan_id:
-                raise ConfigurationError(
-                    'no id property defined for SR-IOV vlan %s' % vlan)
-
             vf = vfs.get(link)
             if not vf:
                 # it is possible this is not an error, for instance when
--- a/netplan/cli/sriov.py
+++ b/netplan/cli/sriov.py
@@ -1,7 +1,8 @@
 #!/usr/bin/python3
 #
-# Copyright (C) 2020 Canonical, Ltd.
+# Copyright (C) 2020-2022 Canonical, Ltd.
 # Author: Łukasz 'sil2100' Zemczak <lukasz.zemczak@canonical.com>
+# Author: Lukas Märdian <slyon@ubuntu.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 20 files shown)


netplan/cli/sriov.py
```diff
@@ -27,6 +28,148 @@
 import netifaces
 
 
+# Taken from mlnx_switchdev_mode/sriovify.py
+class PCIDevice(object):
```
slyon, line 33: This uses an Apache-2.0 license and can be combined with our GPL3+ code

doc/netplan.md
```diff
@@ -808,6 +808,24 @@ Example:
 
      **Requires feature: sriov**
 
+``eswitch-mode`` (scalar) – since **0.104**
+
```
vorlonofportland, line 812: 'eswitch' seems jargony.  Would `embedded-switch-mode` be better?

doc/netplan.md
```diff
@@ -808,6 +808,24 @@ Example:
...
+:    (SR-IOV devices only) Change the operational mode of the embedded switch
+     of a supported SmartNIC PCI device (e.g. Mellanox ConnectX-5). Possible
+     values are ``switchdev`` or ``legacy``, if unspecified the vendor's
+     default configuration is used.
+
+     **Requires feature: eswitch-mode**
+
+``rebind-virtual-functions-delayed`` (bool) – since **0.104**
```
vorlonofportland, line 820: Awkward grammar in this key name.  Even if this is matching a string on the mellanox side, would recommend `delay-virtual-functions-rebind` or `delay-rebinding-virtual-functions` instead.

doc/netplan.md
slyon: Thanks for your comments! Yes, I had thought about that variation as well and will change it accordingly.

slyon:
We got some positive feedback from Canonical's OpenStack team about the new functionality, especially that the VF LAG (bonding of SR-IOV physical functions) is working as expected.

netplan/cli/sriov.py
```diff
@@ -500,12 +500,12 @@ def apply_sriov_config(config_manager, rootdir='/'):
...
             link = settings.get('link')
             vlan_id = settings.get('id')
-            if not vlan_id:
+            if not vlan_id:  # pragma: nocover (this is being reworked in PR#255)
                 raise ConfigurationError(
                     'no id property defined for SR-IOV vlan %s' % vlan)
 
             vf = vfs.get(link)
-            if not vf:
+            if not vf:  # pragma: nocover (this is being reworked in PR#255)
```
schopin-pro, line 508: question (non-blocking): given that #254 and #255 are pushed back for later, shouldn't we add tests for those?

This is totally your call, I just want to raise this point since the commit predates the decision.

src/sriov.h
```diff
@@ -19,7 +19,7 @@
 #include "netplan.h"
 
 NETPLAN_INTERNAL void
-write_sriov_conf_finish(const char* rootdir);
+write_sriov_conf_finish(const char* rootdir, GHashTable* sriov_pfs);
```
schopin-pro, line 22: question (blocking, abi): breaking the ABI here used by a binary in `netplan.io`, is this deliberate?

I vaguely recall doing some work in this area, but I don't remember the specifics. More to the point, I don't recall if I tried to maintain ABI with the generate binary or not.

src/generate.c
```diff
@@ -339,5 +345,7 @@ int main(int argc, char** argv)
         netplan_parser_clear(&npp);
     if (np_state)
         netplan_state_clear(&np_state);
+    if (sriov_pfs)
+        g_hash_table_destroy(sriov_pfs);
```
schopin-pro, line 349: nitpick: the hashtable should be cleared before the state.

There's a brief window where we have a hashtable full of dangling pointers. This isn't a big deal here but is easily avoidable.

tests/generator/base.py
```diff
@@ -456,3 +456,29 @@ def assert_ovs(self, file_contents_map):
...
+            # in this case we assume no SR-IOV configuration should be present
+            self.assertFalse(glob.glob(os.path.join(systemd_dir, '*netplan-sriov-*.service')))
+            return
+
+        self.assertEqual(set(os.listdir(self.workdir.name)) - {'lib'}, {'etc', 'run'})
+        sriov_systemd_dir = set(os.listdir(systemd_dir))
+        sriov_systemd_dir.remove('systemd-networkd.service.wants')
+        sriov_systemd_dir.remove('multi-user.target.wants')
+        sriov_systemd_dir.remove('netplan-ovs-cleanup.service')
+        self.assertEqual(sriov_systemd_dir, {'netplan-sriov-' + f for f in file_contents_map})
```
schopin-pro, line 472: suggestion (non-blocking): use the glob to produce the service file

(absolutely untested)
This should take care of the empty map for free, and voids the need to list all the extra service files.

```python
    def assert_sriov(self, file_contents_map):
        systemd_dir = os.path.join(self.workdir.name, 'run', 'systemd', 'system')
        sriov_systemd_dir = glob.glob(os.path.join(systemd_dir, '*netplan-sriov-*.service'))
        self.assertEqual(sriov_systemd_dir, {'netplan-sriov-' + f for f in file_contents_map})
        self.assertEqual(set(os.listdir(self.workdir.name)) - {'lib'}, {'etc', 'run'})
```

netplan/cli/sriov.py
```diff
@@ -230,11 +400,17 @@ def apply_vlan_filter_for_vf(pf, vf, vlan_name, vlan_id, prefix='/'):
...
     """
     Go through all interfaces, identify which ones are SR-IOV VFs, create
     them and perform all other necessary setup.
     """
+    parser = libnetplan.Parser()
+    parser.load_yaml_hierarchy(rootdir)
+
+    np_state = libnetplan.State()
+    np_state.import_parser_results(parser)
+
```
schopin-pro, line 413: thought: potential regressions here (same as #255, really) with newer, stricter checks

schopin-pro (commented):
This PR is in great shape, and the code looks overall solid :slightly_smiling_face: 

I have a few items, most of them non-blocking, but there's one in particular regarding potential ABI break that needs answering before merging.

tests/test_sriov.py
```diff
@@ -68,41 +68,76 @@ def setUp(self):
...
+            os.makedirs(vf_iface_path)
+            # symlink it to /sys/bus/pci/devices
+            os.symlink(os.path.join('../../../devices/pci0000:00', vf_pci_addr),
+                       os.path.join(pci_devices, vf_pci_addr))
+
+            # populate the VF data
+            with open(os.path.join(vf_dev_path, 'vendor'), 'w') as f:
+                f.write('0x001f\n')
+            with open(os.path.join(vf_dev_path, 'device'), 'w') as f:
+                f.write('0xb33f\n')
```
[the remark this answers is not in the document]
slyon: yes.

netplan/cli/sriov.py
slyon: ACK. That was done conscientiously. I had all of the new SR-IOV functionality implemented in a way not using libnetplan parser, but felt it would be better if new functionality would be using the new parser (so no rework is needed later on). A potential regression is limited to the SR-IOV usecase and I could relatively easily change that to using just native python `configmanager.py` in case we hit any issues.

We're also exposing the same potential regression for the new `netplan get` usecase/CLI, which could be mitigated by reverting 331ca01908b68446a9a1f395616136678e617a1a in case of any issue.

But I actually like the idea of enabling the new parser in _some_ selected usecases, to give it some limited exposure before enabling it throughout the stack.

src/netplan.c
```diff
@@ -707,6 +707,9 @@ _serialize_yaml(
     if (def->sriov_link)
         YAML_STRING(def, event, emitter, "link", def->sriov_link->id);
     YAML_UINT_DEFAULT(def, event, emitter, "virtual-function-count", def->sriov_explicit_vf_count, G_MAXUINT);
+    YAML_STRING(def, event, emitter, "embedded-switch-mode", def->embedded_switch_mode);
+    YAML_BOOL_TRUE(def, event, emitter, "delay-virtual-functions-rebind",
```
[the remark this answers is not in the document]
slyon: ACK.

tests/generator/base.py
slyon: Nice. Applied with small modifications.

netplan/cli/sriov.py
slyon: Yes, good point.
* For the first case (`if not vlan_id:`) we do not actually need a test, as this is now just handled a bit earlier by libnetplan, using a slightly different error message; `test_apply_sriov_config_invalid_vlan` is adopted to account for that.
* For the 2nd case (`if not vf:`) I've added a new test case.

src/sriov.h
slyon: Ohh, good spot! I did not do this deliberately, it slipped my mind! We do not want to break the ABI that is used from the `generate` binary (we did that last time, and it was bad).

I've now introduced new API `netplan_state_finish_sriov_write()` and `netplan_sriov_cleanup()` to handle this and moved the old functions to `abi_compat.c`. That should work.

Could you please double check my V2 of commit 9c99162816b6f255173b0adaa906ff9af02b3654 ?

slyon:
Integration tests are still good, even if the CI is doing funny things by skipping most (all?) tests, due to some package dependency issue. I re-ran it locally:
```
autopkgtest [16:07:51]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              PASS
vlans                PASS
wifi                 PASS
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
```

netplan/cli/sriov.py
schopin-pro: ACK, I'm just happy I'm not the one introducing potential regressions for once :smile:

schopin-pro (approved):
LGTM!
I have a small suggestion for the new API, but feel free to merge if you don't want to pick it up.

slyon:
Thanks for the re-review! Turned out the `write_sriov_conf_finish()` and `cleanup_sriov_conf()` functions have been part of the `generate` binary itself, so it didn't break ABI in the first place. But still it's nicer to have all this handling of SR-IOV logic in the sriov module and call into the library in future version.

I will be merging this.

dshcherb:
Thanks a lot for this, this is a much needed feature to utilize SR-IOV 👍

In case setting num_vfs needs to be done via systemd in the future, there was some recent work on that too:

https://github.com/systemd/systemd/pull/21865
https://github.com/systemd/systemd/commit/41ce9d769d394d19f1caab8b8a89b7dea50db0bc

slyon:
> In case setting num_vfs needs to be done via systemd in the future, there was some recent work on that too:

Nice! That's good to know. We should probably switch to that upstream solution in the future, once it lands in stable series (maybe in parallel to our manual way of doing it).

PR #254, opened 2022-01-11
YAML consolidation: `netplan set` (FR-702)

schopin-pro:
## Description

As part of the larger PyYAML cleanup effort, this patch set rewrites the `netplan set` to use libnetplan's own YAML parser instead of PyYAML. This meant enriching the parser to deal with null fields, that need to be tracked out-of-band, as well as creating new APIs to create YAML files out of a Netplan state, with different nuances to match the previous behavior.

## Checklist

### DIFF
--- a/netplan/cli/commands/set.py
+++ b/netplan/cli/commands/set.py
@@ -17,19 +17,14 @@
 
 '''netplan set command line'''
 
-import os
-import yaml
 import tempfile
 import re
-import logging
-import shutil
-import glob
+import io
 
 from netplan.cli.utils import NetplanCommand
 import netplan.libnetplan as libnetplan
-from netplan.configmanager import ConfigManager
 
-FALLBACK_HINT = '70-netplan-set'
+FALLBACK_FILENAME = '70-netplan-set.yaml'
 GLOBAL_KEYS = ['renderer', 'version']
 
 
@@ -54,151 +49,37 @@ class NetplanSet(NetplanCommand):
         self.parse_args()
         self.run_command()
 
-    def is_emtpy_yaml(self, tree):
-        if isinstance(tree, dict) and list(tree.keys()) == ['network'] and tree['network'] is None:
-            return True
-        return False
-
-    def split_tree_by_hint(self, set_tree) -> (str, dict):
-        network = set_tree.get('network', {})
-        # A mapping of 'origin-hint' -> YAML tree (one subtree per netdef)
-        subtrees = dict()
-        for devtype in network:
-            if devtype in GLOBAL_KEYS:
-                continue  # special handling of global keys down below
-            devtype_content = network.get(devtype, [])
-            # Special case: removal of a whole devtype.
-            # We replace the devtype null node with a dict of all defined netdefs
-            # set to None.
-            if devtype_content is None:
-                devtype_content = {dev: None for dev in libnetplan.netplan_get_ids_for_devtype(devtype, self.root_dir)}
-                network[devtype] = devtype_content
-            for netdef in devtype_content:
-                hint = FALLBACK_HINT
-                filename = libnetplan.netplan_get_filename_by_id(netdef, self.root_dir)
-                if filename:
-                    hint = os.path.basename(filename)[:-5]  # strip prefix and .yaml
-                netdef_tree = {'network': {devtype: {netdef: network.get(devtype).get(netdef)}}}
-                # Merge all netdef trees which are going to be written to the same file/hint
-                subtrees[hint] = self.merge(subtrees.get(hint, {}), netdef_tree)
-
-        # Merge GLOBAL_KEYS into one of the available subtrees
-        # Write to same file (if only one hint/subtree is available)
-        # Write to FALLBACK_HINT if multiple hints/subtrees are available, as we do not know where it is supposed to go
-        if any(network.get(key) for key in GLOBAL_KEYS):
-            # Write to the same file, if we have only one file-hint or to FALLBACK_HINT otherwise
-            hint = list(subtrees)[0] if len(subtrees) == 1 else FALLBACK_HINT
-            for key in GLOBAL_KEYS:
-                tree = {'network': {key: network.get(key)}}
-                subtrees[hint] = self.merge(subtrees.get(hint, {}), tree)
-
-        # return a list of (str:hint, dict:subtree) tuples
-        return subtrees.items()
-
     def command_set(self):
         if self.origin_hint is not None and len(self.origin_hint) == 0:
             raise Exception('Invalid/empty origin-hint')
+        if self.origin_hint:
+            filename = '.'.join((self.origin_hint, 'yaml'))
+        else:
+            filename = None
         split = self.key_value.split('=', 1)
         if len(split) != 2:
             raise Exception('Invalid value specified')
+
         key, value = split
-        set_tree = self.parse_key(key, yaml.safe_load(value))
+        if not key.startswith('network'):
+            key = '.'.join(('network', key))
 
-        # special case: clear all YAML (or a specific hint file) if "network=null" is set
-        if self.is_emtpy_yaml(set_tree):
-            path = os.path.join('etc', 'netplan')
-            if self.origin_hint:  # clear specific hint file, it it does exist
-                hint_path = os.path.join(self.root_dir, path, self.origin_hint + '.yaml')
-                if os.path.isfile(hint_path):
-                    os.remove(hint_path)
-            else:  # clear all YAML files in <ROOT_DIR>/etc/netplan/*.yaml
-                yaml_files = glob.glob(os.path.join(self.root_dir, path, '*.yaml'))
-                for f in yaml_files:
-                    os.remove(f)
-            return
+        # Split the string into a list on the dot separators, and unescape the remaining dots
+        yaml_path = [s.replace(r'\.', '.') for s in re.split(r'(?<!\\)\.', key)]
 
-        hints = [(self.origin_hint, set_tree)]
-        # Override YAML config in each individual netdef file if origin-hint is not set
-        if self.origin_hint is None:
-            hints = self.split_tree_by_hint(set_tree)
+        parser = libnetplan.Parser()
+        with tempfile.TemporaryFile() as tmp:
+            libnetplan.create_yaml_patch(yaml_path, value, tmp)
+            tmp.flush()
+            tmp.seek(0, io.SEEK_SET)
+            parser.load_nullable_fields(tmp)
+            parser.load_yaml_hierarchy(self.root_dir)
+            tmp.seek(0, io.SEEK_SET)
+            parser.load_yaml(tmp)
 
-        for hint, subtree in hints:
-            self.write_file(subtree, hint + '.yaml', self.root_dir)
-
-    def parse_key(self, key, value):
-        # The 'network.' prefix is optional for netsted keys, its always assumed to be there
-        if not key.startswith('network.') and not key == 'network':
-            key = 'network.' + key
-        # Split at '.' but not at '\.' via negative lookbehind expression
-        split = re.split(r'(?<!\\)\.', key)
-        tree = {}
-        i = 1
-        t = tree
-        for part in split:
-            part = part.replace('\\.', '.')  # Unescape interface-ids, containing dots
-            val = {}
-            if i == len(split):
-                val = value
-            t = t.setdefault(part, val)
-            i += 1
-        return tree
-
-    def merge(self, a, b, path=None):
-        """
-        Merges tree/dict 'b' into tree/dict 'a'
-        """
-        if path is None:
-            path = []
-        for key in b:
-            if key in a:
-                if isinstance(a[key], dict) and isinstance(b[key], dict):
-                    self.merge(a[key], b[key], path + [str(key)])
-                elif b[key] is None:
-                    del a[key]
-                else:
-                    # Overwrite existing key with new key/value from 'set' command
-                    a[key] = b[key]
-            else:
-                a[key] = b[key]
-        return a
-
-    def write_file(self, set_tree, name, rootdir='/'):
-        tmproot = tempfile.TemporaryDirectory(prefix='netplan-set_')
-        path = os.path.join('etc', 'netplan')
-        os.makedirs(os.path.join(tmproot.name, path))
-
-        config = {'network': {}}
-        absp = os.path.join(rootdir, path, name)
-        # check stat(absp), as we don't care about empty hint files
-        if os.path.isfile(absp) and os.stat(absp).st_size > 0:
-            with open(absp, 'r') as f:
-                c = yaml.safe_load(f)
-                if c is not None:  # ignore empty file, even if it contains whitespace
-                    config = c
-
-        new_tree = self.merge(config, set_tree)
-        stripped = ConfigManager.strip_tree(new_tree)
-        logging.debug('Writing file {}: {}'.format(name, stripped))
-        if 'network' in stripped and list(stripped['network'].keys()) == ['version']:
-            # Clear file if only 'network: {version: 2}' is left
-            logging.debug('Empty YAML, deleting file {}'.format(absp))
-            if os.path.isfile(absp):
-                os.remove(absp)
-        elif 'network' in stripped:
-            tmpp = os.path.join(tmproot.name, path, name)
-            with open(tmpp, 'w+') as f:
-                new_yaml = yaml.dump(stripped, indent=2, default_flow_style=False)
-                f.write(new_yaml)
-            # Validate the newly created file, by parsing it via libnetplan
-            libnetplan.netplan_parse(tmpp)
-            # Valid, move it to final destination
-            shutil.copy2(tmpp, absp)
-            os.remove(tmpp)
-        elif stripped == {}:
-            # Clear file (if it exists) if the last/only key got removed
-            # do nothing otherwise
-            logging.debug('Removed last key from YAML, deleting file {}'.format(absp))
-            if os.path.isfile(absp):
-                os.remove(absp)
-        else:  # pragma nocover
-            raise Exception('Invalid input: {}'.format(set_tree))
+        state = libnetplan.State()
+        state.import_parser_results(parser)
+        if self.origin_hint:
+            state.write_yaml_file(filename, self.root_dir)
+        else:
+            state.update_yaml_hierarchy(FALLBACK_FILENAME, self.root_dir)
--- a/src/netplan.c
+++ b/src/netplan.c
@@ -1061,3 +1061,83 @@ netplan_state_dump_yaml(const NetplanState* np_state, int out_fd, GError** error
 
     return netplan_netdef_list_write_yaml(np_state, np_state->netdefs_ordered, out_fd, error);
 }
+
+/**
+ * Regenerate the YAML configuration files from a given state. Any state that
+ * hasn't an associated filepath will use the default_filename output in the
+ * standard config directory.
+ *
+ * @np_state: the state for which to generate the config
+ * @default_filename: Default config file, cannot be NULL or empty
+ * @rootdir: If not %NULL, generate configuration in this root directory
+ *           (useful for testing).
+ */
+gboolean
+netplan_state_update_yaml_hierarchy(const NetplanState* np_state, const char* default_filename, const char* rootdir, GError** error)
+{
+    g_autofree gchar *default_path = NULL;
+    gboolean ret = FALSE;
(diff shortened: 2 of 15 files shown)


schopin-pro:
This builds up on #252 hence the draft status.

src/netplan.c
```diff
@@ -991,6 +991,52 @@ netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int
...
+    if (to_write == NULL)
+        return TRUE;
+
+    int out_fd = open(path, O_WRONLY | O_CREAT | O_TRUNC, 0640);
+    if (out_fd < 0) {
+        g_set_error(error, G_FILE_ERROR, errno, "%s", strerror(errno));
+        return FALSE;
+    }
+
+    gboolean ret = netplan_netdef_list_write_yaml(np_state, to_write, out_fd, error);
```
schopin-pro, line 1035: `to_write` leaks.

schopin-pro:
Ran the autopkgtests manually:

```
autopkgtest [17:20:13]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              PASS
vlans                PASS
wifi                 PASS
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
```

src/util.c
```diff
@@ -523,10 +523,10 @@ netplan_delete_connection(const char* id, const char* rootdir)
     del = g_strdup_printf("network.%s.%s=NULL", netplan_def_type_name(nd->type), id);
 
     /* TODO: refactor logic to actually be inside the library instead of spawning another process */
-    const gchar *argv[] = { SBINDIR "/" "netplan", "set", del, "--origin-hint" , filename, NULL, NULL, NULL };
```
slyon, line 418: We should delete the `filename` variable processing above, as we don't make any use of it anymore.
Especially this part:
```
    filename = g_path_get_basename(nd->filename);
    filename[strlen(filename) - 5] = '\0'; //stip ".yaml" suffix
```

tests/test_cli_get_set.py
```diff
@@ -63,53 +63,54 @@ def test_set_scalar(self):
...
+            self.assertIs(True, yaml.safe_load(f)['network']['ethernets']['eth0']['dhcp4'])
 
     def test_set_scalar2(self):
         self._set(['ethernets.eth0.dhcp4="yes"'])
         self.assertTrue(os.path.isfile(self.path))
         with open(self.path, 'r') as f:
-            self.assertIn('network:\n  ethernets:\n    eth0:\n      dhcp4: \'yes\'', f.read())
+            # XXX: the previous version using PyYAML would keep the "yes" variant but the round-trip
+            # through libnetplan doesn't keep formatting choices (yes is a keyword same as true)
+            self.assertIs(True, yaml.safe_load(f)['network']['ethernets']['eth0']['dhcp4'])
```
slyon, line 74: thought (non-blocking): Could we somehow make use of the new "dirty" tracking functionality to keep formatting choices?

tests/test_cli_get_set.py
```diff
@@ -192,15 +193,17 @@ def test_set_append(self):
             f.write('''network:
   version: 2
   ethernets:
+    ens2: {dhcp4: yes}
```
slyon, line 196: nitpick (non-blocking): This looks unrelated and unused, why was it introduced?

tests/test_cli_get_set.py
```diff
@@ -210,39 +213,42 @@ def test_set_overwrite_eq(self):
...
+    ens3: {dhcp4: "no"}''')
         self._set(['ethernets.ens3.dhcp4=true'])
         self.assertTrue(os.path.isfile(self.path))
         with open(self.path, 'r') as f:
-            out = f.read()
-            self.assertIn('network:\n  ethernets:\n', out)
-            self.assertIn('    ens3:\n      dhcp4: true', out)
+            self.assertIs(
+                    True,
+                    yaml.safe_load(f)['network']['ethernets']['ens3']['dhcp4'])
```
slyon, line 230: (non-blocking): We could change the formatting to make this look a bit nicer:
```
out = yaml.safe_load(f)
self.assertTrue(out['network']['ethernets']['ens3']['dhcp4'])
```

tests/test_cli_get_set.py
```diff
@@ -300,9 +432,9 @@ def test_set_override_existing_file(self):
         self.assertFalse(os.path.isfile(self.path))
         self.assertTrue(os.path.isfile(override))
         with open(override, 'r') as f:
-            out = f.read()
-            self.assertIn('network:\n  ethernets:\n    eth0:\n      dhcp4: false', out)  # new
-            self.assertIn('eth1:\n      dhcp6: false', out)  # old
+            out = yaml.safe_load(f)
+            self.assertIs(False, out['network']['ethernets']['eth0']['dhcp4'])
+            self.assertIs(False, out['network']['ethernets']['eth1']['dhcp6'])
```
slyon, line 437: (non-blocking): I think we could be using `assertFalse()` here and below.

tests/test_cli_get_set.py
```diff
@@ -321,25 +454,25 @@ def test_set_override_multiple_existing_files(self):
...
-                    r'eth0.2:{dhcp4: false}},'
-                    r'bridges:{'
-                    r'br99:{dhcp4: false}}}')])
+        self._set([(r'network={renderer: NetworkManager, version: 2, '
+                    r'ethernets: {'
+                    r'eth1: {dhcp4: false},'
+                    r'eth0.1: {dhcp4: false},'
+                    r'eth0.2: {dhcp4: false}},'
+                    r'bridges: {'
+                    r'br99: {dhcp4: false}}}')])
```
slyon, line 463: question: Is this just some cleanup or an actual change in behavior, that we now require properly space separated mappings? We would need to take care not to break any consumer (like our own `netplan-dbus` daemon and/or `snapd`'s get/set functionality).

slyon (commented):
This is another big PR, but the change makes the implementation so much cleaner and more explicit. I really like your overall approach!

I left several comments inline about some details that we should address in a 2nd version of PR.

schopin-pro:
For the sake of being able to test stuff, I included the commit of #257 since without it the null-handling breaks stuff.

src/parse.c
```diff
@@ -2907,3 +2986,61 @@ netplan_parser_clear(NetplanParser** npp_p)
...
+    yaml_node_pair_t* entry;
+    switch (node->type) {
+        // LCOV_EXCL_START
+        case YAML_NO_NODE:
+            g_hash_table_insert(null_fields, key_prefix, key_prefix);
+            key_prefix = NULL;
+            break;
+        // LCOV_EXCL_STOP
+        case YAML_SCALAR_NODE:
+            if (       g_ascii_strcasecmp("null", scalar(node)) == 0
```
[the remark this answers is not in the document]
schopin-pro: We already do. `g_ascii_strcasecmp` is case-insensitive.

tests/test_cli_get_set.py
schopin-pro: Sadly, no. `assertFalse` and `assertTrue` coerce the tested value into a boolean, and we want to make sure we have the actual boolean values here, not any truthy or falsy value.

tests/test_cli_get_set.py
```diff
@@ -63,53 +63,54 @@ def test_set_scalar(self):
         self._set(['ethernets.eth0.dhcp4=true'])
         self.assertTrue(os.path.isfile(self.path))
         with open(self.path, 'r') as f:
-            self.assertIn('network:\n  ethernets:\n    eth0:\n      dhcp4: true', f.read())
+            self.assertIs(True, yaml.safe_load(f)['network']['ethernets']['eth0']['dhcp4'])
```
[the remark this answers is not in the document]
schopin-pro: As commented elsewhere, there's a subtle difference between `assertIs(True, ` and `assertTrue(`, the latter doing a boolean conversion, which we don't want here.

tests/test_cli_get_set.py
schopin-pro: I suppose we could, by storing the actual scalar data as the value in the hashtable. I'm not sure it's worth the hassle, though.

tests/test_cli_get_set.py
schopin-pro: Probably a leftover of some debugging session or another. I'll drop it.

tests/test_cli_get_set.py
schopin-pro: Oh crap. It seems it is a change in behavior, I get `netplan.libnetplan.LibNetplanException: Error parsing YAML: found unexpected ':'` when reverting the changes.
This seems weird to me, as the underlying parsing library is the same, isn't it?

FWIW, I checked the YAML spec, and at least in 1.1 and 1.2 the whitespace after the colon is mandatory.

slyon: I don't know for sure if PyYAML is using libyaml in the background, too. If those two YAML implementations differ that would be a problem, especially as we have consumers using that (out of spec?) syntax: https://github.com/snapcore/snapd/pull/10752/files#diff-e380767e9d06e93d76ea779341439e7d4c878b2708bcc23731453c17c2442616R557

I wonder if we can somehow handle both cases and maybe print a warning that this does not comply with the YAML specification?

schopin-pro: After investigations, I note the following things:

* Even though PyYAML loads libyaml, it does *not* use it in the default case. You have to explicitly use the CLoader class in order to do so. By default it uses a pure Python parser, which is apparently less strict than the C one.
* When using CLoader on the original string, I get the same error.
* However, when using the one in the snapd source code, it loads just fine. The difference is that the snapd code quotes their keys, since they're using a JSON serializer to format their string (which is valid since JSON is a subset of YAML). So, on that front I think we're OK.

src/parse.c
```diff
@@ -187,7 +233,7 @@ netplan_netdef_new(NetplanParser *npp, const char* id, NetplanDefType type, Netp
 
 typedef gboolean (*node_handler) (NetplanParser* npp, yaml_node_t* node, const void* data, GError** error);
 
-typedef gboolean (*custom_map_handler) (NetplanParser* npp, yaml_node_t* node, const void* data, GError** error);
+typedef gboolean (*custom_map_handler) (NetplanParser* npp, yaml_node_t* node, const char *prefix, const void* data, GError** error);
```
slyon, line 236: key_prefix?

src/parse.c
```diff
@@ -2873,6 +3021,17 @@ netplan_parser_reset(NetplanParser* npp)
     }
 
     npp->missing_ids_found = 0;
+
+    if (npp->null_fields) {
+        g_hash_table_destroy(npp->null_fields);
+        npp->null_fields = NULL;
+    }
+
+    if(npp->sources) {
```
slyon, line 3030:
```suggestion
    if (npp->sources) {
```

src/types.c
```diff
@@ -384,6 +394,12 @@ netplan_state_reset(NetplanState* np_state)
 
     np_state->backend = NETPLAN_BACKEND_NONE;
     reset_ovs_settings(&np_state->ovs_settings);
+
+    if(np_state->sources) {
```
slyon, line 398:
```suggestion
    if (np_state->sources) {
```

tests/test_cli_get_set.py
slyon: ACK. we could still do this in the future, if needed. Let's keep it out of this (already big) PR.

slyon (commented):
Thank you very much, we're making good progress with this big PR and are getting much closer to a unified YAML parser. kudos!

I've left some smaller inline comments, nothing critical.

Another thing to mention is that `make check-coverage` currently fails, due to `netplan/cli/sriov.py:363`, which can probably just be marked as "nocover", as it's just about (non-existing) hardware specific quirks.

tests/test_cli_get_set.py
schopin-pro: To close up this thread: I've added a new commit that explicitly switches this to a JSON-payload to match the snapd use-case.

src/netplan.c
```diff
@@ -989,6 +989,53 @@ netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int
...
+gboolean
+netplan_state_write_yaml_file(const NetplanState* np_state, const char* filename, const char* rootdir, GError** error)
+{
+    GList* iter = np_state->netdefs_ordered;
+    g_autofree gchar *path = NULL;
+    GList* to_write = NULL;
+
+    path = g_build_path(G_DIR_SEPARATOR_S, rootdir ?: G_DIR_SEPARATOR_S, "etc", "netplan", filename, NULL);
+
+    if (unlink(path) && errno != ENOENT) {
```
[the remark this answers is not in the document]
schopin-pro: Thanks for this, that was a good idea! I used a slightly different approach with similar results: create a temporary file /etc/netplan/FILENAME.yaml.$RANDOM, write stuff to it, and rename it to the proper filename if all is well.

src/netplan.c
```diff
@@ -1039,6 +1039,76 @@ netplan_state_dump_yaml(const NetplanState* np_state, int out_fd, GError** error
...
+            g_hash_table_insert(perfile_netdefs, default_path, NULL);
+        }
+    } else {
+        GList* iter = np_state->netdefs_ordered;
+        while (iter) {
+            NetplanNetDefinition* netdef = iter->data;
+            const char* filename = netdef->filename ? netdef->filename : default_path;
+            GList* list = NULL;
+            g_hash_table_steal_extended(perfile_netdefs, filename, NULL, (gpointer*)&list);
+            g_hash_table_insert(perfile_netdefs, (gpointer)filename, g_list_append(list, netdef));
```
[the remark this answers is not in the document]
schopin-pro: NAK. Except for `default_path`, our keys are owned by a `netdef` object so we simply cannot have any destructor function for keys. We *could* strdup them, but IMHO the current code works fine as is. Furthermore, we need to actually do the steal + insert, as we need the previous value to construct the new one *and* we really don't want to run the destructor on it.

src/netplan.c
```diff
@@ -1050,3 +1050,76 @@ netplan_state_dump_yaml(const NetplanState* np_state, int out_fd, GError** error
...
+ */
+gboolean
+netplan_state_update_yaml_hierarchy(const NetplanState* np_state, const char* default_filename, const char* rootdir, GError** error)
+{
+    g_autofree gchar *default_path = NULL;
+    gboolean ret = FALSE;
+    GHashTableIter hash_iter;
+    gpointer key, value;
+    GHashTable *perfile_netdefs = g_hash_table_new_full(g_str_hash, g_str_equal, NULL, (GDestroyNotify)g_list_free);
+    default_path = g_build_path(G_DIR_SEPARATOR_S, rootdir ?: G_DIR_SEPARATOR_S, "etc", "netplan", default_filename, NULL);
```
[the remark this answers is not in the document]
schopin-pro: Fixed by using g_assert (since we do it elsewhere for netdef validation).

netplan/cli/commands/set.py
```diff
@@ -54,151 +49,37 @@ def run(self):
...
-            if os.path.isfile(absp):
-                os.remove(absp)
-        else:  # pragma nocover
-            raise Exception('Invalid input: {}'.format(set_tree))
+        key, value = split
+        if not key.startswith('network'):
+            key = '.'.join(('network', key))
+
+        # Replace the '.' with '\t' but not at '\.' via negative lookbehind expression
+        yaml_path = [s.replace(r'\.', '.') for s in re.split(r'(?<!\\)\.', key)]
```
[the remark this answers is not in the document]
schopin-pro: No we're not, we're actually splitting the string into a list, and replacing the escaped dots by their standard counterpart. The '\t' encoding is done in the bindings, as it's an implementation detail of the C library.

I fixed the comment :)

src/netplan.c
```diff
@@ -1050,3 +1050,76 @@ netplan_state_dump_yaml(const NetplanState* np_state, int out_fd, GError** error
...
+    }
+
+    /* Remove any referenced source file that doesn't have any associated data.
+       Presumably, it is data that has been obsoleted by files loaded
+       afterwards, typically via `netplan set`. */
+    if (np_state->sources) {
+        g_hash_table_iter_init(&hash_iter, np_state->sources);
+        while (g_hash_table_iter_next (&hash_iter, &key, &value)) {
+            if (!g_hash_table_contains(perfile_netdefs, key))
+                g_unlink(key);
```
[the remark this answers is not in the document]
schopin-pro: Done, with special casing for the "file does not exist" error which is ignored.

slyon (approved):
Thank you! All my remarks have been addressed. This is ready to be merged IMO.

Abigail is detecting some ABI changes, but they are all compatible (added functions & appended `sources` to `NetplanState`). Feel free to update the ABI XML to something like this (after a rebase, to have `abi-compat/focal_0.104.xml` available): https://paste.ubuntu.com/p/BK8HGnzPrG/ and merge this PR afterwards.
```
abidiff abi-compat/focal_0.104.xml libnetplan.so.0.0 --headers-dir2 include/ --header-file2 src/abi.h
Functions changes summary: 0 Removed, 0 Changed (26 filtered out), 5 Added functions
Variables changes summary: 0 Removed, 1 Changed, 0 Added variable

5 Added functions:

  [A] 'function gboolean netplan_parser_load_nullable_fields(NetplanParser*, int, GError**)'    {netplan_parser_load_nullable_fields}
  [A] 'function gboolean netplan_parser_load_yaml_from_fd(NetplanParser*, int, GError**)'    {netplan_parser_load_yaml_from_fd}
  [A] 'function gboolean netplan_state_update_yaml_hierarchy(const NetplanState*, const char*, const char*, GError**)'    {netplan_state_update_yaml_hierarchy}
  [A] 'function gboolean netplan_state_write_yaml_file(const NetplanState*, const char*, const char*, GError**)'    {netplan_state_write_yaml_file}
  [A] 'function gboolean netplan_util_create_yaml_patch(const char*, const char*, int, GError**)'    {netplan_util_create_yaml_patch}

1 Changed variable:

  [C] 'NetplanState global_state' was changed at abi_compat.c:6[4](https://github.com/canonical/netplan/runs/6174614861?check_suite_focus=true#step:5:4):1:
    size of symbol changed from 1[6](https://github.com/canonical/netplan/runs/6174614861?check_suite_focus=true#step:5:6)8 to 1[7](https://github.com/canonical/netplan/runs/6174614861?check_suite_focus=true#step:5:7)6
```

slyon:
> IMHO we should completely exclude `global_state` from the ABI check

ACK. Should be resolved by a4a8ac8264c613315b314d3c395dd0985fa63743

src/netplan.c
```diff
@@ -989,6 +989,64 @@ netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int
...
+        if (g_strcmp0(fname, path) == 0)
+            to_write = g_list_append(to_write, netdef);
+        iter = iter->next;
+    }
+
+    /* Remove any existing file if there is no data to write */
+    if (to_write == NULL) {
+        if (unlink(path) && errno != ENOENT) {
+            g_set_error(error, G_FILE_ERROR, errno, "%s", strerror(errno));
+            return FALSE;
```
slyon, line 1025: This is not covered with the current testing coverage. Otherwise, I think we're good to merge this after a4a8ac8264c613315b314d3c395dd0985fa63743 landed

src/netplan.c
```diff
@@ -1003,3 +1061,83 @@ netplan_state_dump_yaml(const NetplanState* np_state, int out_fd, GError** error
...
+
+    /* Remove any referenced source file that doesn't have any associated data.
+       Presumably, it is data that has been obsoleted by files loaded
+       afterwards, typically via `netplan set`. */
+    if (np_state->sources) {
+        g_hash_table_iter_init(&hash_iter, np_state->sources);
+        while (g_hash_table_iter_next (&hash_iter, &key, &value)) {
+            if (!g_hash_table_contains(perfile_netdefs, key)) {
+                if (unlink(key) && errno != ENOENT)
+                    goto file_error;
```
slyon, line 1128: Same here, needs coverage or exclusion.

schopin-pro: I went for exclusion for this one as testing it out was really convoluted for relatively little gain.

schopin-pro:
Regarding editing the ABI doc, shouldn't we hold until release? That way we can still tweak symbols introduced in previous PRs

slyon:
> Regarding editing the ABI doc, shouldn't we hold until release? That way we can still tweak symbols introduced in previous PRs

Yes, makes sense! Feel free to merge this PR.

PR #257, opened 2022-02-01
parse: fix the handle_match_driver type

schopin-pro:
The YAML_NO_NODE type marker means that the callback called is via the
`.variable` union field, not `.generic`. Since this is a union type, and
both callbacks have the same signature for now, it works, but future
evolutions of the code will make those signatures evolve independently,
leading to hard-to-debug issues.

## Checklist


slyon (approved):
LGTM!

PR #255, opened 2022-01-11
YAML consolidation: final patch set for `configmanager.py` (FR-702)

schopin-pro:
## Description

This patch set, the last one in the YAML consolidation series, ports ConfigManager to use libnetplan's YAML parser. In order to do so, we had to expose a lot of the internal data of the Netplan state to the Python bindings, as the `netplan apply` logic needs to inspect the definitions to know what to do.

## Checklist

- [x] Runs `make check` successfully.
- [ ] Retains 100% code coverage (`make check-coverage`).

PR #259, opened 2022-02-05
Have ConfigManager cleanup its temporary directory on destruction (LP #1959729)

Caligatio:
## Description
`ConfigManager` creates a temporary directory on instantiation and relies on the caller to clean up that temporary directory despite never exposed directly to the caller. Due to the code structure in several places, it would be onerous to call `ConfigManager.cleanup()` everywhere it's needed so this patch instead calls `.cleanup()` whenever the ConfigManager instance is garage collected.

This addresses https://bugs.launchpad.net/netplan/+bug/1959729

## Checklist

### DIFF
--- a/netplan/configmanager.py
+++ b/netplan/configmanager.py
@@ -224,6 +224,13 @@ class ConfigManager(object):
     def cleanup(self):
         shutil.rmtree(self.tempdir)
 
+    def __del__(self):
+        try:
+            self.cleanup()
+        except FileNotFoundError:
+            # If cleanup() was called before, there is nothing to delete
+            pass
+
     def _copy_file(self, src, dst):
         shutil.copy(src, dst)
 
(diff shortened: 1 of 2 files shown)


slyon (approved):
Thank you Brian, this issue has been bugging me for a long time!
Your code provides a clean and self-contained solution to this problem and I like this approach.

`configmanager.py` is currently being reworked heavily (in https://github.com/canonical/netplan/pull/255) but the tmpdir cleanup on destruction is a good thing to have independently. All tests and coverage are passing (I ran the integration tests locally, see below). This is ready for merging.

```
autopkgtest [10:44:40]: @@@@@@@@@@@@@@@@@@@@ summary
ovs                  PASS
ethernets            PASS
bridges              PASS
bonds                PASS
routing              PASS
vlans                PASS
wifi                 PASS
tunnels              PASS
scenarios            PASS
regressions          PASS
autostart            PASS
cloud-init           PASS
```

PR #260, opened 2022-02-07
cli:apply: fix potential race with rename/creation of netdevs and start networkd if off (LP: #1962095)

slyon:
## Description
Calling `networkctl_reload()` before `networkd_interfaces()` makes sure that newly created netdevs will show up in the interface list.

Making use of the index number (instead of interface name) ensures that renamed interfaces are properly reconfigured even if they changed their name.

Starting networkd if not active is making sure we can process network changes even if systemd-networkd was stopped before (e.g. by subiquity), see LP#1962095

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad: LP#1962095

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -252,8 +252,13 @@ class NetplanApply(utils.NetplanCommand):
                            if not f.endswith('/' + OVS_CLEANUP_SERVICE)]
             # Run 'systemctl start' command synchronously, to avoid race conditions
             # with 'oneshot' systemd service units, e.g. netplan-ovs-*.service.
-            utils.networkctl_reload()
-            utils.networkctl_reconfigure(utils.networkd_interfaces())
+            try:
+                utils.networkctl_reload()
+                utils.networkctl_reconfigure(utils.networkd_interfaces())
+            except subprocess.CalledProcessError:
+                # (re-)start systemd-networkd if it is not running, yet
+                logging.warning('Falling back to a hard restart of systemd-networkd.service')
+                utils.systemctl('restart', ['systemd-networkd.service'], sync=True)
             # 1st: execute OVS cleanup, to avoid races while applying OVS config
             utils.systemctl('start', [OVS_CLEANUP_SERVICE], sync=True)
             # 2nd: start all other services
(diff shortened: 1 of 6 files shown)


slyon:
This is related to https://github.com/snapcore/core-base/issues/31

slyon:
@valentindavid just confirmed via PM that this PR fixes the issue they see on a real system. Tested via PPA https://launchpad.net/~slyon/+archive/ubuntu/netplan/+sourcepub/13303987/+listing-archive-extra

netplan/cli/commands/apply.py
```diff
@@ -252,7 +252,13 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
...
             # Run 'systemctl start' command synchronously, to avoid race conditions
             # with 'oneshot' systemd service units, e.g. netplan-ovs-*.service.
-            utils.networkctl_reconfigure(utils.networkd_interfaces())
+            try:
+                utils.networkctl_reload()
+                utils.networkctl_reconfigure(utils.networkd_interfaces())
+            except subprocess.CalledProcessError:
+                # (re-)start systemd-networkd if it is not running, yet
+                logging.warning('Falling back to a hard restart of systemd-networkd.service')
+                utils.systemctl('restart', ['systemd-networkd.service'], sync=True)
```
xnox, line 261: Generally, I don't like try/except blocks, and especially when they wrap two commands without distinction if networkd is not running (hence networctl fails to connect to networkd) or if there is something broken with the config (and thus reconfigure of a device fails, but networkd is running and gets pointlessly restarted).

Can you check if systemd-networkd.service is running? If yes, execute networkctl reload & reconfigure. Otherwise check if networkd is masked and fail. If it is not masked, restart it.

slyon: > Can you check if systemd-networkd.service is running? If yes, execute networkctl reload & reconfigure. Otherwise restart networkd.

That is actually how I had it implemented initially, but it does not catch cases where systemd-networkd.service is running but "stuck" (for some unkown reason), as reported in the original bug report at LP#1962095 (comment 4). Sure, networkd being stuck is not really a netplan problem... but using the try/except block we can hit two birds with one stone.

sil2100 (approved):
This looks good to me, and feels sufficiently safe. Sure, as per Dimitri's comment it might lead to a case where there's a pointless restart, but I don't think that will be too frequent, and other ways might just unnecessarily make the logic more complicated. So I'd be +1 on getting this merged as-is. But I leave it up to you!

PR #227, opened 2021-09-15
API/ABI: restrict the symbol export to a determined public API

schopin-pro:
## Description

The goal of this PR is to properly determine what the public interface of libnetplan is. There is potentially a bit of ABI breakage, as symbols now default to hidden status unless explicitly marked otherwise, and said marking was done manually by looking at the current consumers I had on my system and trying to fill in the blanks.

The API, on the other hand, has been conscientiously broken, as most of what was exposed in the headers are now safely hidden away in our internal headers. This is by design, and AFAICT should not break existing code.

This PR depends on #230, as denoted by the merge commit in there. It should make independent review a bit easier :).

Given the fact that the main purpose of this PR is to limit the amount of symbols we export, we can safely assume that strictly speaking, there is ABI breakage. However, the whole point was to only expose the symbols that are in actual use out there in the real world.

As usual, the patchset has been edited expecting a review commit by commit, as the whole diff can be a bit daunting.

## Checklist

- [x] Runs `make check` successfully.
- [x] Retains 100% code coverage (`make check-coverage`).
- [x] New/changed keys in YAML format are documented.

PR #262, opened 2022-02-15
Drop unused legacy/compat API

slyon:
## Description
Libnetplan at out latest version 0.103 did not expose several symbols that we carry along (and would newly introduce as "legacy"/"compat" symbols in 0.104). This is due to the fact that the `Makefile` changed and those functions had been linked directly into the `generate` binary before, whereas they are now part of `libnetplan.so`. As those symbols have not been part of the library before, we should not introduce them now as legacy symbols, as no consumer can be using it from the old library.

Old Makefile at 0.103:
```
libnetplan.so.$(NETPLAN_SOVER): parse.o netplan.o util.o validation.o error.o parse-nm.o
	$(CC) -shared -Wl,-soname,libnetplan.so.$(NETPLAN_SOVER) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ `pkg-config --libs glib-2.0 gio-2.0 yaml-0.1`
	ln -snf libnetplan.so.$(NETPLAN_SOVER) libnetplan.so

generate: libnetplan.so.$(NETPLAN_SOVER) nm.o networkd.o openvswitch.o generate.o sriov.o
	$(CC) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -o $@ $^ -L. -lnetplan `pkg-config --cflags --libs glib-2.0 gio-2.0 yaml-0.1 uuid`
```

## Checklist


schopin-pro:
Indeed, prior to #227 those functions were statically linked into `generate`.

I'm liking the added bonus of actual error checking in the ovs module, too :)

LGTM!

PR #263, opened 2022-02-16
nm:parse-nm: handle differing ip6-privacy default value

slyon:
## Description
Commit 64e381e22bcc71563644454b41e852ec4eed8a1c (#244) changed netplan's
behaviour to always write out the ipv6-privacy value (default to 0=off),
in order to comply with what is written in our documentation and how it is
implemented in the networkd backend.

This is a problem for our NetworkManager integration, as NM's default value
is -1=unknown. Netplan only supports values 0=off and 2=on while NM supports
-1, 0, 1 and 2. We need to handle unsupported cases (-1 and 1) via passthrough.

## Checklist


schopin-pro (approved):
LGTM

PR #267, opened 2022-03-12
Makefile: fix parallel build failure

yizhao1:
Add src/_features.h as dependency for dbus.o to fix the parallel build
failure:
src/dbus.c:17:10: fatal error: _features.h: No such file or directory

Signed-off-by: Yi Zhao <yi.zhao@windriver.com>

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,11 +52,8 @@ NOSETESTS3 ?= $(shell command -v nosetests-3 || command -v nosetests3 || echo tr
 
 default: netplan/_features.py generate netplan-dbus dbus/io.netplan.Netplan.service doc/netplan.html doc/netplan.5 doc/netplan-generate.8 doc/netplan-apply.8 doc/netplan-try.8 doc/netplan-dbus.8 doc/netplan-get.8 doc/netplan-set.8
 
-%.o: src/%.c
+%.o: src/%.c src/_features.h
 	$(CC) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -c $^ `pkg-config --cflags --libs glib-2.0 gio-2.0 yaml-0.1 uuid`
-
-dbus.o: src/dbus.c src/_features.h
-	$(CC) $(BUILDFLAGS) $(CFLAGS) $(LDFLAGS) -c $< `pkg-config --cflags --libs glib-2.0 gio-2.0 yaml-0.1 uuid`
 
 libnetplan.so.$(NETPLAN_SOVER): $(SRCS) abicompat.lds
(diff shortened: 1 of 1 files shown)


slyon (commented):
Hi, thank you very much for your contribution!

Could you please define how exactly we can reporduce the failure case?

I know `src/_features.c` is currently only being used in dbus.c, but I wonder if we could just add the dependency to the generic `%.o` rule, instead of duplicating it into a specific `dbus.o` rule (see inline comment/suggestion)?

That should keep the Makefile cleaner an easier to adopt in the future. What do you think?

yizhao1:
> Hi, thank you very much for your contribution!
> 
> Could you please define how exactly we can reporduce the failure case?
> 
> I know `src/_features.c` is currently only being used in dbus.c, but I wonder if we could just add the dependency to the generic `%.o` rule, instead of duplicating it into a specific `dbus.o` rule (see inline comment/suggestion)?
> 
> That should keep the Makefile cleaner an easier to adopt in the future. What do you think?



> 



> Hi, thank you very much for your contribution!
> 
> Could you please define how exactly we can reporduce the failure case?
> 
> I know `src/_features.c` is currently only being used in dbus.c, but I wonder if we could just add the dependency to the generic `%.o` rule, instead of duplicating it into a specific `dbus.o` rule (see inline comment/suggestion)?
> 
> That should keep the Makefile cleaner an easier to adopt in the future. What do you think?



> Hi, thank you very much for your contribution!
> 
> Could you please define how exactly we can reporduce the failure case?

We encountered this issue on Yocto master branch. In Yocto, we can set PARALLEL_MAKE to specify the maximum number of parallel threads make can run.
E.g: Add the following lines to conf/local.conf:
PARALLEL_MAKE = "-j 16"

This issue happens occasionally, so we can reproduce it by a loop command:
$ while true; bitbake netplan -c cleansstate && bitbake netplan || break; done 


> 
> I know `src/_features.c` is currently only being used in dbus.c, but I wonder if we could just add the dependency to the generic `%.o` rule, instead of duplicating it into a specific `dbus.o` rule (see inline comment/suggestion)?
> 
> That should keep the Makefile cleaner an easier to adopt in the future. What do you think?

It's OK for me.

//Yi

slyon:
> We encountered this issue on Yocto master branch. In Yocto, we can set PARALLEL_MAKE to specify the maximum number of parallel threads make can run. E.g: Add the following lines to conf/local.conf: PARALLEL_MAKE = "-j 16"
> 
> This issue happens occasionally, so we can reproduce it by a loop command: $ while true; bitbake netplan -c cleansstate && bitbake netplan || break; done
> 
> > I know `src/_features.c` is currently only being used in dbus.c, but I wonder if we could just add the dependency to the generic `%.o` rule, instead of duplicating it into a specific `dbus.o` rule (see inline comment/suggestion)?
> > That should keep the Makefile cleaner an easier to adopt in the future. What do you think?
> 
> It's OK for me.
> 
> //Yi

Hi Yi! Thanks for giving this explanation.
Although I cannot reproduce it locally (using `while true; do make clean && make -j 16 || break; done`), I can see what's happening here and that the current Makefile could hit this race condition occasionally.

I cleaned it up to modify the generic `%.o: src/%.c src/_features.h` rule and will merge after the tests are passing.

PR #268, opened 2022-03-17
Enable building via Meson build system

slyon:
## Description
Adding a Meson build environment in addition to the plain `Makefile` netplan is using currently.
The new build system should be able to support all the Makefile's features and implement them in a more standard conform way. Furthermore, using Meson we'll be able to extend the build-time features more easily in the future, like implementing build-time options and properly acknowledging dpkg buildflags (hardening/LTO/...).

Meson is easy to use, here are some sample commands, to do the usual netplan stuff:
* meson setup build --prefix=/usr [-Db_coverage=true]
* meson compile -C build
* meson test -C build --verbose [TEST_NAME]
* meson install -C build --destdir ../tmproot

All things are done inside the build directory, specified during the `meson setup` command ("build" in the example above), except the code generated in `src/_features.h` and `python/_features.py`, that is still inside the source directory.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,8 @@ clean:
 	find . | grep -E "(__pycache__|\.pyc)" | xargs rm -rf
 
 check: default linting
-	tests/cli.py
-	LD_LIBRARY_PATH=. $(NOSETESTS3) -v --with-coverage
+	PYTHONPATH=. LD_LIBRARY_PATH=. tests/cli.py
+	PYTHONPATH=. LD_LIBRARY_PATH=. $(NOSETESTS3) -v --with-coverage
 	tests/validate_docs.sh
 
 linting:
--- a/README.md
+++ b/README.md
@@ -18,10 +18,10 @@ The full documentation for netplan is available in the [doc/netplan.md file](../
 
 Steps to build netplan using the [Meson](https://mesonbuild.com) build system inside the `build/` directory:
 
-* meson setup build --prefix=/usr
+* meson setup build --prefix=/usr -Db_coverage=true
 * meson compile -C build
-* meson install -C build --destdir ../tmp && tree tmp/
-* meson test -C build
+* meson test -C build --verbose [TEST_NAME]
+* meson install -C build --destdir ../tmproot
 
 # Bug reports
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,15 @@ An overview of the architecture can be found at [netplan.io/design](https://netp
 
 The full documentation for netplan is available in the [doc/netplan.md file](../master/doc/netplan.md)
 
+# Build using Meson
+
+Steps to build netplan using the [Meson](https://mesonbuild.com) build system inside the `build/` directory:
+
+* meson setup build --prefix=/usr
+* meson compile -C build
+* meson install -C build --destdir ../tmp && tree tmp/
+* meson test -C build
+
 # Bug reports
 
 Please file bug reports in [Launchpad](https://bugs.launchpad.net/netplan/+filebug).
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The full documentation for netplan is available in the [doc/netplan.md file](../
 
 Steps to build netplan using the [Meson](https://mesonbuild.com) build system inside the `build/` directory:
 
-* meson setup build --prefix=/usr -Db_coverage=true
+* meson setup build --prefix=/usr [-Db_coverage=true]
 * meson compile -C build
 * meson test -C build --verbose [TEST_NAME]
 * meson install -C build --destdir ../tmproot
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,0 +1,26 @@
+if pandoc.found()
+    custom_target(
+        input: ['manpage-header.md', 'netplan.md', 'manpage-footer.md'],
+        output: 'netplan.5',
+        command: [pandoc, '-s', '-o', '@OUTPUT@', '@INPUT@'],
+        install: true,
+        install_dir: join_paths(get_option('mandir'), 'man5'))
+    custom_target(
+        input: 'netplan.md',
+        output: 'netplan.html',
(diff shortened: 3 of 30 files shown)


.gitignore
```diff
@@ -10,3 +10,4 @@ __pycache__
 src/_features.h
 netplan/_features.py
 dbus/io.netplan.Netplan.service
+build/
```
eli-schwartz, line 13: This is not necessary. The builddir can be named anything, so you cannot predict it in gitignore... but Meson does know at the time of generation, and will create a file named `builddir/.gitignore` (or, if you prefer `build/`, it will create `build/.gitignore`) with the contents `*`, in order to ignore the entire contents of the generated build tree.

Meson knows that no files in a generated build tree ever have a reason to be checked into git, so it is easy to do the right thing here. :)

meson.build
```diff
@@ -0,0 +1,53 @@
...
+uuid = dependency('uuid')
+libsystemd = dependency('libsystemd')
+
+systemd = dependency('systemd')
+completions = dependency('bash-completion')
+systemd_generator_dir = systemd.get_variable(pkgconfig: 'systemdsystemgeneratordir')
+systemd_unit_dir = systemd.get_variable(pkgconfig: 'systemdsystemunitdir')
+bash_completions_dir = completions.get_variable(pkgconfig: 'completionsdir', default_value: '/etc/bash_completion.d')
+
+pandoc = find_program('pandoc', required: false)
```
eli-schwartz, line 24: You might want to use a meson_options.txt feature option for this. That would allow people to specify `-Ddoc=disabled` and force disable trying to detect pandoc or build anything with it.

doc/meson.build
```diff
@@ -0,0 +1,26 @@
...
+        custom_target(
+            input: markdown,
+            output: manpage,
+            command: [pandoc, '-s', '-o', '@OUTPUT@', '@INPUT@'],
+            install: true,
+            install_dir: join_paths(get_option('mandir'), 'man8'))
+    endforeach
+else
+    warning('Program "pandoc" not found!  Cannot generate documentation/man pages')
+endif
```
eli-schwartz, line 26: ... and then you can skip this warning, because people who absolutely want the docs can configure it with enabled, people who are okay gracefully degrading can configure it with auto, and people who absolutely don't want a pandoc dependency can disable it and not get a big yellow warning.

eli-schwartz (changes requested):

.gitignore
slyon: Thanks!

Makefile
```diff
@@ -86,8 +86,8 @@ clean:
 	find . | grep -E "(__pycache__|\.pyc)" | xargs rm -rf
 
 check: default linting
-	tests/cli.py
-	LD_LIBRARY_PATH=. $(NOSETESTS3) -v --with-coverage
+	PYTHONPATH=. LD_LIBRARY_PATH=. tests/cli.py
```
schopin-pro, line 89: praise: Nice, this is indeed much cleaner!

README.md
```diff
@@ -14,6 +14,15 @@ An overview of the architecture can be found at [netplan.io/design](https://netp
...
 The full documentation for netplan is available in the [doc/netplan.md file](../master/doc/netplan.md)
 
+# Build using Meson
+
+Steps to build netplan using the [Meson](https://mesonbuild.com) build system inside the `build/` directory:
+
+* meson setup build --prefix=/usr -Db_coverage=true
+* meson compile -C build
+* meson test -C build --verbose [TEST_NAME]
+* meson install -C build --destdir ../tmproot
```
schopin-pro, line 24: question: Do we actually want the user to install binaries that were compiled with coverage support?

schopin-pro (approved):
I'm not really familiar with Meson, so I'm discovering it here, but this all looks good, I only have a couple of minor questions/suggestions that can be skipped if in a hurry.

README.md
slyon: No, absolutely not. This documentation is mostly given for development purposes. But you're right in that it can easily be mistaken. Let me put that `-Db_coverage=true` into square brackets to make people actually think about it (unable to blindly apply that command).

src/meson.build
```diff
@@ -0,0 +1,44 @@
...
+    'sriov.c',
+    'types.c',
+    'util.c',
+    'validation.c')
+
+libnetplan = library(
+    'netplan',
+    sources,
+    gnu_symbol_visibility: 'hidden',
+    link_args: ['-T', '../abicompat.lds'],
```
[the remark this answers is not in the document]
slyon: Thanks!

This doesn't really work, as it will fail with `src/meson.build:16:0: ERROR: File -T does not exist.`, so I need to keep it as `link_args`, but I guess I could use a `linker_script = meson.project_source_root() / 'abicompat.lds'` variable to use both in link_args and link_depends, to make meson aware of any changes relevant to the linker.

tests/cli.py
```diff
@@ -54,7 +52,8 @@ def test_command_help(self):
         self.assertIn(b'--root-dir', out)
 
     def test_no_command(self):
-        os.environ['NETPLAN_GENERATE_PATH'] = os.path.join(rootdir, 'generate')
+        if not os.environ.get('NETPLAN_GENERATE_PATH'):
+            os.environ['NETPLAN_GENERATE_PATH'] = os.path.join(rootdir, 'generate')
```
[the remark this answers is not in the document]
slyon: ACK. That'd be much cleaner!

src/meson.build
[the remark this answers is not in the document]
eli-schwartz: That's what I mean? You pushed a revision to this PR with the following diff:
```diff
 gnu_symbol_visibility: 'hidden',
-link_depends: '../abicompat.lds',
+link_args: ['-T', '../abicompat.lds'],
 dependencies: [glib, gio, yaml, uuid],
```

slyon:
Thank you for all your comments! I'm going to merge this once all the tests passed.

src/meson.build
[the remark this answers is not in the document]
slyon: Oh, indeed! Giving it a 2nd read, I now understand what you meant in the first place. This is as implemented and merged.
So, thank you very much for your input!

PR #272, opened 2022-04-15
Adding VRF and VXLAN support for systemd-networkd (LP: #1764716, LP: #1773522)

netdever:
## Description

Adding VRF and VXLAN support for systemd-networkd.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. [LP#1764716](https://pad.lv/1764716), [LP#1773522](https://pad.lv/1773522)

### DIFF
--- a/doc/netplan.md
+++ b/doc/netplan.md
@@ -957,6 +957,16 @@ wpasupplicant installed if you let the ``networkd`` renderer handle wifi.
      ``rfkill_release`` or ``tcp`` (NetworkManager only). Or the exclusive
      ``default`` flag (the default).
 
+``regulatory-domain`` (scalar) – since **0.105**
+
+:    This can be used to define the radio's regulatory domain, to make use of
+     additional WiFi channels outside the "world domain". Takes an ISO /
+     IEC 3166 country code (like `GB`) or `00` to reset to the "world domain".
+     See [wireless-regdb](https://git.kernel.org/pub/scm/linux/kernel/git/sforshee/wireless-regdb.git/tree/db.txt) for available values.
+
+     **Requires dependency: iw**, if it is to be used outside the `networkd`
+     (wpa_supplicant) backend.
+
 ## Properties for device type ``bridges:``
 
 ``interfaces`` (sequence of scalars)
--- a/src/abi.h
+++ b/src/abi.h
@@ -357,4 +357,7 @@ struct netplan_net_definition {
 
     /* netplan-feature: infiniband */
     NetplanInfinibandMode ib_mode; /* IPoIB */
+
+    /* netplan-feature: regdom */
+    char* regulatory_domain;
 };
--- a/src/netplan.c
+++ b/src/netplan.c
@@ -780,6 +780,8 @@ _serialize_yaml(
         YAML_SEQUENCE_CLOSE(event, emitter);
     }
 
+    YAML_STRING(def, event, emitter, "regulatory-domain", def->regulatory_domain);
+
     if (def->optional_addresses) {
         YAML_SCALAR_PLAIN(event, emitter, "optional-addresses");
         YAML_SEQUENCE_OPEN(event, emitter);
--- a/abi-compat/jammy_0.104.xml
+++ b/abi-compat/jammy_0.104.xml
@@ -348,7 +348,7 @@
       </data-member>
     </class-decl>
     <typedef-decl name='NetplanBackendSettings' type-id='type-id-62' filepath='../src/abi.h' line='150' column='1' id='type-id-61'/>
-    <class-decl name='netplan_net_definition' size-in-bits='8832' is-struct='yes' visibility='default' filepath='../src/abi.h' line='155' column='1' id='type-id-27'>
+    <class-decl name='netplan_net_definition' size-in-bits='8896' is-struct='yes' visibility='default' filepath='../src/abi.h' line='176' column='1' id='type-id-27'>
       <data-member access='public' layout-offset-in-bits='0'>
         <var-decl name='type' type-id='type-id-13' visibility='default' filepath='../src/abi.h' line='156' column='1'/>
       </data-member>
@@ -558,6 +558,9 @@
       </data-member>
       <data-member access='public' layout-offset-in-bits='8768'>
         <var-decl name='sriov_delay_virtual_functions_rebind' type-id='type-id-39' visibility='default' filepath='../src/abi.h' line='329' column='1'/>
+      </data-member>
+      <data-member access='public' layout-offset-in-bits='8832'>
+        <var-decl name='regulatory_domain' type-id='type-id-40' visibility='default' filepath='../src/abi.h' line='353' column='1'/>
       </data-member>
     </class-decl>
     <class-decl name='__anonymous_struct__' size-in-bits='64' is-struct='yes' is-anonymous='yes' visibility='default' filepath='../src/abi.h' line='188' column='1' id='type-id-67'>
--- a/examples/wireless.yaml
+++ b/examples/wireless.yaml
@@ -3,6 +3,7 @@ network:
   renderer: networkd
   wifis:
     wlp2s0b1:
+      regulatory-domain: "GB"
       dhcp4: no
       dhcp6: no
       addresses: [192.168.0.21/24]
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -156,13 +156,13 @@ class NetplanApply(utils.NetplanCommand):
         if not restart_nm and old_files_nm:
             restart_nm = True
 
+        # Running 'systemctl daemon-reload' will re-run the netplan systemd generator,
+        # so let's make sure we only run it iff we're willing to run 'netplan generate'
+        if run_generate:
+            utils.systemctl_daemon_reload()
         # stop backends
         if restart_networkd:
             logging.debug('netplan generated networkd configuration changed, reloading networkd')
-            # Running 'systemctl daemon-reload' will re-run the netplan systemd generator,
-            # so let's make sure we only run it iff we're willing to run 'netplan generate'
-            if run_generate:
-                utils.systemctl_daemon_reload()
             # Clean up any old netplan related OVS ports/bonds/bridges, if applicable
             NetplanApply.process_ovs_cleanup(config_manager, old_files_ovs, restart_ovs, exit_on_error)
             wpa_services = ['netplan-wpa-*.service']
@@ -256,6 +256,9 @@ class NetplanApply(utils.NetplanCommand):
         # apply any SR-IOV related changes, if applicable
         NetplanApply.process_sriov_config(config_manager, exit_on_error)
 
+        # (re)set global regulatory domain
+        if os.path.exists('/run/systemd/system/netplan-regdom.service'):
+            utils.systemctl('start', ['netplan-regdom.service'])
         # (re)start backends
         if restart_networkd:
             netplan_wpa = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-wpa-*.service')]
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -298,6 +298,32 @@ write_link_file(const NetplanNetDefinition* def, const char* rootdir, const char
     umask(orig_umask);
 }
 
+static gboolean
+write_regdom(const NetplanNetDefinition* def, const char* rootdir, GError** error)
+{
+    g_assert(def->regulatory_domain);
+    g_autofree gchar* id_escaped = NULL;
+    g_autofree char* link = g_strjoin(NULL, rootdir ?: "", "/run/systemd/system/network.target.wants/netplan-regdom.service", NULL);
+    g_autofree char* path = g_strjoin(NULL, "/run/systemd/system/netplan-regdom.service", NULL);
+
+    GString* s = g_string_new("[Unit]\n");
+    g_string_append(s, "Description=Netplan regulatory-domain configuration\n");
+    g_string_append(s, "After=network.target\n");
+    g_string_append(s, "ConditionFileIsExecutable="SBINDIR"/iw\n");
+    g_string_append(s, "\n[Service]\nType=oneshot\n");
+    g_string_append_printf(s, "ExecStart="SBINDIR"/iw reg set %s\n", def->regulatory_domain);
+
+    g_string_free_to_file(s, rootdir, path, NULL);
+    safe_mkdir_p_dir(link);
+    if (symlink(path, link) < 0 && errno != EEXIST) {
+        // LCOV_EXCL_START
+        g_set_error(error, G_FILE_ERROR, G_FILE_ERROR_FAILED, "failed to create enablement symlink: %m\n");
+        return FALSE;
+        // LCOV_EXCL_STOP
+    }
+    return TRUE;
+}
+
 
 static gboolean
 interval_has_suffix(const char* param) {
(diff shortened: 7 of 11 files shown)


src/parse-nm.c
```diff
@@ -55,6 +55,14 @@ type_from_str(const char* type_str)
...
         return NETPLAN_DEF_TYPE_TUNNEL;
     */
+    /* TODO: Vxlans are not yet fully supported by the keyfile parser
+    else if (!g_strcmp0(type_str, "vxlan"))
+        return NETPLAN_DEF_TYPE_VXLAN;
+    */
+    /* TODO: Vrfs are not yet fully supported by the keyfile parser
+    else if (!g_strcmp0(type_str, "vrf"))
+        return NETPLAN_DEF_TYPE_VRF;
+    */
```
Conan-Kudo, line 65: I'm pretty sure this is supported in the keyfile format these days... @thom311, @lkundrak, @tyll, do any of you know?

Conan-Kudo (changes requested):

src/parse-nm.c
thom311: NetworkManager's keyfile format supports (necessarily) everything that NetworkManager supports.

But I think the reference to "the keyfile parser" here is `src/parse-nm.c` from netplan itself. This branch doesn't seem to implement the NetworkManager backend (which is a fine choice).

For the record, NetworkManager supports vxlan and vrfs.

netdever: That's correct - I intentionally did not touch the nm parsers in this pr. Thanks!

Conan-Kudo: @thom311 Is there any documentation of this stuff on networkmanager.dev? I couldn't find it specifically.

@netdever While I'm not part of the project, I would personally prefer to see things like this being implemented for both backends.

thom311: in NM, you always create "connection profiles". So, to create a vxlan/vrf configuration, you also create appropriate profiles.  The fields of a profile are documented in `man nm-settings-{dbus,nmcli,keyfile}` (which is also on the website). But there is no direct documentation how to create a vxlan/vrf configuration.

src/abi.h
```diff
@@ -327,4 +327,43 @@ struct netplan_net_definition {
     /* netplan-feature: eswitch-mode */
     char* embedded_switch_mode;
     gboolean sriov_delay_virtual_functions_rebind;
+
+    /* vxlan */
```
slyon, line 331:
```suggestion
    /* vxlan */
    /* netplan-feature: vxlan */
```

Please let's add a feature flag here for the new VXLAN feature. Also, maybe move the VXLAN fields below the new VRF fields, so they are listed together with the `vxlan_params` and not intertwined.

src/abi.h
```diff
@@ -327,4 +327,43 @@ struct netplan_net_definition {
...
+    guint vxlan_vni;
+    NetplanNetDefinition* vxlan_link;
+    char* vxlan_bridge;
+    GArray* vxlans;
+    gboolean neigh_suppress;
+
+    /* vrf */
```
slyon, line 338:
```suggestion
    /* vrf */
    /* netplan-feature: vrf */
```

Let's add a feature-flag for VRF, too

src/abi.h
```diff
@@ -327,4 +327,43 @@ struct netplan_net_definition {
...
+        gboolean remote_checksum_tx;
+        gboolean remote_checksum_rx;
+        gboolean group_policy_extension;
+        gboolean generic_protocol_extension;
+        guint destination_port;
+        char* port_range;
+        guint flow_label;
+        gboolean ip_do_not_fragment;
+        gboolean independent;
+    } vxlan_params;
```
slyon, line 368: Thank you for moving any new fields in this `netplan_net_definition` struct to the very end, this way we'll only append the ABI and not break it :-)

src/netplan.c
```diff
@@ -203,6 +203,66 @@ write_bond_params(yaml_event_t* event, yaml_emitter_t* emitter, const NetplanNet
...
+        YAML_BOOL_FALSE(def, event, emitter, "route-short-circuit", def->vxlan_params.route_short_circuit);
+        YAML_BOOL_FALSE(def, event, emitter, "udp-checksum", def->vxlan_params.udp_checksum);
+        YAML_BOOL_FALSE(def, event, emitter, "udp6-zero-checksum-tx", def->vxlan_params.udp6_zero_checksum_tx);
+        YAML_BOOL_FALSE(def, event, emitter, "udp6-zero-checksum-rx", def->vxlan_params.udp6_zero_checksum_rx);
+        YAML_BOOL_FALSE(def, event, emitter, "remote-checksum-tx", def->vxlan_params.remote_checksum_tx);
+        YAML_BOOL_FALSE(def, event, emitter, "remote-checksum-rx", def->vxlan_params.remote_checksum_rx);
+        YAML_BOOL_FALSE(def, event, emitter, "group-policy-extension", def->vxlan_params.group_policy_extension);
+        YAML_BOOL_FALSE(def, event, emitter, "generic-protocol-extension", def->vxlan_params.generic_protocol_extension);
+        YAML_UINT_0(def, event, emitter, "destination-port", def->vxlan_params.destination_port);
+        YAML_STRING(def, event, emitter, "port-range", def->vxlan_params.port_range);
```
slyon, line 256: In which way would we specify a "port-range" value? Can you give an example for this?

It seems like NetworkManager uses two separate settings for this:
* source-port-max
* source-port-min

Combining those into a single "port-range" sounds nice, but I'm wondering about the syntax?

https://networkmanager.dev/docs/api/latest/nm-settings-nmcli.html#id-1.2.9.4.43

src/netplan.c
```diff
@@ -203,6 +203,66 @@ write_bond_params(yaml_event_t* event, yaml_emitter_t* emitter, const NetplanNet
...
+        || def->vxlan_params.destination_port
+        || def->vxlan_params.port_range
+        || def->vxlan_params.flow_label
+        || def->vxlan_params.ip_do_not_fragment
+        || def->vxlan_params.independent) {
+        YAML_SCALAR_PLAIN(event, emitter, "parameters");
+        YAML_MAPPING_OPEN(event, emitter);
+        YAML_STRING(def, event, emitter, "remote", def->vxlan_params.remote);
+        YAML_STRING(def, event, emitter, "local", def->vxlan_params.local);
+        YAML_STRING(def, event, emitter, "group", def->vxlan_params.group);
```
slyon, line 238: IIUC Networkmanager uses a single setting to specify "remote" and "group", I wonder if we could do the same? Would the "remote" and "group" settings ever be used at the same time? Or can we deduct from the setting's value which one was meant by the user?

https://networkmanager.dev/docs/api/latest/nm-settings-nmcli.html#id-1.2.9.4.43

src/parse-nm.c
slyon: FTR: NetworkManager's VRF and VXLAN settings are listed here: https://networkmanager.dev/docs/api/latest/nm-settings-nmcli.html#id-1.2.9.4.43

But as Thomas stated, this code snippet in `src/parse-nm.c` is from netplan itself (allowing it to read a NM keyfile connection profile) and not necessarily needed for this PR.

slyon (commented):
Hi @netdever
Thank you very much for your contribution to netplan, this is highly appreciated, especially as the two LP bugs that you're addressing have been long-standing feature requests!

I did an initial review and mentioned some smaller issues and questions in inline comments.

I am now also thinking about bigger changes that might be needed:
* configmanager.py to go away (conflict with #255)
* NetplanTristate type should be added (conflict with #270)
* Linking of network interfaces: I wonder if we should rather define a VRF's child interfaces inside the vrf definition itself, using an `interfaces: [...]` setting (as suggested in the LP#1773522) instead of splitting the definition across parent and child, using the `vrf: ...` link:
```
network:
  version: 2
  renderer: networkd
  vrfs:
    vrf-test:
      interfaces:
        - bond0
        - br0
        - br-bond0
        - tun0
      table: auto # or a specific integer
```
* Similarly, I need to thinkg about the linking of VXLAN devices, as the `ethX.vxlans.names = [...]` seems to be a bit cumbersome. Is there a specific reason why you went for this nestring structure?
* Overall, I want to review all the new YAML settings in detail, to make sure we implement a simple and easily comprehensible wording. There is nothing to change about this right now... for the current stage we can just go with systemd-networkd's naming scheme and we can easily change/adopt that, once agreed upon (but before we merge the PR :))
* NetworkManager also supports (most) VRF/VXLAN settings that are implemented in this PR for the systemd-networkd backend renderer. With all the parsing code already in place, it should be relatively easy to implement those in `src/nm.c`. Would you be interested in doing that? But it could also be done as part of a separate PR, in which case we should at least log a warning from the NetworkManager backend renderer, that those settings are not currently supported on that backend.
* Finally, I'd really love to see some actuall integration tests (in `tests/integration/...`) that make use of those new VRF and VXLAN features and check the fully configured system at runtime, to make sure the settings are actually applied. Is that something you could add do your PR? The tests are being run as `autopkgtests`, as can be seen (and reproduced locally) in ".github/workflows/autopkgtest.yml`.

src/abi.h
netdever: No problem!

src/abi.h
netdever: Done.

src/abi.h
netdever: Done.

doc/netplan.md
```diff
@@ -1377,6 +1391,111 @@ Example:
...
+
+     ``group`` (scalar)
+     :    Configures VXLAN multicast group IP address.
+
+     ``tos`` (scalar)
+     :    The Type Of Service byte value for a vxlan interface.
+
+     ``ttl`` (scalar)
+     :    A fixed Time To Live N on Virtual eXtensible Local Area Network packets.
+          Takes a number in the range 0…255.
```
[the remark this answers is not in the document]
netdever: Done.

doc/netplan.md
```diff
@@ -1377,6 +1391,111 @@ Example:
...
+     :    Takes a boolean. When true, sending zero checksums in VXLAN/IPv6 is turned on.
+
+     ``udp6-zero-checksum-rx`` (scalar)
+     :    Takes a boolean. When true, receiving zero checksums in VXLAN/IPv6 is turned on.
+
+     ``remote-checksum-tx`` (scalar)
+     :    Takes a boolean. When true, remote transmit checksum offload of VXLAN is turned on.
+
+     ``remote-checksum-rx`` (scalar)
+     :    Takes a boolean. When true, remote receive checksum offload in VXLAN is turned on.
```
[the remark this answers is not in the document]
netdever: I gave this a shot but hit some snags. I'm not sure optional-addresses behaves the same way, but I could be missing something.

netplan/configmanager.py
```diff
@@ -108,6 +112,14 @@ def tunnels(self):
...
         return self.network['vlans']
 
+    @property
+    def vxlans(self):
+        return self.network['vxlans']
+
+    @property
+    def vrfs(self):
+        return self.network['vrfs']
+
```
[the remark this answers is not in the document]
netdever: Fingers crossed this gets merged first :)

src/abi.h
```diff
@@ -327,4 +327,43 @@ struct netplan_net_definition {
...
+        guint max_fdb_entries;
+        gboolean reduce_arp_proxy;
+        gboolean l2_miss_notification;
+        gboolean l3_miss_notification;
+        gboolean route_short_circuit;
+        gboolean udp_checksum;
+        gboolean udp6_zero_checksum_tx;
+        gboolean udp6_zero_checksum_rx;
+        gboolean remote_checksum_tx;
+        gboolean remote_checksum_rx;
```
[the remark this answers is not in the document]
netdever: Had some trouble trying to bundle these myself, so leaving as is for now. Any help would be appreciated if we feel it's necessary to make this change.

src/netplan.c
netdever: In my testing, I could only specify an integer for `PortRange=` using systemd-networkd. Obviously that's not a "range." Documentation is lacking for this parameter in systemd-networkd.

src/abi.h
```diff
@@ -327,4 +327,45 @@ struct netplan_net_definition {
...
+    /* netplan-feature: vrf */
+    char* vrf;
+    guint vrf_table;
+    NetplanNetDefinition* vrf_link;
+
+    /* vxlan */
+    /* netplan-feature: vxlan */
+    guint vxlan_vni;
+    NetplanNetDefinition* vxlan_link;
+    char* vxlan_bridge;
```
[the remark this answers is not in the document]
netdever: Good call, I overlooked that. I've updated the PR to use the existing definition.

src/netplan.c
slyon: Ineed, Group= will always be a multicast IP address (i.e. from a reserved IPv4/6 range), so we can implicitly derive this information, therefore I'd like to reduce those two fields to just a single `remote` field, that is than translated to Remote= or Group= for systemd-networkd depending on if it is a multicast IP or not.

doc/netplan.md
```diff
@@ -1377,6 +1390,111 @@ Example:
...
+          the normal range of local UDP ports is used.
+
+     ``flow-label`` (scalar)
+     :    Specifies the flow label to use in outgoing packets. The valid range is 0-1048575.
+
+     ``ip-do-not-fragment`` (scalar)
+     :    Allows setting the IPv4 Do not Fragment (DF) bit in outgoing packets. Takes a boolean value.
+          When unset, the kernel's default will be used.
+
+     ``independent`` (scalar)
```
slyon, line 1483: I'd like to avoid this setting, as it is very backend (systemd-networkd) specific. We already have this information if NetplanNetDefinition->vxlan_link is set and can set the value implicitly. No need to expose it to the user via YAML

netdever: Sure - I removed this from the PR.

src/netplan.c
netdever: OK, I was able to get this to work as described. There isn't validation for it yet, though.

netdever:
- Linking of VRFs now looks more like the example in the linked LP bug.
- Linking of VXLANs to source interfaces no longer requires the `names` subvar.
- VXLAN `source-port-range` now works as suggested (lacks validation)

src/abi.h
```diff
@@ -327,4 +327,41 @@ struct netplan_net_definition {
...
+    /* vrf */
+    /* netplan-feature: vrf */
+    char* vrf;
+    guint vrf_table;
+
+    /* vxlan */
+    /* netplan-feature: vxlan */
+    guint vxlan_vni;
+    GArray* vxlans;
+    gboolean neigh_suppress;
```
slyon, line 340: This seems to be a rather generic `[Bridge]` setting, not necessarily related to VXLANs.. Is it required for VXLAN? If so, we should implement it as a generic bridge feature (and probably use a tristate type, too).

Otherwise, I'd like to just drop it from this PR, as it seems unrelated. Especially, as there is currently no implementation in any backend (neither `networkd.c` nor `nm.c` for this `neigh_suppress` variable).

netdever: Neighbor suppression (arp-nd-suppression) is required for VXLAN configuration. It is disabled by default, but many implementations, specifically those utilizing EVPN, will need it to be enabled. In systemd-networkd, it looks like this:

```bash
root@host1:~# cat /run/systemd/network/10-netplan-vxlan10.network 
[Match]
Name=vxlan10

[Link]
MTUBytes=8950

[Network]
LinkLocalAddressing=no
IPv6AcceptRA=no
ConfigureWithoutCarrier=yes
Bridge=bridge10

[Bridge]
NeighborSuppression=True
```

netdever: The command to enable this feature is ```bridge link set dev vxlan10 neigh_suppress on```

slyon: Thank you for your comments about this. Being a `[Bridge]` setting in sd-networkd, I still wonder if this is a setting that should be set on the VXLAN device itself, or rather on a bridge device that is a parent interface to the vxlan?

netdever: It is a VXLAN setting (ie. you cannot enable it on the master Bridge interface). It may be confusing for netplan to have it as a bridge setting, especially since enslaving a VXLAN to a Bridge is not technically a requirement.

```bash
root@host1:~# bridge link set dev bridge10 neigh_suppress on
RTNETLINK answers: Operation not supported
```

slyon:
This PR got closed by accident after rebasing on top of the current `main` branch and then pushing the wrong branch to `netdever/main`. I do not have write access to netdever's branch anymore. So I created a copy of this in #285

PR #274, opened 2022-04-29
Allow key configuration for [ip6]gre[tap] tunnels (LP: #1966476)

slyon:
## Description
The following example should configure a GRE tunnel between 172.16.4.2 and 172.16.1.2, with key 33490175. The effect should be like manually configured with: sudo ip tunnel add test1 mode gre local 172.16.4.2 remote 172.16.1.2 key 33490175.

However, the following Netplan configuration fails under Ubuntu 22.04 (latest development version, March 25, 2022):
```
network:
  version: 2
  renderer: networkd
  tunnels:
    gre1-255-255:
      mode: gre
      ttl: 255
      local: 172.16.4.2
      remote: 172.16.1.2
      key: 33490175
      addresses:
        - 172.20.82.94/30
```
Result of "sudo netplan generate": Error in network definition: gre1-255-255: 'input-key' is not required for this tunnel type

Some investigation into the sources of Netplan (src/validation.c -> https://github.com/canonical/netplan/blob/main/src/validation.c):
It seems that the cases NETPLAN_TUNNEL_MODE_GRE and NETPLAN_TUNNEL_MODE_IP6GRE are missing for systemd-networkd. According to https://github.com/systemd/systemd/issues/12144, systemd-networkd should support configuration with keys since ca. 2 years.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1966476

### DIFF
--- a/src/validation.c
+++ b/src/validation.c
@@ -244,6 +244,10 @@ validate_tunnel_backend_rules(const NetplanParser* npp, NetplanNetDefinition* nd
                 case NETPLAN_TUNNEL_MODE_VTI:
                 case NETPLAN_TUNNEL_MODE_VTI6:
                 case NETPLAN_TUNNEL_MODE_WIREGUARD:
+                case NETPLAN_TUNNEL_MODE_GRE:
+                case NETPLAN_TUNNEL_MODE_IP6GRE:
+                case NETPLAN_TUNNEL_MODE_GRETAP:
+                case NETPLAN_TUNNEL_MODE_IP6GRETAP:
                     break;
 
                 /* TODO: Remove this exception and fix ISATAP handling with the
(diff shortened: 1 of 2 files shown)


dreibh:
The solution is working successfully for configuring a GRE tunnel with key.

slyon:
Thank you very much for the feedback/confirmation. I still need to find some time to polish up this PR (maybe add an integration tests) and get it landed.

slyon:
I cleaned this PR up, added some integration tests (autopkgtest) and rebased on top of the current `origin/main` branch.
Thans again for testing @dreibh! Would you mind doing a code review as well?

slyon:
Thanks for the review! I rebased on top of `origin/master` (and re-ordered the git history, to satisfy Simon's inner Git neat freak). Merging.

PR #275, opened 2022-05-03
libnetplan: String helpers for the new API design + filepath transition

schopin-pro:
## Description

This PR implements some infrastructure to easily implement new APIs according to our guidelines, especially regarding strings.
To demonstrate it, the last commit transition `netdef->filename` into `netdef->filepath`, implementing the new getter accordingly.

## Checklist


src/util.c
```diff
@@ -600,3 +601,16 @@ mark_data_as_dirty(NetplanParser* npp, void* data_ptr)
         npp->current.netdef->_private->dirty_fields = g_hash_table_new(g_direct_hash, g_direct_equal);
     g_hash_table_insert(npp->current.netdef->_private->dirty_fields, data_ptr, data_ptr);
 }
+
+ssize_t
+netplan_copy_string(const char* input, char* out_buffer, size_t out_size)
```
slyon, line 606: The function makes perfect sense.
praise: re-use of existing stpncpy is nice!

A small comment describing the input parameters and return value would be helpful here for future reference.

include/netplan.h
```diff
@@ -72,7 +75,12 @@ netplan_netdef_write_yaml(
...
         GError** error);
 
-NETPLAN_PUBLIC const char *
+NETPLAN_PUBLIC ssize_t
+netplan_netdef_get_filepath(const NetplanNetDefinition* netdef, char* out_buffer, size_t out_buffer_size);
+
+/********** Old API below this ***********/
+
+NETPLAN_DEPRECATED NETPLAN_PUBLIC const char *
 netplan_netdef_get_filename(const NetplanNetDefinition* netdef);
```
slyon, line 84: praise: you nicely handled the backwards compatibility!

slyon (approved):
Thank you very much! LGTM.

I left some minor inline comments that we should fix while rebasing the branch, before we merge it. Very nice handling of the deprecation warning and the python buffer allocation that is automatically increasing in size!

src/util.c
schopin-pro: This should be fixed in the latest revision.

src/util.c
```diff
@@ -705,3 +706,29 @@ mark_data_as_dirty(NetplanParser* npp, void* data_ptr)
         npp->current.netdef->_private->dirty_fields = g_hash_table_new(g_direct_hash, g_direct_equal);
     g_hash_table_insert(npp->current.netdef->_private->dirty_fields, data_ptr, data_ptr);
 }
+
+/**
+ * Copy a NUL-terminated string into a sized buffer, and returns the size of
+ * the copied string, including the final NUL characted. If the buffer is too
```
slyon, line 712:
```suggestion
 * the copied string, including the final NUL character. If the buffer is too
```
nitpick (non-blocking): typo

slyon:
Thank you very much. LGTM.

Feel free to merge it!

src/util.c
schopin-pro: Fixed, thanks!

PR #276, opened 2022-05-03
Improve NetworkManager's device management/ignore logic, using udev matching rules (LP: #1951653)

slyon:
## Description
By default, `systemd-networkd` controls only interfaces that it has explicitly been configured for (via a sd-networkd `.network` file); it considers other interfaces as "unmanaged" and ignores them. NetworkManager OTOH tries to claim control of any interface that is UP, which might lead to conflicts if multiple networking daemons are used at the same time.

NetworkManager has different ways to control/avoid this behavior:
* `NM_UNMANAGED=1` udev environment, the weakest condition
* NetworkManager.conf `[keyfile].unmanaged-devices=` setting (affected by some unexpected behavior: LP#1615044)
* NetworkManager.conf `[device].managed=` setting, the strongest condition, overruling all other settings, created a runtime in `/run/NetworkManager/devices/*`

The stronger rules via NetworkManager.conf are unfortunately a bit limited in their matching logic, as multiple checks can only be ORed (e.g. we cannot check for: `interface-name=wl* AND driver=(virt* OR iwlwifi)`). udev OTOH has a very powerful matching logic (but unfortunately cannot match on the interface type: ethernet/wifi/bridge/...).

So we're reimplementing/refactoring the flawed NetworkManager "unmanaged vs managed" logic in netplan, by having an explicit **allow-list** of interfaces that NM is supposed to control (as some udev rules shipped system-wide would otherwise disable certain interfaces in containers and other special conditions, even if configured via netplan, c.f. `/lib/udev/rules.d/85-nm-unmanaged.rules`) in addition to an explicit **deny-list** of interfaces to ignore (e.g. because they are to be managed by sd-networkd) as udev rules in `/run/udev/rules.d/90-netplan.rules` and keeping the special-case of matching a whole class/type of interfaces (e.g. ethernets/wifis/bridges/...) in `/run/NetworkManager/conf.d/netplan.conf` as a NetworkManager.conf `[device*].managed=` rule.
Furthermore, we're applying this matching logic whenever ANY interface (in netplan YAML) is supposed to be controlled by NetworkManager (that is in contrast to only applying it when NetworkManager is selected as the global renderer).

It is worth mentioning that this "managed vs unmanaged" matching logic is slightly different from netplan's usual matching logic (e.g. for configuring the interfaces), as we might ignore an interface based on it's old AND new/renamed interface name, or it's old and new/changed MAC address, while we cannot do this for the normal netplan matching.

Commits:
* generate: ignore 10-globally-managed-devices.conf if any NM config is given in netplan
* cli: re-apply udev rules (NM_UNMANAGED) if needed
  * Also, make sure we clear any runtime state that NetworkManager left in /run/NetworkManager/devices/* as this could overwrite our udev rules.
* nm: extend type_str() to return passthrough types
* nm: tests: improve NM manage/ignore logic, using udev matching rules
* tests:base: cleanup udev quirks
  * We have an explicit NM-managed-devices allow-list, now. So we don't need those quirks anymore.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1951653

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -128,7 +128,9 @@ type_str(const NetplanNetDefinition* def)
             return "ip-tunnel";
         case NETPLAN_DEF_TYPE_NM:
             /* needs to be overriden by passthrough "connection.type" setting */
-            return NULL;
+            g_assert(def->backend_settings.nm.passthrough);
+            GData *passthrough = def->backend_settings.nm.passthrough;
+            return g_datalist_get_data(&passthrough, "connection.type");
         // LCOV_EXCL_START
         default:
             g_assert_not_reached();
@@ -595,7 +597,7 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
     }
 
     nm_type = type_str(def);
-    if (nm_type)
+    if (nm_type && def->type != NETPLAN_DEF_TYPE_NM)
         g_key_file_set_string(kf, "connection", "type", nm_type);
 
     if (ap && ap->backend_settings.nm.uuid)
--- a/src/nm.c
+++ b/src/nm.c
@@ -26,58 +26,15 @@
 #include <glib/gprintf.h>
 #include <uuid.h>
 
+#include "names.h"
 #include "netplan.h"
 #include "nm.h"
 #include "parse.h"
 #include "parse-globals.h"
+#include "parse-nm.h"
 #include "util.h"
 #include "util-internal.h"
 #include "validation.h"
-#include "parse-nm.h"
-
-GString* udev_rules;
-
-/**
- * Append NM device specifier of @def to @s.
- */
-static void
-g_string_append_netdef_match(GString* s, const NetplanNetDefinition* def)
-{
-    g_assert(!def->match.driver || def->set_name);
-    if (def->match.mac || def->match.original_name || def->set_name || def->type >= NETPLAN_DEF_TYPE_VIRTUAL) {
-        if (def->match.mac) {
-            g_string_append_printf(s, "mac:%s,", def->match.mac);
-        }
-        /* MAC could change, e.g. for bond slaves. Ignore by interface-name as well */
-        if (def->match.original_name || def->set_name || def->type >= NETPLAN_DEF_TYPE_VIRTUAL) {
-            /* we always have the renamed name here */
-            g_string_append_printf(s, "interface-name:%s,",
-                    (def->type >= NETPLAN_DEF_TYPE_VIRTUAL) ? def->id
-                                            : (def->set_name ?: def->match.original_name));
-        }
-    } else {
-        /* no matches → match all devices of that type */
-        switch (def->type) {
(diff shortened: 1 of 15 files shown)


src/nm.c
```diff
@@ -962,49 +919,108 @@ netplan_netdef_write_nm(
...
-    s = g_string_new("[keyfile]\n# devices managed by networkd\nunmanaged-devices+=");
-    len = s->len;
-    g_list_foreach(np_state->netdefs_ordered, nd_append_non_nm_ids, s);
-    if (s->len > len)
-        g_string_free_to_file(s, rootdir, "run/NetworkManager/conf.d/netplan.conf", NULL);
+     * auto-connect and interferes.
+     * Also, mark all devices managed by us explicitly, so it won't get in
+     * conflict with the system's udev rules that might ignore some devices
+     * in containers via usr/lib/udev/rules.d/85-nm-unmanaged-devices.rules */
+    GList* iter = np_state->netdefs_ordered;
```
schopin-pro, line 939: praise: great work on the whole function body, very readable and well commented despite the complex logic!

src/nm.c
```diff
@@ -962,49 +919,108 @@ netplan_netdef_write_nm(
...
+            if (nd->set_mac) {
+                tmp = g_string_new(nd->set_mac);
+                g_string_append_printf(udev_rules, "%s ATTR{address}==\"%s\",%s", prefix, g_string_ascii_down(tmp)->str, suffix);
+                g_string_free(tmp, TRUE);
+            }
+            /* Finally, add a full match, using all rules & globs available
+             * from the "match" stanza (e.g. original_name/mac/drivers)
+             * This will match the "old" interface (i.e. original MAC and/or
+             * interface name) if it got changed */
+            if (nd->has_match && (nd->match.original_name || nd->match.mac || nd->match.driver)) {
```
schopin-pro, line 979: nitpick (non-blocking): this block could perhaps have been split off into its own function to make it a bit more bitesize ?

schopin-pro (approved):
LGTM!
Only one small optional stylistic nitpick, feel free to merge if you disagree!

Small question though, where does that last commit come from? Has this bug surfaced due to the NM changes?

Conan-Kudo:
@thom311 is this reasonable?

thom311:
> @thom311 is this reasonable?

yes. lgtm.

I am surprised that netplan aims to support a mixed system with networkd and NetworkManager. But sure.

slyon:
> yes. lgtm.

Thank you for your feedback @thom311! merging.

> I am surprised that netplan aims to support a mixed system with networkd and NetworkManager. But sure.

Yes. A combined setup can lead to funny situations, but we try our best to make it work.

src/nm.c
slyon: Ugh.. I was a bit quick with merging and didn't remember this suggestion. It makes sense to encapsulate this into its own function. I will probably work on this in a follow-up commit.

PR #279, opened 2022-05-18
nm: fix rendering of password for unknown/passthrough WPA types (LP: #1972800)

slyon:
## Description
The NetworkManager backend should not take a shortcut and skip rendering the password, in case of an unknown WPA type, as that might be overwritten by a passthrough value.

## Checklist

- [ ] \(Optional\) Closes an open bug in Launchpad. LP#1972800

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -431,9 +431,6 @@ write_tunnel_params(const NetplanNetDefinition* def, GKeyFile *kf)
 static void
 write_dot1x_auth_parameters(const NetplanAuthenticationSettings* auth, GKeyFile *kf)
 {
-    if (auth->eap_method == NETPLAN_AUTH_EAP_NONE)
-        return;
-
     switch (auth->eap_method) {
         case NETPLAN_AUTH_EAP_TLS:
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato (approved):
LGTM, thanks!

src/nm.c
```diff
@@ -431,9 +431,6 @@ write_tunnel_params(const NetplanNetDefinition* def, GKeyFile *kf)
 static void
 write_dot1x_auth_parameters(const NetplanAuthenticationSettings* auth, GKeyFile *kf)
 {
-    if (auth->eap_method == NETPLAN_AUTH_EAP_NONE)
-        return;
-
```
schopin-pro, line 436: question: this is because this value is the default when the EAP method is unknown, right?

schopin-pro (approved):
LGTM, just had a small question :)

src/nm.c
slyon: exactly.

It is initialized like that (`auth->eap_method = NETPLAN_AUTH_EAP_NONE;`) in types.c:reset_auth_settings()

PR #280, opened 2022-05-25
manpages: fix long options rendered with an en-dash instead of two dash

ogayot:
## Description

Hello,

In generated manpages, the long options look as if they were prefixed with a single-dash "-", as in the following example:

```
 netplan [–debug] generate [–root-dir ROOT_DIR] [–mapping MAPPING]
```

In reality, the double-dash notation was replaced by an en-dash (U+2013) when generating the manpage.

This is caused by the "smart" extension that pandoc enables by default when reading markdown.

Fixed by preventing pandoc from enabling the smart extension when generating manpages. This is only possible when we explicitly specify the input format.

Disabling the smart extension can have a slight impact on how other things like quotes are rendered, but that should not be much of an inconvenience in comparison to the drawbacks that the extension brings.

Thanks,

## Checklist

- [x] New/changed keys in YAML format are documented. _(no new or updated key)_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,9 +153,9 @@ install: default
 	pandoc -s --metadata title="Netplan reference" --toc -o $@ $<
 
 doc/netplan.5: doc/manpage-header.md doc/netplan.md doc/manpage-footer.md
-	pandoc -s -o $@ $^
+	pandoc -s -o $@ --from=markdown-smart $^
 
 %.8: %.md
-	pandoc -s -o $@ $^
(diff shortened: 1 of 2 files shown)


slyon:
Yes, this makes perfect sense. Thanks for spotting and fixing it right away!
LGTM, merging!

slyon (approved):
Yes, this makes perfect sense. Thanks for spotting and fixing it right away.

LGTM, merging!

PR #283, opened 2022-06-07
parse: accept 20 byte InfiniBand MAC addresses (LP: #1848471)

slyon:
## Description
parse: accept 20 byte InfiniBand MAC addresses

## Checklist

- [ ] \(Optional\) Closes an open bug in Launchpad. LP#1848471)


schopin-pro (commented):
I only have one holdout about this, which is the use of a generic string in the data model. However, since we already do it elsewhere you can disregard the comment entirely, I won't hold it against you ;-)

src/abi.h
```diff
@@ -348,4 +348,7 @@ struct netplan_net_definition {
     /* netplan-feature: eswitch-mode */
     char* embedded_switch_mode;
     gboolean sriov_delay_virtual_functions_rebind;
+
+    /* netplan-feature: infiniband */
+    char* ib_mode; /* IPoIB */
```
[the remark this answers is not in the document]
slyon: Yes, I was pondering about this, too. I will go ahead and just do it. Thanks!

slyon:
> I'm not in a position to review this - I don't know C and don't have access to Infiniband

Alright @sparkiegeek, I just wanted to keep you in the loop as this feature request originated from a MAAS context a while ago.

We already got @schopin-pro's review (thanks for that!) and I adopted the remarks accordingly. Let's try to get the new `infiniband-mode` stanza approved schema wise, otherwise this PR should be good to be merged!

schopin-pro (approved):
LGTM with the new changes :)

sparkiegeek (commented):
As noted - am an interested end-customer but not qualified to review :)

slyon:
> As noted - am an interested end-customer but not qualified to review :)

ACK. We're waiting for schema review of the new `infiniband-mode: datagram|connected` setting, which is already on Steve's list. Afterwards we are good to merge this.

doc/netplan.md
```diff
@@ -831,6 +831,14 @@ Example:
 
      **Requires feature: eswitch-mode**
 
+``infiniband-mode`` (scalar) – since **0.105**
+
+:    (InfiniBand devices only) Change the operational mode of a IPoIB device.
+     Possible values are ``datagram`` or ``connected``, if unspecified the
+     kernel's default configuration is used.
```
vorlonofportland, line 838: this is a run-on sentence, should say: "Possible values are ``datagram`` or ``connected``.  If unspecified [...]"

examples/infiniband.yaml
```diff
@@ -0,0 +1,9 @@
+network:
+  version: 2
+  renderer: networkd
+  ethernets:
+    ib0:
+      match:
+        macaddress: "11:22:33:44:55:66:77:88:99:00:11:22:33:44:55:66:77:88:99:00"
```
vorlonofportland, line 7: Is this the preferred format for representing infiniband mac addresses?  It doesn't seem very readable.  It does seem to be the format that NM uses, but are there alternatives in common use that we should consider?

vorlonofportland (changes requested):
Proposed schema additions look ok, with one question inline (and a minor editorial fix requested).

slyon:
> Proposed schema additions look ok, with one question inline (and a minor editorial fix requested).

Thanks, I've added the "schema ok" label. Fixed your documentation remark and commented on the MAC address format. This should be good for merging now, if no further remarks are added.

slyon:
Merging. I think all comments have been addressed. If we'd want to support an additional format in the future, we can still add this as a follow-up.

PR #284, opened 2022-06-30
cli:ip:leases: search for NM's internal DHCP client and dhclient leases (LP: #1979674)

slyon:
## Description
Check for NetworkManager leases in `var/lib/NetworkManager/internal-{UUID}-{INTERFACE}.lease` and fallback to `var/lib/NetworkManager/dhclient-{UUID}-{INTERFACE}.lease` if that does not exist. Dhclient is deprecated, but might still be used in some legacy setups.

Also, use some mocking (custom `MockCmd` class, as used elsewhere in our code) to actually run some unit tests on this code and increase the coverage.

The whole code in `ip.py` is pretty old and in bad shape; it should be refactored, but I don't have the time for that right now, just wanted to fix LP#1979674 as a first step.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad.: LP#1979674

### DIFF
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -122,10 +122,15 @@ class NetplanIpLeases(utils.NetplanCommand):
                 # we'll rely on open() throwing an error.
                 # This might happen if networkd doesn't use DHCP for the interface,
                 # for instance.
-                with open(os.path.join('/',
-                                       os.path.abspath(self.root_dir) if self.root_dir else "",
-                                       lease_pattern.format(interface=self.interface,
-                                                            lease_id=lease_id))) as f:
+                path = os.path.join('/',
+                                    os.path.abspath(self.root_dir) if self.root_dir else "",
+                                    lease_pattern.format(interface=self.interface,
+                                                         lease_id=lease_id))
+                # Fallback to 'dhclient' if no lease of NetworkManager's
+                # internal DHCP client is found
+                if not os.path.isfile(path):
+                    path = path.replace('NetworkManager/internal-', 'NetworkManager/dhclient-')
+                with open(path) as f:
                     for line in f.readlines():
                         print(line.rstrip())
             except Exception as e:
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -31,7 +31,7 @@ lease_path = {
         'method': 'ifindex',
     },
     'NetworkManager': {
-        'pattern': 'var/lib/NetworkManager/dhclient-{lease_id}-{interface}.lease',
+        'pattern': 'var/lib/NetworkManager/internal-{lease_id}-{interface}.lease',
         'method': 'nm_connection',
     },
 }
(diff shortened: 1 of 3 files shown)


netplan/cli/commands/ip.py
```diff
@@ -90,21 +90,19 @@ def lease_method_ifindex():
...
-            def lease_method_nm_connection():  # pragma: nocover (covered in autopkgtest)
+            def lease_method_nm_connection():
                 # FIXME: handle older versions of NM where 'nmcli dev show' doesn't exist
                 try:
-                    nmcli_dev_out = subprocess.Popen(['nmcli', 'dev', 'show', self.interface],
-                                                     env={'LC_ALL': 'C'},
-                                                     stdout=subprocess.PIPE)
-                    for line in nmcli_dev_out.stdout:
-                        line = line.decode('utf-8')
+                    nmcli_dev_out = subprocess.check_output(['nmcli', 'dev', 'show', self.interface],
```
ogayot, line 96: Not strictly related to the change from this PR but if you want (and if you're targetting a version of NM that has this option) you can use the following construct to simplify how the output is parsed:

```sh
nmcli --get-values GENERAL.CONNECTION dev show "$interface"
nmcli --get-values connection.uuid show id "$conn_id"
```

This is strictly up to you.

netplan/cli/commands/ip.py
```diff
@@ -90,21 +90,19 @@ def lease_method_ifindex():
...
+                    for line in nmcli_dev_out.splitlines():
                         if 'GENERAL.CONNECTION' in line:
                             conn_id = line.split(':')[1].rstrip().strip()
-                            nmcli_con_out = subprocess.Popen(['nmcli', 'con', 'show', 'id', conn_id],
-                                                             env={'LC_ALL': 'C'},
-                                                             stdout=subprocess.PIPE)
-                            for line in nmcli_con_out.stdout:
-                                line = line.decode('utf-8')
+                            nmcli_con_out = subprocess.check_output(['nmcli', 'con', 'show', 'id', conn_id],
+                                                                    env=dict(LC_ALL='C', PATH=os.environ.get('PATH')),
```
ogayot, line 103: `subprocess` can't deal with an environment having a None value, so I would suggest:
```diff
-os.environ.get('PATH')
+os.environ['PATH']
```
Unless you wanted to use `os.environ.get('PATH', os.defpath)` here ?
For sure it will also fail if PATH is not present in the environment, but the error should be less obscure, I think.

ogayot (approved):
LGTM!
I've added two minor comments. I think you should have a look at the `os.environ.get` thing at least to confirm that this is what you originally intended.

netplan/cli/commands/ip.py
slyon: Thank you, good spot! I opted to using `os.environ.get('PATH', os.defpath)` as this should give the least obscure errors (i.e. no errors about the environment, but about `nmcli` if it does not exist).

netplan/cli/commands/ip.py
slyon: This is a good suggestion, but I'll leave it open for a bigger refactoring of `ip.py`. For now, I just want to provide a minimal fix to solve the reported bug (LP#1979674).

PR #285, opened 2022-07-04
Add support for VRF devices (LP: #1773522)

slyon:
## Description
Adding support for Virtual Routing and Forwarding devices using the sd-networkd and NetworkManager backends.
Any VRF device is linked to a (new) routing table, specified via the `table` stanza and can define its child interfaces via `interfaces: [...]`. If any `routes:` or `routing-policy:` settings are given, those are implicitly connected to the VRF's routing table.

It builds upon #272 (squashed & rebased), extracted VRF bits.

Example:
```
network:
  version: 2
  [...]
  vrfs:
    vrf1005:
      table: 1005
      interfaces:
        - br1
        - br1005
      routes:  # uses "table: 1005" implicitly
        - to: default
          via: 10.10.10.4
      routing-policy:
        - from: 10.10.10.42
  bridges:
    br1:
      interfaces: [...]
    br1005:
      interfaces: [...]
```

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. [LP#1773522](https://pad.lv/1773522)

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -2023,6 +2023,29 @@ static const mapping_entry_handler bond_params_handlers[] = {
 };
 
 static gboolean
+handle_vrf_interfaces(NetplanParser* npp, yaml_node_t* node, const void* data, GError** error)
+{
+    /* all entries must refer to already defined IDs */
+    for (yaml_node_item_t *i = node->data.sequence.items.start; i < node->data.sequence.items.top; i++) {
+        yaml_node_t *entry = yaml_document_get_node(&npp->doc, *i);
+        NetplanNetDefinition *component;
+
+        assert_type(npp, entry, YAML_SCALAR_NODE);
+        component = g_hash_table_lookup(npp->parsed_defs, scalar(entry));
+        if (!component) {
+            add_missing_node(npp, entry);
+        } else {
+            if (component->vrf_link && component->vrf_link != npp->current.netdef)
+                return yaml_error(npp, node, error, "%s: interface '%s' is already assigned to vrf %s",
+                                  npp->current.netdef->id, scalar(entry), component->vrf_link->id);
+            component->vrf_link = npp->current.netdef;
+        }
+    }
+
+    return TRUE;
+}
+
+static gboolean
 handle_bonding(NetplanParser* npp, yaml_node_t* node, const char* key_prefix, const void* _, GError** error)
 {
     return process_mapping(npp, node, key_prefix, bond_params_handlers, NULL, error);
@@ -2546,6 +2569,16 @@ static const mapping_entry_handler vlan_def_handlers[] = {
     {NULL}
 };
 
+static const mapping_entry_handler vrf_def_handlers[] = {
+    COMMON_BACKEND_HANDLERS,
+    {"renderer", YAML_SCALAR_NODE, {.generic=handle_netdef_renderer}},
+    {"interfaces", YAML_SEQUENCE_NODE, {.generic=handle_vrf_interfaces}, NULL},
+    {"table", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(vrf_table)},
+    {"routes", YAML_SEQUENCE_NODE, {.generic=handle_routes}},
+    {"routing-policy", YAML_SEQUENCE_NODE, {.generic=handle_ip_rules}},
+    {NULL}
+};
+
 static const mapping_entry_handler modem_def_handlers[] = {
     COMMON_LINK_HANDLERS,
     COMMON_BACKEND_HANDLERS,
@@ -2763,6 +2796,7 @@ handle_network_type(NetplanParser* npp, yaml_node_t* node, const char* key_prefi
             case NETPLAN_DEF_TYPE_MODEM: handlers = modem_def_handlers; break;
             case NETPLAN_DEF_TYPE_TUNNEL: handlers = tunnel_def_handlers; break;
             case NETPLAN_DEF_TYPE_VLAN: handlers = vlan_def_handlers; break;
+            case NETPLAN_DEF_TYPE_VRF: handlers = vrf_def_handlers; break;
             case NETPLAN_DEF_TYPE_WIFI: handlers = wifi_def_handlers; break;
             case NETPLAN_DEF_TYPE_NM:
                 g_warning("netplan: %s: handling NetworkManager passthrough device, settings are not fully supported.", npp->current.netdef->id);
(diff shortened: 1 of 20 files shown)


Conan-Kudo:
Can we have NetworkManager supported in this PR (or shortly afterward)?

Conan-Kudo:
Awesome!

slyon:
The new integration tests (`routing.test_vrf_basic`) are finally passing in CI and the commit history should now be rather clean.

Thank you very much for the initial implementation @netdever for the initial implementation.
May I ask you to do a review of the VRF code that I added/adopted on top of yours?

Also, @Conan-Kudo seems to be interested, especially in the NetworkManager part.
May I ask you to do a code-review on this PR as well?

Conan-Kudo (approved):

src/validation.c
```diff
@@ -415,6 +420,51 @@ validate_sriov_rules(const NetplanParser* npp, NetplanNetDefinition* nd, GError*
...
+        NetplanNetDefinition *nd = value;
+        if (nd->type != NETPLAN_DEF_TYPE_VRF || !nd->routes)
+            continue;
+
+        /* Routes */
+        for (size_t i = 0; i < nd->routes->len; i++) {
+            NetplanIPRoute* r = g_array_index(nd->routes, NetplanIPRoute*, i);
+            if (r->table == nd->vrf_table)
+                continue;
+            else if (r->table != NETPLAN_ROUTE_TABLE_UNSPEC && r->table != nd->vrf_table) {
```
schopin-pro, line 441: question: can't we remove the later clause of the `&&` as already proven by the earlier `if` ?

src/validation.c
```diff
@@ -415,6 +420,51 @@ validate_sriov_rules(const NetplanParser* npp, NetplanNetDefinition* nd, GError*
...
+                g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT,
+                            "%s: VRF routes table mismatch (%d != %d)", nd->id, nd->vrf_table, r->table);
+                return FALSE;
+            } else {
+                r->table = nd->vrf_table;
```
schopin-pro, line 446: question: will the table be rendered out via `netplan get` ? Not sure, since it's not marked as dirty...

doc/netplan.md
```diff
@@ -72,7 +72,7 @@ Physical devices
 
 Virtual devices
 
-:  (Examples: veth, bridge, bond) These are fully under the control of the
+:  (Examples: veth, bridge, bond, vrf) These are fully under the control of the
```
schopin-pro, line 75: question (blocking): Seeing `vrf` in this list, how is revertability handled in `netplan try`?

Does `networkd` handle it on its own?

schopin-pro (commented):
Looks good, but I'd need confirmation on the revertability issue.

src/parse.c
```diff
@@ -2022,6 +2022,29 @@ static const mapping_entry_handler bond_params_handlers[] = {
...
+        } else {
+            if (component->vrf_link && component->vrf_link != npp->current.netdef)
+                return yaml_error(npp, node, error, "%s: interface '%s' is already assigned to vrf %s",
+                                  npp->current.netdef->id, scalar(entry), component->vrf_link->id);
+            component->vrf_link = npp->current.netdef;
+        }
+    }
+
+    return TRUE;
+}
```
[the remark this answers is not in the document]
slyon: ACK. The other functions (bonds/bridges) use the "netplan id" string as a link here, while I reworked this new function to use an actuall pointer to the other netdef. I think in the long run we want to switch to using pointers everywhere.

src/validation.c
slyon: Yes we can!

src/validation.c
slyon: It will not be rendered out via `netplan get`, which is what we want, as we want to rely on the VRF's `table` setting only. So I think that's all fine.

doc/netplan.md
slyon: This is a very good question! And I needed to do some investigation in order to answer this. But this is working as expected (and similar to `vlans`): `vrfs` are cleaned up on `netplan apply` (if using the `--state` parameter properly) or on `netplan try` (using the `--config-file` parameter). `netplan try --state ...` still behaves a bit odd, but that's unrelated to this PR.

VRFs are different to bonds/bridges, in that the backend (networkd/NM) does not need to initialize certain parameters, which cannot be changed afterwards and is therefore much more similar to a VLAN interface, which can just be deleted.
So I think this is all fine!

```
$ cat /etc/netplan/main_test.yaml 
network:
  version: 2
  renderer: networkd
  ethernets:
    enp0s31f6:
      dhcp4: true
      dhcp6: true
$ cat ./test.yaml
network:
  vrfs:
    vrf0:
      table: 1004
      interfaces: [enp0s31f6]
$ sudo LD_LIBRARY_PATH=. PYTHONPATH=. src/netplan.script --config-file=test.yaml try
Do you want to keep these settings?


Press ENTER before the timeout to accept the new configuration


Changes will revert in 111 seconds

# === meanwhile in another terminal === 
$ ip l show vrf0
110: vrf0: <NOARP,MASTER,UP,LOWER_UP> mtu 65575 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether a2:f7:52:d9:7c:5c brd ff:ff:ff:ff:ff:ff
# === terminal switch ===
^C
Reverting.
$ ip l show vrf0
Device "vrf0" does not exist.
```

(I actually found a bug/typo while testing this, which I fixed in b39dbea179f5b81609af002fd42986dccb97d1fc)

src/validation.c
schopin-pro: Well, we've usually tried to write out the YAML as close to the original input, even if redundant, hence the dirtyness tracking. If the `table` field is defined in the route *and* in the vrf (and both match), we might want to output it in both places, or alternatively warn that it's redundant, perhaps in verbose mode?

schopin-pro (approved):
LGTM, the `netplan get` question is not blocking for me.

src/validation.c
slyon: Thanks for the clarification, that makes a lot of sense.
I've added a debug log about ignoring the redundant `table` value in V3 of 362b85afdb6f8bd40737ca9989cccb1342bd10c2 to account for that for now.

slyon:
Thanks for your final review. I've added a debug log about ignoring the redundant table value in V3 of 362b85afdb6f8bd40737ca9989cccb1342bd10c2 to account for that for now.

Merging.

PR #288, opened 2022-07-14
Add support for VXLAN tunnels (LP: #1764716)

slyon:
## Description
Adding support for VXLAN tunnels using the sd-networkd and NetworkManager backends.
It builds upon #272 (squashed & rebased), extracted VXLAN bits. Applied on top of #285.
Schema according to (internal) spec FO042.

The first 3 commits e83ea8914223841b231843f630eeb382ee18d601 6d3597e356358de2d9a0f8d5e25af3fb27b2e542 02c4f0081ac5cf43adf8300bf49c28865f98d99a are pretty much unrelated to this PR. Those are drive-by fixes that I had been tracking in a separate branch, but the diff is fairly small, so I thought it'd make sense to include them into this PR instead of doing an extra one just for this.

Example:
```
network:
  ethernets:
    lo:
      addresses: [ 192.168.10.10/32 ]
  tunnels:
    vx1:
      mode: vxlan
      id: 1
      link: lo
      mtu: 8950
      accept-ra: no
      neigh-suppress: true
      mac-learning: false
      port: 4789
      local: 192.168.10.10
  bridges:
    br1:
      interfaces: [ vx1 ]
```

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1764716

### DIFF
--- a/src/abi.h
+++ b/src/abi.h
@@ -373,8 +373,6 @@ struct netplan_net_definition {
     NetplanNetDefinition* vrf_link;
     guint vrf_table;
 
-    /* TODO: `link` can also be used for vrf/vlan/sriov/... in the future */
-    NetplanNetDefinition* link;
     NetplanTristate bridge_neigh_suppress;
 
     /* vxlan */
--- a/src/abi.h
+++ b/src/abi.h
@@ -20,6 +20,8 @@
 #include <parse.h>
 #include <uuid.h>
 
+typedef int NetplanFlags;
+
 /* Those types are part of our ABI as they have been exposed in older versions */
 
 typedef enum {
--- a/src/abi.h
+++ b/src/abi.h
@@ -91,6 +91,7 @@ typedef enum {
     NETPLAN_TUNNEL_MODE_IPIP6       = 7,
     NETPLAN_TUNNEL_MODE_IP6GRE      = 8,
     NETPLAN_TUNNEL_MODE_VTI6        = 9,
+    NETPLAN_TUNNEL_MODE_VXLAN       = 10,
 
     /* systemd-only, apparently? */
     NETPLAN_TUNNEL_MODE_GRETAP      = 101,
@@ -179,6 +180,8 @@ typedef enum
     NETPLAN_TRISTATE_FALSE,      /*  0 */
     NETPLAN_TRISTATE_TRUE,       /*  1 */
 } NetplanTristate;
+
+typedef struct netplan_vxlan NetplanVxlan;
 
 /* Keep 'struct netplan_net_definition' in a separate header file, to allow for
  * abidiff to consider it "public API" (although it isn't) and notify us about
@@ -369,4 +372,13 @@ struct netplan_net_definition {
     /* netplan-feature: vrf */
     NetplanNetDefinition* vrf_link;
     guint vrf_table;
+
+    /* TODO: `link` can also be used for vrf/vlan/sriov/... in the future */
+    NetplanNetDefinition* link;
+    NetplanTristate bridge_neigh_suppress;
+
+    /* vxlan */
+    /* netplan-feature: vxlan */
+    gboolean has_vxlans;
+    NetplanVxlan* vxlan;
 };
--- a/src/abi.h
+++ b/src/abi.h
@@ -53,6 +53,8 @@ typedef enum {
     NETPLAN_IB_MODE_KERNEL,
     NETPLAN_IB_MODE_DATAGRAM,
     NETPLAN_IB_MODE_CONNECTED,
+
+    NETPLAN_IB_MODE_MAX_,
 } NetplanInfinibandMode;
 
 typedef enum {
--- a/src/names.c
+++ b/src/names.c
@@ -105,6 +105,14 @@ netplan_infiniband_mode_to_str[NETPLAN_IB_MODE_MAX_] = {
     return (val < sizeof( netplan_ ## _radical ## _to_str )) ?  netplan_ ## _radical ## _to_str [val] : NULL; \
 }
 
+/* @num_flags needs to account for the 0x0 value (index 0), which doesn't
+ * represent any flag, but still exists. So subtract 1 from the array length. */
+#define NAME_FUNCTION_FLAGS(_radical) const char *netplan_ ## _radical ## _name( NetplanFlags val) \
+{ \
+    size_t num_flags = sizeof(netplan_ ## _radical ## _to_str) / sizeof(char *) - 1; \
+    return (val <= 1 << (num_flags - 1)) ?  netplan_ ## _radical ## _to_str [__builtin_ffs(val)] : NULL; \
+}
+
 NAME_FUNCTION(backend, NetplanBackend);
 NAME_FUNCTION(def_type, NetplanDefType);
 NAME_FUNCTION(auth_key_management_type, NetplanAuthKeyManagementType);
@@ -113,6 +121,9 @@ NAME_FUNCTION(tunnel_mode, NetplanTunnelMode);
 NAME_FUNCTION(addr_gen_mode, NetplanAddrGenMode);
 NAME_FUNCTION(wifi_mode, NetplanWifiMode);
 NAME_FUNCTION(infiniband_mode, NetplanInfinibandMode);
+NAME_FUNCTION_FLAGS(vxlan_notification);
+NAME_FUNCTION_FLAGS(vxlan_checksum);
+NAME_FUNCTION_FLAGS(vxlan_extension);
 
 #define ENUM_FUNCTION(_radical, _type) _type netplan_ ## _radical ## _from_name(const char* val) \
 { \
--- a/src/names.c
+++ b/src/names.c
@@ -94,7 +94,7 @@ netplan_addr_gen_mode_to_str[NETPLAN_ADDRGEN_MAX] = {
 };
 
 static const char* const
-netplan_infiniband_mode_to_str[NETPLAN_ADDRGEN_MAX] = {
+netplan_infiniband_mode_to_str[NETPLAN_IB_MODE_MAX_] = {
     [NETPLAN_IB_MODE_KERNEL] = NULL,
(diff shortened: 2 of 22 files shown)


slyon:
I think this is shaping up nicely for review now. @netdever may I ask for your review, please, as you have been the original author of this work, especially if the `neigh-suppress` part is still functioning as expected?

I've mostly rebased it on top of the new `origin/main` branch (like using `NetplanTristate` types), transformed your work from introducing a new `vxlans` interface type to using an additional `tunnels.mode = vxlan` tunnel type and adopted the YAML schema a bit. The new YAML schema is pre-approved according to (internal) spec FO042.

netdever:
Awesome work! We're very excited for this. One quick note that at least in your example yml above, the `mode` parameter is missing, and this will result in an error:

```bash
/etc/netplan/50-cloud-init.yaml:57:13: Error in network definition: vx1: missing 'mode' property for tunnel
            id: 1
```

Regarding `neigh-suppress`, this must be present under the `Bridge` heading in the `.network` file when using systemd-networkd, otherwise it has no effect.

```bash
[Bridge]
NeighborSuppression=True
```

I initially handled this with:

```bash
    if (def->neigh_suppress) {
        g_string_append_printf(network, "\n[Bridge]\n");
        g_string_append_printf(network, "NeighborSuppression=%s\n", def->neigh_suppress ? "True" : "False");
    }
```

src/abi.h
```diff
@@ -367,4 +368,29 @@ struct netplan_net_definition {
...
+        guint source_port_max;
+        gboolean mac_learning;
+        gboolean arp_proxy;
+        gboolean short_circuit;
+        gboolean independent;
+        NetplanFlags notifications;
+        NetplanFlags checksums;
+        NetplanFlags extensions;
+        NetplanTristate do_not_fragment;
+    } vxlan;
```
schopin-pro, line 395: suggestion: how about using a private struct separately allocated?

That way we avoid exposing even more data through the legacy ABI, and non-vxlan netdefs don't have to pay the full price of this relatively big new addition.

```c
typedef struct netplan_vxlan NetplanVxlan;
struct netplan_net_definition {
    // ...
    NetplanVxlan* vxlan;
};

// In types.h
struct netplan_vxlan {
        guint vni;
        guint ageing;
        guint limit;
        guint tos;
        guint flow_label;
        guint source_port_min;
        guint source_port_max;
        gboolean mac_learning;
        gboolean arp_proxy;
        gboolean short_circuit;
        gboolean independent;
        NetplanFlags notifications;
        NetplanFlags checksums;
        NetplanFlags extensions;
        NetplanTristate do_not_fragment;
};
```

Of course there'd need to be some more changes around to account for the new redirection, as well as dealing with allocaitng and freeing the struct.

src/parse.c
```diff
@@ -2108,6 +2108,33 @@ handle_vrf_interfaces(NetplanParser* npp, yaml_node_t* node, const void* data, G
     return TRUE;
 }
 
+static gboolean
+handle_vxlan_source_port(NetplanParser* npp, yaml_node_t* node, const void* _, GError** error)
+{
+    assert_type(npp, node, YAML_SEQUENCE_NODE);
+    if (node->data.sequence.items.top - node->data.sequence.items.start != 2)
+        return yaml_error(npp, node, error, "%s: Expected exactly two values for port-range",
+                          npp->current.netdef->id);
```
schopin-pro, line 2117: question: how common is it to want to specifiy a single port?

I've never used VXLAN, but is it at all common to specify a range of 1? If so, we might want to accept scalar values? I'm guessing the answer is "not that common" given what I'm seeing from cursory googling, but I still think the question merits asking.

src/validation.c
```diff
@@ -331,6 +339,24 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, yaml
             return yaml_error(npp, node, error, "%s: invalid id '%u' (allowed values are 0 to 4094)", nd->id, nd->vlan_id);
     }
 
+    if (nd->type == NETPLAN_DEF_TYPE_TUNNEL &&
+        nd->tunnel.mode == NETPLAN_TUNNEL_MODE_VXLAN) {
+        if (nd->link)
+            nd->link->has_vxlans = TRUE;
+        else
+            nd->vxlan.independent = TRUE;
```
schopin-pro, line 347: nitpick: I'm not wild about mutating a netdef in the validation phase. Couldn't this be done at the end of the vxlan parsing instead?

schopin-pro (commented):
Solid work!

I have a few nitpicks here and there, and I'd really like to see the ABI question addressed before approving this PR. ("addressed" can very well be "not a problem, moving on", of course ;-) )

src/parse.c
slyon: I'm not an expert on VXLAN myself and agree with your "not that common" assessment. Therefore, I'll leave this as-is.
It can still be defined as e.g.: `port-range: [4200, 4200]`. If it turns out to be a common case, we could easily implement this usability optimization, but should also be reconsidering the "port-range" settings name in this case.

src/names.c
```diff
@@ -94,7 +94,7 @@ netplan_addr_gen_mode_to_str[NETPLAN_ADDRGEN_MAX] = {
 };
 
 static const char* const
-netplan_infiniband_mode_to_str[NETPLAN_ADDRGEN_MAX] = {
+netplan_infiniband_mode_to_str[] = {
```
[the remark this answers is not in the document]
slyon: The purpose was to not use `NETPLAN_ADDRGEN_MAX` for `netplan_infiniband_mode` as that's a different enum.

I'll be introducing `NETPLAN_INFINIBAND_MODE_MAX_` instead.

src/abi.h
slyon: This is quite a rework.. But you're right it makes sense not pulling it into the legacy ABI. Luckily the tests are all there already, so we can easily refactor this!

src/validation.c
slyon: ACK! Moving it into `parse.c`, where we now have a "Preprocessing" and "Postprocessing" stage directly before/after running the individual settings handlers.

slyon:
> Solid work!
> 
> I have a few nitpicks here and there, and I'd really like to see the ABI question addressed before approving this PR. ("addressed" can very well be "not a problem, moving on", of course ;-) )

Thank you for the quality review! I addressed all of your comments, could you please double-check my "V2" commits?

schopin-pro (changes requested):
Sadly, there's an issue with the bitfield name macro that needs to be solved before merging :/

src/nm.c
```diff
@@ -458,6 +460,60 @@ maybe_generate_uuid(const NetplanNetDefinition* def)
         uuid_generate((unsigned char*)def->uuid);
 }
 
+static void
+write_vxlan_parameters(const NetplanNetDefinition* def, GKeyFile* kf)
```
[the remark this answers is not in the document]
slyon: We can't, as we're also accessing some of the generic tunnel fields, like `def->tunnel.local_ip`/`def->tunnel.remove_ip`.

src/names.c
```diff
@@ -105,6 +105,11 @@ netplan_infiniband_mode_to_str[NETPLAN_IB_MODE_MAX_] = {
     return (val < sizeof( netplan_ ## _radical ## _to_str )) ?  netplan_ ## _radical ## _to_str [val] : NULL; \
 }
 
+#define NAME_FUNCTION_FLAGS(_radical) const char *netplan_ ## _radical ## _name( NetplanFlags val) \
+{ \
+    return (val < sizeof( netplan_ ## _radical ## _to_str )) ?  netplan_ ## _radical ## _to_str [__builtin_ffs(val)] : NULL; \
```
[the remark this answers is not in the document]
slyon: Good point! I added a small commit to fixup the original NAME_FUNCTION macro.

And modified your suggestion a bit more (adding an explanatory comment):
```C
/* @num_flags needs to account for the 0x0 value (index 0), which doesn't
 * represent any flag, but still exists. So subtract 1 from the array length. */
#define NAME_FUNCTION_FLAGS(_radical) const char *netplan_ ## _radical ## _name( NetplanFlags val) \
{ \
    size_t num_flags = sizeof(netplan_ ## _radical ## _to_str) / sizeof(char *) - 1; \
    return (val <= 1 << (num_flags - 1)) ?  netplan_ ## _radical ## _to_str [__builtin_ffs(val)] : NULL; \
}
```

slyon:
TA! I updated 055a6e07828d19b0b73fe0aeb3f4ae6abe37c192 (V3), added 72774d8 and ae582f2

My initial idea was to use `link` as an universal/multi-purpose field for different types, but I think it makes sense to have the `link` filed as part of the `NetplanVxlan` struct and making it exclusive to VXLAN, especially if its only allocated if needed. It's easier to grasp this way, so I implemented that in ae582f2.


Could you have another look, especially about the dirty tracking in `handle_vxlan_id_ref`?

slyon:
> Could you have another look, especially about the dirty tracking in `handle_vxlan_id_ref`?

Thinking about it... dirty tracking doesn't make too much sense for the `handle_netdef/vxlan_id_ref` handlers, does it? Either we have a ID reference set, or not. So the default value is no value. So I think we can probably drop the `mark_data_as_dirty()` call in both of those handlers.

schopin-pro:
Considering that the point of the dirty tracking is to be able to differentiate between default value and user input that happens to be the same as the default, the mechanism wouldn't work when said value is `null`, since the null handling would erase the user input before it hits the parsing code. So there's no need to do dirty tracking there.

Furthermore, does it actually make sense to have a vxlan entry without upstream link?

slyon:
> Furthermore, does it actually make sense to have a vxlan entry without upstream link?

Yes, there is a concept of "Independent" VXLAN interfaces: https://systemd.network/systemd.netdev.html#Independent=

schopin-pro (approved):
LGTM :)

schopin-pro:
I'd say leave the dirty tracking in place. While I think it isn't used (yet?), it doesn't hurt either and has the merits of being consistent with other data.

PR #293, opened 2022-11-05
lib:util: Add internal g_clear_list copy for old glib versions

marex:
Glib version before 2.64 do not contain g_clear_list() macro, add local copy of the macro for backward compatibility sake.

## Checklist

### DIFF
--- a/src/util-internal.h
+++ b/src/util-internal.h
@@ -100,3 +100,22 @@ complex_object_is_dirty(const NetplanNetDefinition* def, const void* obj, size_t
 
 gboolean
 is_multicast_address(const char*);
+
+/* Available since Glib 2.64 */
+#ifndef g_clear_list
+#define  g_clear_list(list_ptr, destroy)       \
+  G_STMT_START {                               \
+    GList *_list;                              \
+                                               \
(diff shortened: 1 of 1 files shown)


slyon:
Hello @marex, thank you very much for your contribution to Netplan!

Could you please explain a bit more about the context you want to use this patch in?
I think we had a similar discussion a while ago about using Netplan in older versions of RHEL and decided it would make more sense to keep such adaptions as downstream patches, instead of the upstream project.

https://github.com/slyon/netplan/commit/c827800ad6a24f1d43e1542874095b47eb795b2f

@Conan-Kudo WDYT?

Conan-Kudo:
I'm shipping it in the Fedora packaging: https://src.fedoraproject.org/rpms/netplan/blob/rawhide/f/netplan-glib-2.56-compat-rhel8.patch

But if we ever care to have CI for RHEL 8, then we probably need to mainline the patch I'm shipping now.

slyon:
Closing as per the argumentation above: backwards compatibility for older versions of glib should be handled in downstream packaging, depending on relevant context/versions. Instead of this upstream repository.

PR #295, opened 2022-11-09
parse: Add the filepath to OVS ports netdefs

daniloegea:
The filepath isn't being populated for OVS ports netdefs.

Add a unit test to catch this issue.

New unit test outcome without the fix:
```
FAIL: test_filepath_for_ovs_ports (test_libnetplan.TestNetDefinition)                                                                                                    
  File "/workspace/netplan-filepath/tests/test_libnetplan.py", line 382, in test_filepath_for_ovs_ports
    self.assertEqual(os.path.join(self.confdir, "a.yaml"), netdef.filepath)
    AssertionError: '/tmp/tmpp3mx4zw6/etc/netplan/a.yaml' != None

```

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -2842,10 +2842,10 @@ handle_network_ovs_settings_global_ports(NetplanParser* npp, yaml_node_t* node, 
         }
 
         if (npp->current.filepath) {
-            if (component->filepath)
-                g_free(component->filepath);
+            if (component1->filepath)
+                g_free(component1->filepath);
 
-            component->filepath = g_strdup(npp->current.filepath);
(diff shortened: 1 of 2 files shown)


daniloegea:
You are right! My patch wasn't updating the filepath in case of redefinition.

About setting the filepath in the netplan_netdef_new() function, we could do that. But the filepath update still must be done in the handle_*() functions as we are not creating new netdefs in case of redefinition.

slyon (approved):
Thank you! LGTM.
I just rebased it on top of `origin/main`. Let's wait for the CI to pass, before merging it.

PR #297, opened 2022-11-17
Fix some memory leaks

daniloegea:
* generate.c: free the options context allocated at the start of the main
      function.

    * openvswitch.c: make sure the GString "cmds" is free'd before the
      function returns.

    * parse.c: free the temporary list "values" before the function
      returns.

    * parse.c: parsed_defs is set to NULL preventing
      netplan_parser_reset() to free the memory allocated for the
      hash table. Not setting it to NULL is safe because all the data is moved to
      another hash table, leaving parsed_defs with no elements. Not
      nullifying parsed_defs here allows netplan_parser_reset() to free
      it.

    * types.c: in this particular case we g_strdup both key and value
      strings when inserting them in the hash table in
      parse.c:handle_generic_map(). Calling the destructor for both of
      them will make sure all the memory is free'd. When using
      free_hashtable_with_destructor in the future we'll need to make
      sure both key and value are not poiting to the same place to avoid
      a double free.

    * util.c: release the glob_t allocated in find_yaml_glob().

## Description
The before and after can be seen by running the generate program with valgrind

assuming you have an etc/netplan in /tmp/fakeroot:
`valgrind --leak-check=full ./generate -r /tmp/fakeroot/`

It can also be checked by compiling netplan with ASAN:

`CFLAGS=-fsanitize=address meson setup build --prefix=/usr`
`meson compile -C build --verbose`
and then
`./generate -r /tmp/fakeroot/`

## Checklist

### DIFF
--- a/src/openvswitch.c
+++ b/src/openvswitch.c
@@ -319,7 +319,7 @@ cleanup:
 gboolean
 netplan_netdef_write_ovs(const NetplanState* np_state, const NetplanNetDefinition* def, const char* rootdir, gboolean* has_been_written, GError** error)
 {
-    GString* cmds = g_string_new(NULL);
+    g_autoptr(GString) cmds = g_string_new(NULL);
     gchar* dependency = NULL;
     const char* type = netplan_type_to_table_name(def->type);
     g_autofree char* base_config_path = NULL;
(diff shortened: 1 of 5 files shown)


src/openvswitch.c
```diff
@@ -371,7 +371,7 @@ netplan_netdef_write_ovs(const NetplanState* np_state, const NetplanNetDefinitio
                 /* Set controller target addresses */
                 if (def->ovs_settings.controller.addresses && def->ovs_settings.controller.addresses->len > 0) {
                     if (!write_ovs_bridge_controller_targets(settings, &(def->ovs_settings.controller), def->id, cmds, error))
-                            return FALSE;
+                        return FALSE;
```
slyon, line 374: Unrelated, but makes sense to fix this up, while touching the file!

slyon (approved):
Thank you very much, LGTM!
I really like the systematic approach of searching for memory leaks using valgrind!

PR #298, opened 2022-11-24
API: implement APIs from the new specification

daniloegea:
* Implement some of the new APIs described in [1] and add some functions that were intended for internal use to the public header files.
 * Implement/change unit tests related to the changes.
 * Introduce some unit tests written in C using the cmocka framework. The idea is to make it easier to catch runtime problems like memory errors. One can just build the tests with -fsanitize=address or run them with valgrind to check for problems. The new files will not be compiled by default, the new option -Dunit_testing=true must be used during the meson setup step.
 * Integrate the new tests with the meson build configuration.
 * Add a .gcovr file to exclude the new tests from the coverage test. It a future commit we might just move the new tests directory to another place.
 * Refactor the public header files Definitions used by other public header files were moved to a new types.h file. The new file was included where it's needed and it shouldn't break external consumers.
 * Add new netplan_error functions with tests.
 * Enable cmocka unittests during coverage test.

[1] - https://discourse.ubuntu.com/t/spec-api-and-bindings-definition-for-libnetplan/29106

## Description

Some APIs still need to be implemented:
netplan_netdef_get_output_filename() 
netplan_netdef_match_interface()
_netplan_netdef_get_vf_count/vlan_filter/embedded_switch_mode/delay_rebind

The new cmocka unit testing infrastructure could be in another PR. Although I decided to include it here because it was easy to integrate, the new tests are automatically taken into account during the coverage test (and I'm already taking advantage of it) and it helps catching memory errors. My goal is to have a complete test suite for the C code with it.

The [Netplan + NetworkManager](https://github.com/slyon/NetworkManager/commit/5e97635f7935f2094599cda9d37cd4849d829bc1) integration patch builds fine with these changes.

## Checklist


slyon (changes requested):
First of all, thank you very much for this big refactoring!

I really like your re-structuring of the public header files. It seems to be API backwards compatible, too. (But we should probably still run a test re-build of NetworkManager+netplan integration patch, to make sure it doesn't break our current libnetplan consumer.)

Although I'd like to try keeping our dependency tree small, your introduction of CMocka as a new testing framework, is OK, as it is covering a new usecase: To catch memory leaks, as already used in #297 Also, it's only used during build/test time and when coding locally (i.e. no new runtime dependency).

You modified the `meson.build` system, while not keeping the `Makefile` based build up to date... I wonder if we should just drop the Makefile (or make it call `meson` instead), now that major Distros like Debian and Ubuntu are using the meson based build? If not, we'd want to keep both systems updated in parallel.

Wrt to applying the libnetplan API spec, I found a few things missing, as already lined out in the description above:
* netplan_netdef_match_interfaces() => exists, but is not exported publicly.
* netplan_netdef_get_output_filename() => missing. we'll probably need a new internal util for this, to calculate the path at runtime. It should return something like this path `conf_path = g_strjoin(NULL, "run/NetworkManager/system-connections/netplan-", def->id, "-", escaped_ssid, ".nmconnection", NULL);` (also for other backends like sd-networkd).
* _netplan_netdef_get_vf_count/vlan_filter/embedded_switch_mode/delay_rebind. missing or partially implemented, but not using the `_` "private" prefix, also not using the new string handling method (e.g. for ..._embedded_switch_mode).

Other observations from the API spec (spec should be updated accordingly):
* netplan_get_id_from_nm_filename => should be using the new string handling method, e.g. by copying it into "netplan_get_id_from_nm_filepath" (keeping the _filename function for backwards compat)
* netplan_get_filename_by_id => should be NETPLAN_DEPRECATED, as it isn't used anywhere and implemented in `abi_compat.c`.

There are several inline comments for you to have a look at, too.

README.md
```diff
@@ -21,7 +21,7 @@ Netplan's [documentation objectives](https://docs.google.com/document/d/1n47hwLm
 
 Steps to build netplan using the [Meson](https://mesonbuild.com) build system inside the `build/` directory:
 
-* meson setup build --prefix=/usr [-Db_coverage=true]
+* meson setup build --prefix=/usr [-Db_coverage=true] [-Dunit_testing=true]
```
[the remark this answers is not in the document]
daniloegea: Agree. In this case this option could be 'true' by default then.

include/parse.h
```diff
@@ -58,24 +33,24 @@ NETPLAN_PUBLIC void
...
-netplan_state_import_parser_results(NetplanState* np_state, NetplanParser* npp, GError** error);
+netplan_parser_load_yaml_from_fd(NetplanParser* npp, int input_fd, NetplanError** error);
 
 NETPLAN_PUBLIC gboolean
-netplan_parser_load_yaml_from_fd(NetplanParser* npp, int input_fd, GError** error);
+netplan_parser_load_nullable_fields(NetplanParser* npp, int input_fd, NetplanError** error);
 
 NETPLAN_PUBLIC gboolean
-netplan_parser_load_nullable_fields(NetplanParser* npp, int input_fd, GError** error);
+netplan_state_import_parser_results(NetplanState* np_state, NetplanParser* npp, NetplanError** error);
```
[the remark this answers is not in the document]
daniloegea: It was purely aesthetic :) the netplan_state one was in the middle of a block of netplan_parser functions, I felt I should move it hehe. But I agree with your point, reduce the diff now is more important, it could be done later.

tests/ctests/test_netplan_error.c
```diff
@@ -0,0 +1,66 @@
...
+    netplan_error_free(gerror);
+}
+
+void
+test_netplan_error_code(void** state)
+{
+    GError *gerror = g_error_new(1234, 5678, "%s: error message", "it failed");
+    uint64_t error_code = netplan_error_code(gerror);
+    GQuark domain = error_code >> 32;
+    gint error = error_code & ~0;
```
slyon, line 36: question: Hmm... I wonder what the ` & ~0` is doing here? Does it have any effect? We're basically matching a (32 bit?) bitmask, like 0xFFFFFFFF, i.e. take the full integer. Should this rather be an (explicit) typecast?

tests/ctests/test_netplan_misc.c
```diff
@@ -0,0 +1,241 @@
...
+    const char* filename = "INVALID/netplan-some-id.nmconnection";
+    char id[8];
+
+    ssize_t bytes_copied = netplan_get_id_from_nm_filepath(filename, "SOME-SSID", id, sizeof(id));
+
+    assert_int_equal(bytes_copied, 0);
+}
+
+void
+test_netplan_get_id_from_nm_filename_no_ssid(void **state)
```
slyon, line 101: Please, add a comment to the _from_nm_filename tests that this API is deprecated. Or even move it into a "test_netplan_deprecated.c" file?

tests/ctests/test_utils.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "types.h"
+#include "netplan.h"
+#include "parse.h"
+
+NetplanState *
+load_fixture_to_netplan_state(const char* filename)
+{
+
+    GError *error = NULL;
```
slyon, line 13:
```suggestion
    g_autoptr(GError) error = NULL;
```

We can make use of some glib goodies here, as it is included via "types.h". Could also use `g_autofree gchar* filepath` and/or `filepath = g_build_path(G_DIR_SEPARATOR_S, FIXTURESDIR, filename, NULL);` below. Or at least clear the GError at the end.

src/types.c
```diff
@@ -522,48 +567,66 @@ netplan_netdef_get_embedded_switch_mode(const NetplanNetDefinition* netdef)
...
+    return netdef->sriov_delay_virtual_functions_rebind;
 }
 
-NETPLAN_INTERNAL NetplanNetDefinition*
-_netplan_netdef_get_sriov_link(const NetplanNetDefinition* netdef)
+__attribute((alias("netplan_netdef_get_delay_virtual_functions_rebind"))) NETPLAN_INTERNAL gboolean
+_netplan_netdef_get_delay_vf_rebind(const NetplanNetDefinition* netdef);
+
+NETPLAN_INTERNAL guint
+_netplan_netdef_get_vf_count(const NetplanNetDefinition* netdef)
```
slyon, line 588: I'm sorry for the back and forth.. but I feel like we should actually drop this function, as we already have `_netplan_state_get_vrf_count_for_def`, which is more complete implementation wise, making use of the full "netplan_state" context (i.e. implicit sriov_link definitions).

slyon (changes requested):
Thank you very much for addressing my comments! We're getting really close to getting this merged now, but I left a 2nd round of (mostly small) inline comments.

tests/ctests/test_netplan_error.c
daniloegea: the idea was to mask the upper half of error code, but (looking at the generated code here) I suppose an implicit type casting is happening and my & ~0 is being optimized out in favor of an implicit cast. So yeah, an explicit cast would suffice.

btw, I assumed this was the semantics for netplan_error_code() based on the comment "uint64 to encode both domain and code" from the spec. Was that the idea? I feel like it would be better to return something like a `struct error` with 2 members...

src/util.c
```diff
@@ -576,12 +600,49 @@ netplan_get_id_from_nm_filename(const char* filename, const char* ssid)
...
+        if (ssid) {
+            g_autofree char* escaped_ssid = g_uri_escape_string(ssid, NULL, TRUE);
+            conf_path = g_strjoin(NULL, "run/NetworkManager/system-connections/netplan-", netdef->id, "-", escaped_ssid, ".nmconnection", NULL);
+        } else {
+            conf_path = g_strjoin(NULL, "run/NetworkManager/system-connections/netplan-", netdef->id, ".nmconnection", NULL);
+        }
+
+    } else if (netdef->backend == NETPLAN_BACKEND_NETWORKD || netdef->backend == NETPLAN_BACKEND_OVS) {
+        conf_path = g_strjoin(NULL, "run/systemd/network/10-netplan-", netdef->id, ".network", NULL);
+    }
```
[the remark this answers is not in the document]
daniloegea: Maybe we could make use of this function in an upcoming work? I mean, refactoring the code to use the new APIs are not really part of this work I guess (although I've done some refactoring here and there). Also, as using the new APIs in the old ones could potentially break "generate" (if libnetplan is not updated), maybe we should leave it for later.

About returning the absolute path, should we also take into account the root dir passed by the caller in this function? In write_nm_conf_access_point() we leave the final path build for last and append the root dir to it there. I think it make sense including the root dir in the output file path.

src/types.c
daniloegea: I don't have a strong opinion on that, so I'll drop it. We can add it later if we miss it...

src/util.c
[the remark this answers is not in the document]
slyon: > Maybe we could make use of this function in an upcoming work? I mean, refactoring the code to use the new APIs are not really part of this work I guess (although I've done some refactoring here and there). Also, as using the new APIs in the old ones could potentially break "generate" (if libnetplan is not updated), maybe we should leave it for later.

Agreed. But could you please add a  `/* TODO: make use of netplan_netdef_get_output_filename() */` comment to the corresponding places in nm.c and networkd.c in this case?

> About returning the absolute path, should we also take into account the root dir passed by the caller in this function? In write_nm_conf_access_point() we leave the final path build for last and append the root dir to it there. I think it make sense including the root dir in the output file path.

No strong opinion about this, but I feel like this util does not really need to know/care about the `rootdir`, so I'd rather not add it as an additional parameter and building of the final/full path to the caller, who is aware about the rootdir/context. We also have other utils, like `g_string_free_to_file` (e.g. used from networkd.c), that will help the caller to build the full path, incl. rootdir.

tests/ctests/test_netplan_error.c
```diff
@@ -0,0 +1,66 @@
...
+#include "types.c"
+#include "util.c"
+#include "parse.c"
+
+void
+test_netplan_error_message(void** state)
+{
+    const gchar* message = "it failed";
+    char error_message[100] = {0};
+    GError *gerror = g_error_new(1, 2, "%s: error message", message);
```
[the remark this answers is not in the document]
slyon: Right. I was thinking about calling `g_clear_error()` (https://docs.gtk.org/glib/func.clear_error.html) but all it does is calling `g_error_free()/netplan_error_free()` and setting the `*gerror` variable to `NULL`. Setting to `NULL` is not needed here as the variable goes out of scope at the end of the unit test anyways. So nvm.

tests/ctests/test_netplan_error.c
slyon: Good observation! But yes (having a 2nd look), that was the idea. We want to become GLib independent, so we don't really need the `domain` field, but also we don't want to drop that information for the time being. So masking/casting the relevant 32 bits will get the caller the correct error code now and in the future, without the need to change the API.

slyon (approved):
Thank you very much for resolving all the remarks! This is looking good to me, after comparing it to the spec again.

Let me rebase this on top of `origin/main` and add a small comment about the `netplan_error_code` return type, to clarify this situation in the header file (public API). And get it merged after it passes all the tests.

PR #299, opened 2022-11-24
cli:set: update only specific origin-hint if given (LP: #1997467)

slyon:
## Description
This PR solves multiple issues in `netplan set`. First, deletion of origin-hint files on update (i.e. if the netdef is already contained inside the origin-hint file). This happens when repeatedly calling netplan set on the same origin-hint and netdef, e.g.:
```
    Starting with an empty Netplan config in /etc/netplan:
    $ netplan set --origin-hint test "bridges.br54.dhcp4=false"
    => will create test.yaml, containing the br54 definition
    $ netplan set --origin-hint test "bridges.br54.dhcp4=true"
    => updating the same br54 netdef (re-using 'netplan set') will delete test.yaml
```

This is fixed by the first commit:
* 8931bf367b4c5185c67113da05668c34eb190562 generate:util: fix double-slash root filepath

Next, when parsing the full YAML hierarchy, stanza updates might be siphoned away, as the same interface might have been defined in another file (e.g. 0-snapd-defaults.yaml) and therefore any updates are redirected to that previous file, or ignored if an origin-hint is passed, which does not match the previous filename (e.g. origin-hint: 90-snapd-config.yaml). Still we always need to parse the full YAML hierarchy in order to validate our data. For example, somebody might call `netplan set --origin-hint=test "ethernets.br54.interfaces=[eth0]"`, which will end up having the new `br54` data in `hint.yaml` but still relies on a definition of `eth0` in some other YAML file. If somebody calls `netplan set --origin-hint=test "ethernets.eth0.dhcp4=true"` OTOH, we want to ignore any previous definition of `eth0` and write our update into `hint.yaml` (assuming it will be a higher order file, compared to the existing YAML hierarchy).
We implemented a `load_nullable_overrides()` function, to ignore global values or netdefs that are supposed to be written into an origin-hint file.

Last, we want global values (like "renderer") to be conserved per file, instead of written to any new file. When writing a new file (e.g. origin-hint.yaml) we want any `renderer` value inside the YAML hierarchy to stay unchanged. Also, we only want to add a `renderer` stanza to the new file if that was given via a "netplan set ..." YAML patch (or an already existing origin-hint file already contains this stanza). So we need to track the global values on a per YAML file base and confirm the filename/path before writing those.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1997467

### DIFF
--- a/netplan/cli/commands/set.py
+++ b/netplan/cli/commands/set.py
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 #
-# Copyright (C) 2020 Canonical, Ltd.
-# Author: Lukas Märdian <lukas.maerdian@canonical.com>
+# Copyright (C) 2020-2023 Canonical, Ltd.
+# Author: Lukas Märdian <slyon@ubuntu.com>
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -71,15 +71,59 @@ class NetplanSet(NetplanCommand):
         with tempfile.TemporaryFile() as tmp:
             libnetplan.create_yaml_patch(yaml_path, value, tmp)
             tmp.flush()
+
+            # Load fields that are about to be deleted (e.g. some.setting=NULL)
+            # Ignore those fields when parsing subsequent YAML files
             tmp.seek(0, io.SEEK_SET)
             parser.load_nullable_fields(tmp)
+
+            # Parse the full, existing YAML config hierarchy
             parser.load_yaml_hierarchy(self.root_dir)
+
+            # Load YAML patch, containing our update (new or deleted settings)
             tmp.seek(0, io.SEEK_SET)
             parser.load_yaml(tmp)
 
-        state = libnetplan.State()
-        state.import_parser_results(parser)
-        if self.origin_hint:
-            state.write_yaml_file(filename, self.root_dir)
-        else:
-            state.update_yaml_hierarchy(FALLBACK_FILENAME, self.root_dir)
+            # Validate the final parser state
+            state = libnetplan.State()
+            state.import_parser_results(parser)
+
+            if filename:  # only act on the output file (a.k.a. "origin-hint")
+                parser_output_file = libnetplan.Parser()
+
+                # Load fields that are about to be deleted ("some.setting=NULL")
+                # Ignore those fields when parsing subsequent YAML files
+                tmp.seek(0, io.SEEK_SET)
+                parser_output_file.load_nullable_fields(tmp)
+
+                # Load globals/netdefs that are to be ignored from the existing
+                # YAML hierarchy, as our patch is supposed to override settings
+                # in those netdefs via the output file.
+                # Those netdefs and globals must end up in the output file
+                # (a.k.a. "origin-hint", <filename>), have they been defined in
+                # pre-existing YAML files or not.
+                tmp.seek(0, io.SEEK_SET)
+                parser_output_file.load_nullable_overrides(tmp, constraint=filename)
+
+                # Parse the full YAML hierarchy and new patch, ignoring any
+                # nullable overrides (netdefs/globals) from pre-existing files
+                # and ignoring any nullable fields (settings to be deleted).
+                # This way we can avoid updates to certain netdefs/globals to be
+                # redirected into existing YAML files (defining those same
+                # stanzas) or ignored, but have them written out to the single
+                # output file.
+                # XXX: The origin file of each individual YAML setting/stanza
+                #      should be tracked individually, to avoid this
+                #      double-parsing workaround (LP: #2003727)
+                parser_output_file.load_yaml_hierarchy(self.root_dir)
+                tmp.seek(0, io.SEEK_SET)
+                parser_output_file.load_yaml(tmp)
+
+                # Import the partial parser state, ignoring duplicated netdefs
+                # from pre-existing YAML files, so we can force write the patch
+                # contents to the output file or update this file if exists.
+                state_output_file = libnetplan.State()
+                state_output_file.import_parser_results(parser_output_file)
+                state_output_file.write_yaml_file(filename, self.root_dir)
+            else:
+                state.update_yaml_hierarchy(FALLBACK_FILENAME, self.root_dir)
--- a/src/netplan.c
+++ b/src/netplan.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Canonical, Ltd.
+ * Copyright (C) 2021-2023 Canonical, Ltd.
  * Author: Lukas Märdian <slyon@ubuntu.com>
  *
  * This program is free software; you can redistribute it and/or modify
@@ -994,7 +994,7 @@ contains_netdef_type(gconstpointer value, gconstpointer user_data)
 }
 
 static gboolean
-netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int out_fd, GError** error)
+netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int out_fd, const char* out_fname, gboolean is_fallback, GError** error)
 {
     GHashTable *ovs_ports = NULL;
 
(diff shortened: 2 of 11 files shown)


netplan/cli/commands/set.py
```diff
@@ -73,7 +74,12 @@ def command_set(self):
             tmp.flush()
             tmp.seek(0, io.SEEK_SET)
             parser.load_nullable_fields(tmp)
-            parser.load_yaml_hierarchy(self.root_dir)
+            if self.origin_hint:  # update only this specific file
+                path = os.path.join(self.root_dir, 'etc', 'netplan', filename)
+                if os.path.isfile(path):
+                    parser.load_yaml(path)
+            else:
+                parser.load_yaml_hierarchy(self.root_dir)
```
schopin-pro, line 82: issue: this won't work in the general case

The only reason this works now is because you have standalone definitions. If your new definition refers to an interface defined in another file, `netplan set` will fail.

A quickfix would be to add a field to netdefs (in the `private_data` field) to mark them as "touched" when parsing a file without name (`npp->current.file_path == NULL`), and dump them out in `write_yaml_file` (or add another variant of this function with these semantics).

The problem with this quickfix is that you'll get the entire netdef dumped, not just the patched parts. A more involved fix would be to change the dirtiness tracker to store origin file rather than a simple boolean. This would allow for fine-grained control over which field we write out, but the implementation could get tricky (or *very* memory-hungry)

schopin-pro (changes requested):
See the only inline comment for details as to why this fix doesn't work :-(

netplan/cli/commands/set.py
slyon: Thank you very much for your review and hints. I need to allocate some extra time to dive deeper into this and fix it properly!

src/validation.c
```diff
@@ -322,6 +322,12 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, yaml
     if (missing_id_count > 0)
         return TRUE;
 
+    /* filepath: needs to be valid (i.e. no duplicated '//' chars) */
+    if (nd->filepath && g_strrstr(nd->filepath, "//"))
```
daniloegea, line 326: Well, a path with multiple slashes can be perfectly valid on UNIX. Consecutive slashes must be handled as a single one (unless the slash is the last character in the path and the last component is not a directory). Apparently the way double trailing slashes are handled is defined as implementation specific, but I guess Linux just handle it as though it is a single slash.

So, we can't throw an error because the path has multiple slashes, it would break POLA.

See [pathname](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271) and [pathname resolution](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13)

slyon: ACK. I'll revert this validation commit.

schopin-pro:
I'm not sure I fully grasp the new approach, but I have a first question in mind:

What happens if I have this file:

```yaml
# read-only file
network:
  ethernets:
    eth0:
      dhcp4: true
```

and I do `netplan set --origin-hint=writable.yaml ethernets.eth0=null` ?

slyon:
> I'm not sure I fully grasp the new approach, but I have a first question in mind:
> 
> What happens if I have this file:
> 
> ```yaml
> # read-only file
> network:
>   ethernets:
>     eth0:
>       dhcp4: true
> ```
> 
> and I do `netplan set --origin-hint=writable.yaml ethernets.eth0=null` ?

I think this example has a flaw.. we're generally expecting `read/write` permissions on the config files (at least for the owner/root), as of course we cannot update or modify read-only files and would error out.

Assuming that the "read-only file" pasted above is not the same as `writable.yaml`, this wouldn't matter in this case, though. If the `--origin-hint=writable.yaml` parameter is given, it means that Netplan is supposed to only work/write on that specific origin-hint file, considering the rest of the YAML hierarchy as read-only, just for context/reference of dependent interface definitions.

Basically, this `netplan set` commands means: "Delete eth0 from writable.yaml".
If `writable.yaml` exists (and contains an "eth0" definition) this definition would be deleted.
If `writable.yaml` doesn't exist, it would not exist after executing the command.
The `read-only_file.yaml` would stay untouched at all times.

schopin-pro (changes requested):
You're almost there, but there are still holes in your solution.

There's a way to paper over said holes by parsing twice the data, once for validation and once for rendering, but that's only a short-term kludge. The underlying issue is the baked-in assumption that netdef is defined only in one place, whereas even individual pieces of data can have several different origins. Fixing the root cause would take much more efforts though, so let's fix the issue at hand first :)

netplan/cli/commands/set.py
```diff
@@ -71,9 +71,43 @@ def command_set(self):
...
+                # Clear all nullable fields and overrides (i.e. netdefs/globals
+                # to be ignored) and re-load just the acutal nullable fields
+                # (i.e. things that are about to be deleted via "=NULL" inside
+                # the yaml patch)
+                # This allows to parse the yaml patch below, using
+                # load_yaml(tmp), not ignoring the globals/netdefs it is
+                # supposed to override.
+                parser.clear_nullable_fields()
+                tmp.seek(0, io.SEEK_SET)
+                parser.load_nullable_fields(tmp)
```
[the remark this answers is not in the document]
slyon: Yes, handling that internally, instead of adding public API is the way to go.

netplan/cli/commands/set.py
```diff
@@ -71,9 +71,43 @@ def command_set(self):
...
+
+            # Load YAML patch.
+            # If origin_hint was given any update from the yaml patch will be
+            # applied inside the origin_hint output file, ignoring instances of
+            # the same stanzas in the rest of the YAML hierarchy.
```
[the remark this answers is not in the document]
slyon: ACK. I added a testcase where we have an explicit VRF table `1005` and a VRF route for table `1005` then calling "netplan set" to change the VRF table to `1004` in the output-file/origin-hint, i.e. making it incompatible with its own route.

src/netplan.c
```diff
@@ -1018,11 +1018,26 @@ netplan_netdef_list_write_yaml(const NetplanState* np_state, GList* netdefs, int
...
+    /* Try to find a file specific (global) renderer.
+     * If this is the fallback file (70-netplan-set.yaml or <origin-hint>.yaml),
+     * a renderer parsed from a YAML patch takes precedence. */
+    if (out_fname && np_state->global_renderer) {
+        renderer = GPOINTER_TO_INT(g_hash_table_lookup(np_state->global_renderer, out_fname));
+        /* A renderer parsed from an (anonymous) YAML patch takes precendence
+         * (e.g. "netplan set ..."). Such data does not have any filename
+         * associated to it in the global_renderer map (i.e. empty string). */
+        if (is_fallback && g_hash_table_contains(np_state->global_renderer, ""))
+            renderer = GPOINTER_TO_INT(g_hash_table_lookup(np_state->global_renderer, ""));
```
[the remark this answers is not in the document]
slyon: Nice performance optimization, to avoid the duplicated lookup!

netplan/cli/commands/set.py
```diff
@@ -71,15 +73,55 @@ def command_set(self):
...
+                    validation_parser = libnetplan.Parser()
+                    tmp.seek(0, io.SEEK_SET)
+                    validation_parser.load_nullable_fields(tmp)
+                    validation_parser.load_yaml_hierarchy(self.root_dir)
+                    tmp.seek(0, io.SEEK_SET)
+                    validation_parser.load_yaml(tmp)
+                    validation_state = libnetplan.State()
+                    validation_state.import_parser_results(validation_parser)
+                except libnetplan.LibNetplanException as e:
+                    logging.error('New settings invalidate existing configuration: ' + str(e))
```
schopin-pro, line 95: nitpick (not-blocking): we're actually handling the validation error in a cleaner way in this very specific case than in the general case, where we basically just let import_parser_results throw an exception and hope something will catch it up the stack.

netplan/cli/commands/set.py
```diff
@@ -71,15 +73,55 @@ def command_set(self):
...
+            # be applied inside the output file, ignoring instances of the same
+            # stanzas in the rest of the YAML hierarchy.
             tmp.seek(0, io.SEEK_SET)
             parser.load_yaml(tmp)
 
         state = libnetplan.State()
         state.import_parser_results(parser)
-        if self.origin_hint:
+        if filename:
             state.write_yaml_file(filename, self.root_dir)
```
schopin-pro, line 125: suggestion (not-blocking): keep the main validation in the `state` state, and consider the de-facto incomplete state with the overridden fields as the secondary

This would allow you to only have one branching point: At first you go through the motions of pre-loading the patch, loading the sources, import them in the state, and only once you have a valid state do you actually bother to check whether you have an actual filename target. At that point, you either write out your state, or you create a second one with the additional override state, which you proceed to dump out.

Bonus points:
* that would solve my earlier nitpick about validation errors ;)
* if validation fails we'll have done less work

netplan/cli/commands/set.py
```diff
@@ -71,15 +73,55 @@ def command_set(self):
...
             tmp.flush()
+
             tmp.seek(0, io.SEEK_SET)
             parser.load_nullable_fields(tmp)
+            if filename:
+                # Validate the full YAML hierarchy without ignoring any NetDefs.
+                # Exit early if this YAML patch would lead to an invalid config.
+                # XXX: The origin file of each individual YAML setting/stanza
+                #      should be tracked individually, to avoid this
+                #      double-parsing workaround (LP: #2003727)
```
schopin-pro, line 84: thought (not-blocking): I'm still fairly convinced there are situations where the overall config passes validation but the overridden config doesn't, I just can't come up with one yet.

This bugs me. ;-)

schopin-pro (approved):
LGTM to me, I just have the usual couple of stray thoughts and remarks but nothing blocking, feel free to ignore them and merge :)

netplan/cli/commands/set.py
slyon: I will keep it this way, as the reason (override) for this specific parsing error might not be clear to the user, so I think it makes sense to give some extra context here.

netplan/cli/commands/set.py
slyon: Thanks, this should make the code much easier to grasp. I like that!

netplan/cli/commands/set.py
slyon: Right. I cannot rule that out either. If we can come up with any other failure case, let's document them in LP#2003727, so we can get them fixed properly.

slyon:
Thank you for all the input. CI is passing and I verified the reproducers from LP#1997467 are passing (using PPA build: https://launchpad.net/~slyon/+archive/ubuntu/netplan-set-lp1997467)

Merging.

PR #300, opened 2022-11-24
NetworkManager 1.40 compat & file permission fixes (LP: #1862600, LP: #1997348)

slyon:
## Description
* add default "wakeonlan" settings for ethernet devices only (NetworkManager 1.40 compat)
* Write YAML files using 0o600 permission
* YAML files written by the user or a (external) tool (e.g. "netplan set", NetworkManager, subiquity, ...) can contain sensitive information (e.g. WiFi passwords), so should stay secret (root/owner read-only).

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1997348, LP#1862600

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -25,6 +25,7 @@
 #include <glib/gstdio.h>
 #include <gio/gio.h>
 
+#include <sys/stat.h>
 #include <yaml.h>
 
 #include "parse.h"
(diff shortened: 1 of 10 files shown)


src/parse.c
```diff
@@ -3150,6 +3151,17 @@ gboolean
...
+     * It could contain sensitive information (e.g. WiFi passwords), so should
+     * stay secret. */
+    mode_t mask = S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+    struct stat info;
+    if (stat(filename, &info) < 0) {
+        g_set_error(error, G_FILE_ERROR, errno, "Cannot stat %s: %s",
+                    filename, strerror(errno));
+        return FALSE;
+    } else if (info.st_mode & mask)
+        g_warning("Weak permissions on %s", filename);
```
daniloegea, line 3164: suggestion: we probably should tell the user what is the correct permission they must use. Something like `"Weak permissions on %s (recommended is 600)"`

Or something like openssh does: `"Permissions for %s are too open. Netplan files should NOT be accessible by others."`

daniloegea (commented):
Looks good.

It just broke [autopkgtest](https://github.com/canonical/netplan/actions/runs/3541224415/jobs/5945226358#step:7:2312) I guess due to the new output.

daniloegea (approved):
Looks good to me. lgtm

PR #301, opened 2022-11-29
Fix 'netplan ip leases' crash (LP: #1996941)

daniloegea:
When the interface is not defined in the Netplan configuration, the command 'generate --mapping' will return nothing.

## Description

Fix for [LP#1996941](https://bugs.launchpad.net/netplan/+bug/1996941)

## Checklist

### DIFF
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -147,6 +147,9 @@ class NetplanIpLeases(utils.NetplanCommand):
             out = subprocess.check_output(argv, universal_newlines=True)
         except CalledProcessError:  # pragma: nocover (better be covered in autopkgtest)
             sys.exit(1)
+        if not out:
+            print("Interface '%s' is not managed by Netplan" % self.interface, file=sys.stderr)
+            sys.exit(1)
         mapping = {}
         mapping_s = out.split(',')
         for keyvalue in mapping_s:
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -147,9 +147,6 @@ class NetplanIpLeases(utils.NetplanCommand):
             out = subprocess.check_output(argv, universal_newlines=True)
         except CalledProcessError:  # pragma: nocover (better be covered in autopkgtest)
             sys.exit(1)
-        if not out:
-            print("Interface '%s' is not managed by Netplan" % self.interface, file=sys.stderr)
-            sys.exit(1)
         mapping = {}
         mapping_s = out.split(',')
         for keyvalue in mapping_s:
--- a/netplan/cli/commands/ip.py
+++ b/netplan/cli/commands/ip.py
@@ -146,6 +146,7 @@ class NetplanIpLeases(utils.NetplanCommand):
         try:
             out = subprocess.check_output(argv, universal_newlines=True)
         except CalledProcessError:  # pragma: nocover (better be covered in autopkgtest)
+            print("No lease found for interface '%s' (not managed by Netplan)" % self.interface, file=sys.stderr)
             sys.exit(1)
         mapping = {}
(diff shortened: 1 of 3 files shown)


tests/cli.py
```diff
@@ -785,5 +785,27 @@ def test_ip_leases_no_nm_lease(self):
         self.assertIn(b'No lease found', err)
         self.assertNotEqual(p.returncode, 0)
 
+    def test_ip_leases_generate_mapping_is_empty(self):
+        mock_generate = MockCmd('generate')
```
slyon, line 789: There's a MockCmd.set_returncode() method, which should be used here, to mimick `netplan ip`'s behavior.

tests/cli.py
```diff
@@ -785,5 +785,27 @@ def test_ip_leases_no_nm_lease(self):
...
+  version: 2
+  renderer: NetworkManager
+''')
+        p = subprocess.Popen(exe_cli +
+                             ['ip', 'leases', '--root-dir', self.workdir.name, 'enlol'],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        (out, err) = p.communicate()
+        self.assertEqual(out, b'')
+        self.assertIn(b"Interface 'enlol' is not managed by Netplan", err)
+        self.assertNotEqual(p.returncode, 0)
```
slyon, line 808: The (mocked) return code should actually be `1`

netplan/cli/commands/ip.py
```diff
@@ -147,6 +147,9 @@ def lease_method_nm_connection():
             out = subprocess.check_output(argv, universal_newlines=True)
         except CalledProcessError:  # pragma: nocover (better be covered in autopkgtest)
             sys.exit(1)
+        if not out:
```
slyon, line 150: So... I really wonder how `out` can be `None` here?

We're calling `subprocess.check_output()` above, which should throw a "CalledProcessError" in case `generate --mapping ...` returns non-zero, which it should in case of missing ID. Do you know what's the root cause for this? Do we need to fix `generate` to actually return non-zero in all error cases?

tests/cli.py
```diff
@@ -785,5 +785,27 @@ def test_ip_leases_no_nm_lease(self):
...
 
+    def test_ip_leases_generate_mapping_is_empty(self):
+        mock_generate = MockCmd('generate')
+        # When the interface in not defined in any netplan yaml file
+        # the command 'generate --mapping <interface>' will return nothing
+        mock_generate.set_output('')
+        os.environ.setdefault('NETPLAN_GENERATE_PATH', mock_generate.path)
+
+        c = os.path.join(self.workdir.name, 'etc', 'netplan')
+        os.makedirs(c)
```
slyon, line 796: I guess we could/should move this up into the `setUp` method of `class TestIp`, as that might be generically useful to more tests (in the future).

slyon (changes requested):
Thanks, I really like the improved error handling.

I left a few smaller comments inline, below. The bigger thing I'm pondering about is how this case can even happen? Shouldn't `subprocess.check_output()` fail in case `generate --mapping ...` returns non-zero? What's the root-cause for this?

I'd like us to fix the root-cause and move the new error message, into the existing try/except error handler, if possible.

tests/cli.py
daniloegea: The problem is that 'generate' is not returning non-zero when there are no netdefs in the state. So it returns 0 and doesn't print anything. I will update the PR with a fix in the 'generate' code as it seems to be the one that needs to be fixed.

tests/cli.py
daniloegea: Agree. I will do that in another PR not related to this bug fix.

slyon (approved):
Thank you, LGTM!

PR #302, opened 2022-11-30
Migrate from (deprecated) nose to pytest

daniloegea:
Pytest and pycoverage require some extra configuration to track test coverage in tests that spawn subprocesses. The new files .coveragerc and sitecustomize.py are intended to enable support for subprocesses.

The test set test_terminal.py is completely skipped when testing on my system and, because of that, coverage test was reporting a low coverage. I still need to see it running in our automation to validate if we still need to skip them. Another workaround would be add a nocover to the Terminal class.

Also:
  - Fix some deprecation warnings in some tests.
  - Remove this task from the TODO list :)

## Checklist

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -48,4 +48,4 @@ jobs:
           dch -v $(git describe --tags) "Autopkgtest CI testing (Jammy)"
       - name: Run autopkgtest (incl. build)
         run: |
-          sudo autopkgtest . -U --env=DPKG_GENSYMBOLS_CHECK_LEVEL=0 --env=DEB_BUILD_OPTIONS=nocheck -- lxd autopkgtest/ubuntu/jammy/amd64
+          sudo autopkgtest . --setup-commands='apt -y install python3-pytest python3-pytest-cov' -U --env=DPKG_GENSYMBOLS_CHECK_LEVEL=0 --env=DEB_BUILD_OPTIONS=nocheck -- lxd autopkgtest/ubuntu/jammy/amd64
--- a/TODO
+++ b/TODO
@@ -36,8 +36,6 @@
 
 - integrate 'netplan try' in tmux/screen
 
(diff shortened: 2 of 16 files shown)


.github/workflows/autopkgtest.yml
```diff
@@ -48,4 +48,4 @@ jobs:
           dch -v $(git describe --tags) "Autopkgtest CI testing (Jammy)"
       - name: Run autopkgtest (incl. build)
         run: |
-          sudo autopkgtest . -U --env=DPKG_GENSYMBOLS_CHECK_LEVEL=0 --env=DEB_BUILD_OPTIONS=nocheck -- lxd autopkgtest/ubuntu/jammy/amd64
+          sudo autopkgtest . --setup-commands='apt -y install python3-pytest python3-pytest-cov' -U --env=DPKG_GENSYMBOLS_CHECK_LEVEL=0 --env=DEB_BUILD_OPTIONS=nocheck -- lxd autopkgtest/ubuntu/jammy/amd64
```
sil2100, line 51: I assume this will no longer be required once we update the netplan packaging (dependencies)?

TODO
```diff
@@ -48,4 +46,4 @@
 
 - move tunnel_ttl into tunnel struct
 
-- store match.driver as a list rather than a string
```
sil2100, line 51: Newline needed please!

sil2100 (approved):
Okay, this looks like good to go. There is a missing newline, but I can fix that when merging. Also left an inline comment.

I checked the output of the CI test run here and compared this with the previous ones - coverage seems fine, tests seem to be ran as they did before. I'll merge this.

TODO
sil2100: Ah, ignore me! The newline was missing previously and you actually added it! My bad, read the diff wrong. All good!

.github/workflows/autopkgtest.yml
daniloegea: That's correct.

PR #303, opened 2022-12-14
Check and fix non-inclusive laguange

daniloegea:
## Description
This PR introduces a new github action used to check for non-inclusive wording in our code base and documentation using Woke.

While not all the occurrences of non-inclusive words were fixed, it also replaces many occurrences with proper language. Most of the remaining occurrences are due to external APIs and tools we rely on.

It provides alternative names for the options below:
    all-slaves-active -> all-members-active
    packets-per-slave -> packets-per-member

The yaml emitter will produce yaml documents using the new words.

Ideally, every single unit test containing the old words should be duplicated to also handle the new one, but at the moment only two tests were adapted.

## Checklist

### DIFF
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -1,12 +1,5 @@
 ignore_files:
   - abi-compat/*.xml
-  - tests/generator/test_vrfs.py
-  - tests/generator/test_passthrough.py
-  - tests/parser/test_keyfile.py
-  - tests/generator/test_bridges.py
-  - tests/generator/test_bonds.py
-  - tests/generator/test_tunnels.py
-  - tests/generator/test_common.py
 rules:
   # Including some terms from https://inclusivenaming.org/ that are not in the default ruleset
   - name: abort
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -1,15 +1,4 @@
 rules:
-  - name: slave
-    terms:
-      - slave
-    alternatives:
-      - follower
-      - replica
-      - standby
-    options:
-      # trigger findings when they are surrounded by ASCII word boundaries
-      word_boundary: true
-
   # Including some terms from https://inclusivenaming.org/ that are not in the default ruleset
   - name: abort
     terms:
@@ -34,3 +23,160 @@ rules:
       - segment/segmentation
       - separate/separation
 
+  # From https://github.com/canonical/Inclusive-naming
+  # Overtly racist terms
+  - name: whitelist
+    terms:
+      - whitelist
+      - white-list
+      - whitelisted
+      - white-listed
+      - whitelisting
+      - white-listing
+    alternatives:
+      - allowlist
+    severity: warning
+
+  - name: blacklist
+    terms:
+      - blacklist
+      - black-list
+      - blacklisted
+      - black-listed
+      - blacklisting
+      - black-listing
+    alternatives:
+      - denylist
+      - blocklist
+    severity: warning
+
+  - name: blackhat
+    terms:
+      - blackhat
+      - black hat
+    alternatives:
+      - malicious actor
+      - attacker
+    severity: warning
+
+  - name: illegal characters
+    terms:
+      - illegal characters
+    alternatives:
+      - invalid characters
+      - unsupported characters
+    severity: warning
+
+  - name: master
+    terms:
+      - master
+    alternatives:
+      - primary
+      - main
+    severity: warning
+    options:
+      word_boundary: true
+
+  - name: slave
+    terms:
+      - slave
+    alternatives:
+      - secondary
+      - replica
+    severity: warning
+    options:
+      word_boundary: true
+
+  - name: whitehat
+    terms:
+      - whitehat
+      - white hat
+    alternatives:
+      - researcher
+      - security specialist
+    severity: warning
+  # Overtly Sexist/Transphobic/Homophobic/Default Male Gendered Terms/Pejorative about Gender Identity/Discriminative
+  - name: chairman
+    terms:
+      - chairman
+      - foreman
+    alternatives:
+      - chair
+      - foreperson
+    severity: warning
+
+  - name: dummy
+    terms:
+      - dummy
+    alternatives:
+      - placeholder
+      - sample
+    severity: warning
+
+  - name: grandfathered
+    terms:
+      - grandfathered
+    alternatives:
+      - legacied
+    severity: warning
+
+  - name: guys
+    terms:
+      - guys
+    alternatives:
+      - people
+      - folks
+    severity: warning
+
+  - name: hang
+    terms:
+      - hang
+    alternatives:
+      - stop responding
+      - stall
+    severity: warning
+    options:
+      word_boundary: true
+
+  - name: man hours
+    terms:
+      - man hours
+    alternatives:
+      - staff hours
+      - hours of effort
+    severity: warning
+
+  - name: man in the middle
+    terms:
+      - man in the middle
+      - man-in-the-middle
+    alternatives:
+      - machine-in-the-middle
+      - person-in-the-middle
+    severity: warning
+
+  - name: manned
+    terms:
+      - manned
+    alternatives:
+      - staffed
+      - monitored
+    severity: warning
+
+  - name: middleman
+    terms:
+      - middleman
+    alternatives:
+      - middleperson
+      - intermediary
+    severity: warning
+
+  - name: sanity check
+    terms:
+      - sanity
+    alternatives:
+      - confidence check
+      - coherence check
+    severity: warning
+  # Ignore rules
+  - name: he
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -1,0 +1,36 @@
+rules:
+  - name: slave
+    terms:
+      - slave
+    alternatives:
+      - follower
+      - replica
+      - standby
+    options:
+      # trigger findings when they are surrounded by ASCII word boundaries
+      word_boundary: true
+
+  # Including some terms from https://inclusivenaming.org/ that are not in the default ruleset
+  - name: abort
+    terms:
+      - abort
+    alternatives:
+      - force quit
+      - halt
+      - close
+
+  - name: cripple
+    terms:
+      - cripple
+    alternatives:
+      - degraded
+      - restrict
+
+  - name: segregate
+    terms:
+      - segregate
+      - segregation
+    alternatives:
+      - segment/segmentation
+      - separate/separation
+
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -74,8 +74,6 @@ rules:
       - primary
       - main
     severity: warning
-    options:
-      word_boundary: true
 
   - name: slave
     terms:
@@ -84,8 +82,6 @@ rules:
       - secondary
       - replica
     severity: warning
-    options:
-      word_boundary: true
 
   - name: whitehat
     terms:
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -1,5 +1,12 @@
 ignore_files:
   - abi-compat/*.xml
+  - tests/generator/test_vrfs.py
+  - tests/generator/test_passthrough.py
+  - tests/parser/test_keyfile.py
+  - tests/generator/test_bridges.py
+  - tests/generator/test_bonds.py
+  - tests/generator/test_tunnels.py
+  - tests/generator/test_common.py
 rules:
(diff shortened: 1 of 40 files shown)


sil2100:
An initial thought when looking at the non-inclusive language check: the output from the test run isn't very 'useful'. Is that how the standard woke command reports warnings and errors? Since when looking at the CI run, I only see mentions insensitive words and its alternatives, but doesn't quite point to where the string appears in code. Would it be possible to make it pinpoint the location? I must say I never used `woke` myself ever.

Also, I think we'll have to wait with a final review of this card until Lukas and Steve are back as this basically requires changing the netplan schema, so we need to have an overall agreement on what we want it to be.

tests/integration/wifi.py
```diff
@@ -152,7 +152,7 @@ class TestNetworkManager(IntegrationTestsWifi, _CommonTests):
 
     def test_wifi_ap_open(self):
         # we use dev_w_client and dev_w_ap in switched roles here, to keep the
-        # existing device blacklisting in NM; i. e. dev_w_client is the
+        # existing device blocklisting in NM; i. e. dev_w_client is the
```
slyon, line 155: nitpick: I'd prefer `denylisting` here, "blocklisting" makes me think about a block of devices.

tests/generator/test_passthrough.py
```diff
@@ -122,22 +122,22 @@ def test_passthrough_type_nm_devices(self):
       networkmanager:
         passthrough:
           connection.uuid: 87749f1d-334f-40b2-98d4-55db58965f5f
-          connection.type: dummy''')
```
slyon, line 125: We should keep upstream's `dummy` type here: https://developer-old.gnome.org/NetworkManager/stable/settings-dummy.html

src/abi.h
```diff
@@ -269,6 +269,7 @@ struct netplan_net_definition {
         char* transmit_hash_policy;
         char* selection_logic;
         gboolean all_slaves_active;
+#define all_members_active all_slaves_active
```
slyon, line 272: I feel like we should rather rename the struct field and make use of the new `all_members_active` variable name in our internal code. Same goes for `packets_per_member` below.

In any case, I'd make the new, inclusive name the default variable name and have the old, non-inclusive name be the `#define`/alias.

src/parse.c
```diff
@@ -2112,6 +2112,7 @@ static const mapping_entry_handler bond_params_handlers[] = {
     {"transmit-hash-policy", YAML_SCALAR_NODE, {.generic=handle_netdef_str}, netdef_offset(bond_params.transmit_hash_policy)},
     {"ad-select", YAML_SCALAR_NODE, {.generic=handle_netdef_str}, netdef_offset(bond_params.selection_logic)},
     {"all-slaves-active", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(bond_params.all_slaves_active)},
+    {"all-members-active", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(bond_params.all_members_active)},
```
slyon, line 2115: The new/alternative "all-members-active" YAML stanza works for me, but let's also try to get Steve's opinion.

But I think I'd stick to the new, inclusive name for our internal variable naming.
```suggestion
    {"all-slaves-active", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(bond_params.all_members_active)},
    {"all-members-active", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(bond_params.all_members_active)},
```

src/parse.c
```diff
@@ -2125,6 +2126,7 @@ static const mapping_entry_handler bond_params_handlers[] = {
     {"gratuitious-arp", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(bond_params.gratuitous_arp)},
     /* TODO: unsolicited_na */
     {"packets-per-slave", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(bond_params.packets_per_slave)},
+    {"packets-per-member", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(bond_params.packets_per_member)},
```
slyon, line 2129: The new/alternative "packets-per-member" YAML stanza works for me, but let's also try to get Steve's opinion.

But I think I'd stick to the new, inclusive name for our internal variable naming.
```suggestion
    {"packets-per-slave", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(bond_params.packets_per_member)},
    {"packets-per-member", YAML_SCALAR_NODE, {.generic=handle_netdef_guint}, netdef_offset(bond_params.packets_per_member)},
```

doc/netplan-yaml.md
```diff
@@ -1144,7 +1144,7 @@ wpasupplicant installed if you let the `networkd` renderer handle wifi.
     > `bandwidth`, and `count`. This option is only used in 802.3ad
     > mode.
 
-  - **all-slaves-active** (bool)
+  - **all-members-active**, **all-slaves-active** (bool)
```
slyon, line 1147: I'd rather put the "all-slaves-active" title down into the description, e.g. as `Alias: **all-slaves-active**`

doc/netplan-yaml.md
```diff
@@ -1199,41 +1199,41 @@ wpasupplicant installed if you let the `networkd` renderer handle wifi.
...
+    > up on a new port, a notification is sent and possibly repeated if
     > this value is set to a number greater than `1`. The default value
     > is `1` and valid values are between `1` and `255`. This only
     > affects `active-backup` mode.
     >
     > For historical reasons, the misspelling `gratuitious-arp` is also
     > accepted and has the same function.
 
-  - **packets-per-slave** (scalar)
+  - **packets-per-member**, **packets-per-slave** (scalar)
```
slyon, line 1217: I'd rather put the "packets-per-slave" title down into the description, e.g. as Alias: **packets-per-slave**

tests/integration/base.py
```diff
@@ -443,7 +443,7 @@ def create_devices(klass):
         klass.dev_w_client = devs[1]
 
         # don't let NM trample over our fake AP
-        with open('/run/NetworkManager/conf.d/test-blacklist.conf', 'w') as f:
+        with open('/run/NetworkManager/conf.d/test-blocklist.conf', 'w') as f:
```
slyon, line 446: nitpick: I'd prefer denylist instead of blocklist (also above).

slyon (changes requested):
Thanks for setting up the new CI hook!

I think we should make use of `# wokelint:rule=...` comments throughout the code, to ignore exception (i.e. the things we cannot change).

Most cases are inside the tests, where we should be able to apply some search & replace logic to catch many cases at once. Maybe we need to adopt the `assert_nm` method in `tests/generator/base.py` to ignore `# wakeignore: ...` comments inside the provided keyfile.

For (internal) vairable naming, I'd not keep the old name around, but just rename the variable or struct fields accordingly, this should solve another class of warnings.

src/abi.h
daniloegea: I'll go with renaming them

.woke.yaml
```diff
@@ -0,0 +1,187 @@
+ignore_files:
+  - abi-compat/*.xml
+  - tests/generator/test_vrfs.py
+  - tests/generator/test_passthrough.py
+  - tests/parser/test_keyfile.py
+  - tests/generator/test_bridges.py
+  - tests/generator/test_bonds.py
+  - tests/generator/test_tunnels.py
+  - tests/generator/test_common.py
```
slyon, line 9: I thought about ignoring some of the tests as well (for now). But OTOH, now is the time to fix everything `woke` finds... It's only ~60 lines which need fixing left, so I think we should put in the extra effort to fix those instances as well.

We might need to adopt the `assert_nm()`, `assert_networkd()` and `generate_keyfile()` methods in their corresponding `base.py` class, but I think it's worth the effort. Like having a simple RegEx string-replace to search for something like `  # wokeignore:rule...` and replace it with an empty string in the corresponding keyfile/networkd assertion function, before comparing it to the actually generated NM/networkd config. This way we should be able to put our "wokeignore" comments into the test's code, while staying compatible with the generated output.

slyon (changes requested):
Thank you very much for addressing the previous comments and adding a bunch of ignore-rules at the relevant places.

I left some more inline comments, which we could improve upon, while also waiting for @vorlonofportland 's opinion on the new **all-members-active** and **packets-per-member** stanzas (alias).

Generally, I wonder where we are supposed to get our woke ruleset from (and how to keep it updated), besides woke's default ruleset. I see you picked some additional rules from https://inclusivenaming.org and from https://github.com/canonical/Inclusive-naming How did you get to pick those? Is there any policy around this?

Another topic is about the `blocklist` naming, which I'm personally not a big fan of, as I tend to read "block" as a chunk of something (i.e. a "block of code"). I'd prefer the allow/deny terminology. But I see there's ongoing discussions on this topic also in https://github.com/canonical/apport/pull/58

vorlonofportland:
These schema changes look appropriate to me.  The one thing I note is that the underlying backends (NM, networkd, and the kernel itself) still use the "slave" language.  Will there be follow-up to try to promulgate this name change down the stack?

slyon:
> The one thing I note is that the underlying backends (NM, networkd, and the kernel itself) still use the "slave" language. Will there be follow-up to try to promulgate this name change down the stack?

ACK. I think it would make sense to cross-reference existing "inclusive-naming" bug reports in the corresponding backends, if already existing, or create such bug reports upstream.

doc/netplan-yaml.md
```diff
@@ -1144,12 +1144,12 @@ wpasupplicant installed if you let the `networkd` renderer handle wifi.
...
     > mode.
 
-  - **all-slaves-active** (bool)
+  - **all-members-active** (bool)
 
     > If the bond should drop duplicate frames received on inactive ports,
     > set this option to `false`. If they should be delivered, set this
     > option to `true`. The default value is false, and is the desirable
-    > behavior in most situations.
+    > behavior in most situations. Alias: **all-slaves-active**  <!--- wokeignore:rule=slave -->
```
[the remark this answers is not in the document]
slyon: We should keep the "Alias" description as part of the same "quotation" block, IMO (e.g. using the `    > ` prefix)

tests/integration/base.py
slyon: There are more instances of the `test-blocklist.conf` below in `tearDownClass()`.

tests/generator/test_bonds.py
```diff
@@ -197,7 +197,82 @@ def test_bond_with_parameters(self):
...
+        down-delay: 30
+        all-slaves-active: true  # wokeignore:rule=slave
+        transmit-hash-policy: none
+        ad-select: none
+        arp-interval: 15
+        arp-validate: all
+        arp-all-targets: all
+        fail-over-mac-policy: none
+        gratuitious-arp: 10
+        packets-per-slave: 10  # wokeignore:rule=slave
```
sil2100, line 246: Should we also simply use `packets-per-member` here as well?

tests/generator/test_bonds.py
```diff
@@ -197,7 +197,82 @@ def test_bond_with_parameters(self):
...
+  bonds:
+    bn0:
+      parameters:
+        mode: 802.3ad
+        lacp-rate: 10
+        mii-monitor-interval: 10
+        min-links: 10
+        up-delay: 20
+        down-delay: 30
+        all-slaves-active: true  # wokeignore:rule=slave
```
sil2100, line 238: Same here, maybe `all-members-active`?

sil2100 (approved):
This looks good now, I see all comments from Lukas seem to be resolved. There's still 2 lines I'd like to change, but those are nitpicks.

tests/generator/test_bonds.py
daniloegea: These instances were left there on purpose. As these are valid configuration we can't remove just yet, we should have tests making use of them.

slyon:
I created some upstream RFC's about using inclusive language in our backends networkd and NetworkManager:

* https://github.com/systemd/systemd/issues/26257
* https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/1205

The Open vSwitch did the change already in v2.15:
* https://github.com/openvswitch/ovs/commit/91fc374a9c5a2a4d9520aaa3588a7a18338a476e

PR #162, opened 2020-09-07
generate: implement just-in-time behaviour for generate

xnox:
If system is initializing (basic.target not reached yet), and `netplan
generate` is called ensure that any netplan generated service units
are added to the initial boot transaction.

This should resolve cloud-init first-time booting with
systemd-networkd disabled, or with requirement to add wlan/wifi units.

## Description

I've tested this by sideloading this new code branch to lxd container.

Disabling systemd-networkd `systemctl disable systemd-networkd`

Removing `rm -rf /etc/netplan/50-cloud-init.yaml /var/log/cloud-init-output.log /var/log/cloud-init.log`

Executing `cloud-init clean`

And rebooting.

On reboot, I observed, that despite not being enabled systemd-networkd & netplan-ovs-cleanup did start or attempted to start, at the right time during initial boot, and correctly ordered before reaching network.target / network-online.target.

The below was tested on focal.

```
# journalctl -u systemd-networkd -u systemd-networkd.socket -u systemd-networkd-wait-online -u network.target -u basic.target -u network-online.target -u netplan-ovs-cleanup.service -b
-- Logs begin at Mon 2020-09-07 15:46:59 UTC, end at Mon 2020-09-07 23:17:01 UTC. --
Sep 07 22:56:13 wealthy-shiner systemd[1]: Listening on Network Service Netlink Socket.
Sep 07 22:56:13 wealthy-shiner systemd[1]: Condition check resulted in OpenVSwitch configuration for cleanup being skipped.
Sep 07 22:56:13 wealthy-shiner systemd[1]: Starting Network Service...
Sep 07 22:56:13 wealthy-shiner systemd-networkd[183]: eth0: IPv6 successfully enabled
Sep 07 22:56:13 wealthy-shiner systemd-networkd[183]: Enumeration completed
Sep 07 22:56:13 wealthy-shiner systemd[1]: Started Network Service.
Sep 07 22:56:13 wealthy-shiner systemd[1]: Reached target Network.
Sep 07 22:56:13 wealthy-shiner systemd[1]: Starting Wait for Network to be Configured...
Sep 07 22:56:13 wealthy-shiner systemd-networkd[183]: eth0: DHCPv4 address 10.54.137.244/24 via 10.54.137.1
Sep 07 22:56:13 wealthy-shiner systemd-networkd[183]: eth0: Gained IPv6LL
Sep 07 22:56:13 wealthy-shiner systemd-networkd-wait-online[184]: managing: eth0
[...]

PR #304, opened 2022-12-16
generator: Don't reload udev in the generator

daniloegea:
The systemd documentation states that any kind of process communication shouldn't happen inside generators. It can cause problems like the one reported in LP#1999178:

`Generators are run very early at boot and cannot rely on any external services. They may not talk to any other process. That includes simple things such as logging to syslog(3), or systemd itself (this means: no systemctl(1))!`

Ubuntu IS also affected by the same issue! We are observing long pauses during upgrades from Lunar to Mantic due to a package (usbmuxd) installing a udev rules file referring to a user that will be created late in the upgrade process. Every daemon-reload will stuck for 2 minutes, and it's happening many times. While the workaround for this specific case is easy (create the user earlier), users might run in to situations that's out of our control.

This change moves the call to the udevadm command from the generator to the Python code. To keep the behavior as close to the previous one as possible, the call is made as part of the generate command.

We could simply check if the generator is being called as a systemd generator and don't call udevadm, but in the end the binary still is a generator so we shouldn't even have that logic there.

The same is true for the calls we make to systemd using systemctl. Even though they are not executed when the binary is called as a generator, they probably shouldn't be there.

**UPDATE**:

I'm also proposing that we stop using generators to generate configuration. There are 2 main reasons for that:

1. According to systemd.generator(7), it's just wrong: `Generators should only be used to generate unit files, .d/*.conf drop-ins for them and symlinks to them, not any other kind of non-unit related configuration.`
2. Every daemon-reload will run the generator and regenerate all the configuration. It's a lightweight operation if one doesn't have a lot of YAMLs, but that's totally unnecessary.

This PR also includes an experimental systemd service that will call the generate binary before running systemd-networkd, Network Manager and cloud-init. While I'm not completely sure it's correct, it appears to work fine.

I have a PPA for Mantic containing these changes here https://launchpad.net/~danilogondolfo/+archive/ubuntu/netplan.io/+packages

## Checklist

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -42,6 +42,9 @@ jobs:
           pull-lp-source netplan.io
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
+          cp data/netplan-generator.service debian/
+          sed -i '/system-generator/d' debian/netplan-generator.install
+          sed -i '/rm debian\/tmp\/lib\/netplan\/generate/d' debian/rules
           TAG=$(git describe --tags $(git rev-list --tags --max-count=1))  # find latest (stable) tag
           REV=$(git rev-parse --short HEAD)  # get current git revision
           VER="$TAG+git~$REV"
--- a/.github/workflows/debci.yml
+++ b/.github/workflows/debci.yml
@@ -46,6 +46,9 @@ jobs:
           dget -u "https://deb.debian.org/debian/pool/main/n/netplan.io/netplan.io_$V.dsc"
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
+          cp data/netplan-generator.service debian/
+          sed -i '/system-generator/d' debian/netplan-generator.install
+          sed -i '/rm debian\/tmp\/lib\/netplan\/generate/d' debian/rules
           TAG=$(git describe --tags $(git rev-list --tags --max-count=1))  # find latest (stable) tag
           REV=$(git rev-parse --short HEAD)  # get current git revision
           VER="$TAG+git~$REV"
--- a/.github/workflows/network-manager.yml
+++ b/.github/workflows/network-manager.yml
@@ -37,6 +37,9 @@ jobs:
           pull-lp-source netplan.io
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
+          cp data/netplan-generator.service debian/
+          sed -i '/system-generator/d' debian/netplan-generator.install
+          sed -i '/rm debian\/tmp\/lib\/netplan\/generate/d' debian/rules
           echo "3.0 (native)" > debian/source/format  # force native build
           TAG=$(git describe --tags $(git rev-list --tags --max-count=1))  # find latest (stable) tag
           REV=$(git rev-parse --short HEAD)  # get current git revision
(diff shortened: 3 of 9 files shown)


jpsutton:
I rebuilt the Arch pkg for netplan with this PR applied and it resolved the issue for me.

slyon:
Just a note (not a full review, yet): We need to double check if this still works for LP#1669564 (https://github.com/canonical/netplan/commit/f4b0c54243539a9424d0f5987f1ee9e2288a529e)

daniloegea:
These changes break the new dbus integration tests:

```
test_dbus_config_apply (__main__.TestNetworkd.test_dbus_config_apply) ... FAIL
test_dbus_config_get (__main__.TestNetworkd.test_dbus_config_get) ... ok
test_dbus_config_set (__main__.TestNetworkd.test_dbus_config_set) ... ok

FAIL: test_dbus_config_apply (__main__.TestNetworkd.test_dbus_config_apply)
  File "/tmp/autopkgtest.iwnz9O/build.JXT/real-tree/tests/integration/dbus.py", line 202, in test_dbus_config_apply
    self.assertEqual(out.returncode, 0, msg=f"Busctl Apply() failed with error: {out.stderr}")
AssertionError: 1 != 0 : Busctl Apply() failed with error: Call failed: netplan apply failed: Child process exited with code 1
stdout: ''
stderr: 'Traceback (most recent call last):
    netplan.main()
  File "/usr/share/netplan/netplan/cli/core.py", line 56, in main
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 244, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 63, in run
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 244, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 123, in command_apply
    if run_generate and subprocess.call(generator_call) != 0:
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 389, in call
    with Popen(*popenargs, **kwargs) as p:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1024, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1901, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'netplan'
'


Ran 3 tests in 2.343s

FAILED (failures=1)
autopkgtest [15:05:08]: test dbus: -----------------------]
autopkgtest [15:05:09]: test dbus:  - - - - - - - - - - results - - - - - - - - - -
dbus                 FAIL non-zero exit status 1
```

slyon:
I wonder if we could do something similar to #162

Instead of calling udevadm syncronoulsy, we could consider deploying a new "netplan-udev-reload.service" unit, which executes something like `udevadm control --reload && udevadm trigger --attr-match=subsystem=net` as a OneShot unit.

From the generator we could then eque this "oneshot" command by calling `systemctl start --no-block --no-ask-password [--fail] netplan-udev-reload.service` (or maybe use `systemd-run ...`?), which should make it be executed after the current "daemon-reload" transaction is done, while still executing our commands.

daniloegea:
Closing. Superseded by https://github.com/canonical/netplan/pull/488

PR #306, opened 2023-01-03
Netplan api iterator

daniloegea:
## Description

The Netplan state iterator API is intended to enable consumers to walk through the ordered list of network definitions. It is useful, for instance, when you need to save the netdefs to files but you don't know the netdef ID to lookup for it.

The idea of having a non-opaque struct exposed to consumers was borrowed from Glib ([here](https://github.com/GNOME/glib/blob/abd76e02864163b6a627e5c5722cb4d6a8d35501/glib/ghash.h#L46) and [here](https://github.com/GNOME/glib/blob/abd76e02864163b6a627e5c5722cb4d6a8d35501/glib/ghash.c#L1143)). By doing this, the iterator can be placed at the stack, which is much more convenient.

This API was integrated and tested with the NetworkManager+Netplan integration code base.

## Checklist

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -69,4 +69,8 @@ typedef enum {
 
 typedef GError NetplanError;
 
-typedef struct netplan_state_iterator NetplanStateIterator;
+typedef struct _NetplanStateIterator NetplanStateIterator;
+
+struct _NetplanStateIterator {
+    void* placeholder;
+};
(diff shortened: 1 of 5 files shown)


slyon (approved):
LGTM, thanks!
I think this is a useful addition to our API, instead of accessing the internal `netdefs` HashTable/Map datastructure directly, as has been done previously in such cases.

PR #307, opened 2023-01-04
apply: Fix crash when OVS is stopped (LP#1995598)

daniloegea:
If openvswitch is installed but not running, netplan apply will try to call the service anyway and will crash. In this case we probably should ignore the error and log a warning message to the user.

Add an integration test that will be executed by autopkgtest.

## Description

It addresses https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1995598

The problem can be reproduced with:
```
lxc launch ubuntu:22.04 jammy --vm
lxc exec jammy bash
apt install openvswitch-switch
systemctl stop ovsdb-server
netplan apply
```
## Checklist

### DIFF
--- a/netplan/cli/ovs.py
+++ b/netplan/cli/ovs.py
@@ -23,7 +23,7 @@ import re
 from netplan.cli.utils import systemctl_is_active
 
 OPENVSWITCH_OVS_VSCTL = '/usr/bin/ovs-vsctl'
-OPENVSWITCH_OVSDB_SERVER_UNIT = 'ovsdb-server'
+OPENVSWITCH_OVSDB_SERVER_UNIT = 'ovsdb-server.service'
 # Defaults for non-optional settings, as defined here:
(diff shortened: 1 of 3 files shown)


netplan/cli/ovs.py
```diff
@@ -20,7 +20,10 @@
 import subprocess
 import re
 
+from netplan.cli.utils import systemctl_is_active
+
 OPENVSWITCH_OVS_VSCTL = '/usr/bin/ovs-vsctl'
+OPENVSWITCH_OVSDB_SERVER_UNIT = 'ovsdb-server'
```
slyon, line 26: Let's call it `ovsdb-server.service` explicitly

slyon (approved):
Thanks, lgtm!
I added some small bit to call `ovsdb-server.service` by its explicit name (including the .service suffix). Let's merge it once it passes the tests.

PR #308, opened 2023-01-05
parser: return the correct error on failure (LP#2000324)

daniloegea:
If the parsing process finishes with an error, we should return this error instead of assuming it's a missing interface.

For instance, the configuration below will fail with "br0: interface 'ens3' is not defined" when the actual error is "invalid IP family '-1'" due to the invalid IP used as the default route:

```
network:
  bridges:
    br0:
      interfaces:
        - ens3
      routes:
        - to: 0/0
          via: 100.64.0.1
  ethernets:
    ens3: {}
  version: 2
```

## Description

This PR addresses [LP#2000324](https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2000324)

The configuration above was failing with:

```
$ netplan generate --root-dir /workspace/netplan-issues/fakeroot
/workspace/netplan-issues/fakeroot/etc/netplan/static.yaml:5:11: Error in network definition: br0: interface 'ens3' is not defined
        - ens3
          ^
```
With this change this is the result:

```
$ LD_LIBRARY_PATH=build/src/ PYTHONPATH=. src/netplan.script generate --root-dir /workspace/netplan-issues/fakeroot
/workspace/netplan-issues/fakeroot/etc/netplan/static.yaml:7:15: Error in network definition: invalid IP family '-1'
        - to: 0/0
              ^
```

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -3103,6 +3103,10 @@ process_document(NetplanParser* npp, GError** error)
             break;
     } while (still_missing > 0 || npp->missing_ids_found > 0);
 
+    /* If an error already occurred we should return and not assume it's a missing interface*/
+    if (error && *error)
+        return ret;
+
     if (g_hash_table_size(npp->missing_id) > 0) {
         GHashTableIter iter;
         gpointer key, value;
--- a/src/parse.c
+++ b/src/parse.c
@@ -3105,7 +3105,7 @@ process_document(NetplanParser* npp, GError** error)
 
     /* If an error already occurred we should return and not assume it's a missing interface*/
     if (error && *error)
-        return ret;
+        goto cleanup;
 
     if (g_hash_table_size(npp->missing_id) > 0) {
         GHashTableIter iter;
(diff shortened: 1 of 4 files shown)


slyon (approved):
Thanks, lgtm. I've added some small bits to cleanup the `npp->missing_id` hashmap in the error case. WDYT?

src/parse.c
```diff
@@ -3116,11 +3120,12 @@ process_document(NetplanParser* npp, GError** error)
         g_hash_table_iter_next (&iter, &key, &value);
         missing = (NetplanMissingNode*) value;
 
-        return yaml_error(npp, missing->node, error, "%s: interface '%s' is not defined",
-                          missing->netdef_id,
-                          key);
+        ret = yaml_error(npp, missing->node, error, "%s: interface '%s' is not defined",
+                         missing->netdef_id, key);
+        goto cleanup;
```
daniloegea, line 3125: this goto is not needed (but doesn't hurt) as the last lines will be executed anyway

daniloegea: also, this hashtable is released by netplan_parser_reset(), but that's fine keeping it here.

slyon: Thanks for your comments!
* IMO it is nice to clear the hashtable close to its creation.. doesn't hurt to "clear" it again in netplan_parser_reset * ACK. I was pondering about adding the goto or nor. But decided to add it, as this way we don't need to think about it in the future should we add a block of code in between that if condition and the return statement

PR #309, opened 2023-01-05
parser: plug a memory leak

daniloegea:
The route->type field wasn't being free'ed in case of error. Caught by a unit test compiled with ASAN.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -1993,7 +1993,8 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, const void* _, GError** err
 
 err:
     if (npp->current.route) {
-        free_route(npp->current.route);
+        g_free(npp->current.route->type);
+        g_free(npp->current.route);
         npp->current.route = NULL;
     }
     return FALSE;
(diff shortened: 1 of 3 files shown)


src/parse.c
```diff
@@ -1993,7 +1993,7 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, const void* _, GError** err
 
 err:
     if (npp->current.route) {
-        g_free(npp->current.route);
+        free_route(npp->current.route);
         npp->current.route = NULL;
```
slyon, line 1997: Use `ip_rule_clear()` instead of manual free & nullify.

slyon (changes requested):
Thanks, having a closer look myself. I think we can further simplify this, by making use of the `route_clear()` function (instead of manual free & nullify), as defined using the `CLEAR_FROM_FREE(free_route, route_clear, NetplanIPRoute);` macro. Also, the `route_clear` function is already declared in `types.h`, so no need to add `free_route` into `types-internal.h` (and no need to make it non-static).

PR #310, opened 2023-01-10
networkd: make sure VXLAN is in the right section (LP: #2000713)

daniloegea:
If an interface has a list of routes and is linked to a vxlan, the VXLAN keyword is added to the [Route] section in the resulting networkd configuration file.

This change makes the vxlan handling to happen before routes so VXLAN= will be added to the [Network] section.

It fixes [LP#2000713](https://bugs.launchpad.net/netplan/+bug/2000713)

## Description

Can be reproduced with:

```
network:
  renderer: networkd
  ethernets:
    eth0:
      routes:
        - to: 10.20.30.40/32
          via: 10.20.30.1
  tunnels:
    vxlan1:
      mode: vxlan
      id: 1
      link: eth0
```

This is the current output. The VXLAN= keyword was added to the [Route] section.

```
$ cat fakeroot/run/systemd/network/10-netplan-eth0.network
[Match]
Name=eth0

[Network]
LinkLocalAddressing=ipv6

[Route]
Destination=10.20.30.40/32
Gateway=10.20.30.1
VXLAN=vxlan1
```

This is the result with this patch:

```
[Match]
Name=eth0

[Network]
LinkLocalAddressing=ipv6
VXLAN=vxlan1

[Route]
Destination=10.20.30.40/32
Gateway=10.20.30.1
```

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -872,6 +872,20 @@ netplan_netdef_write_network_file(
     if (def->vrf_link)
         g_string_append_printf(network, "VRF=%s\n", def->vrf_link->id);
 
+    /* VXLAN options */
+    if (def->has_vxlans) {
+        /* iterate over all netdefs to find VXLANs attached to us */
+        GList *l = np_state->netdefs_ordered;
+        const NetplanNetDefinition* nd;
+        for (; l != NULL; l = l->next) {
+            nd = l->data;
+            if (nd->vxlan && nd->vxlan->link == def &&
+                nd->type == NETPLAN_DEF_TYPE_TUNNEL &&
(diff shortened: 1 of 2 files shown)


slyon (approved):
Thanks, lgtm.
I've added some bits to make use of the `ND_EMPTY` and `ND_VXLAN` matching templates in your test case, in order to reduce the code/diff size a bit.

slyon:
Just as a note to self: In the future, we want to switch to using GLib's `g_key_file_set_*` function to build that networkd ini files (similar to how we're doing it in `nm.c` nowadays). This should avoid errors such as this one.

PR #311, opened 2023-01-10
vxlan: initialize mac_learning with 1

daniloegea:
Vxlan learning is enabled by default on Linux. On Netplan it's a boolean variable that we initialize with false and only include in the generated configuration if it's true. That means we can't disable it on vxlan interfaces even if we define mac-learning: false explicitly in the yaml file.

That is not the ideal solution! The proper way to do that is defining it as a tristate variable but it requires changes in the netplan_vxlan structure. One bad side effect of this change is that the yaml configuration generated from netdefs and the NM/networkd configuration will contain the option mac-learning even if we don't have it in the original yaml.

This addresses [LP#2000712](https://bugs.launchpad.net/netplan/+bug/2000712)

## Checklist

### DIFF
--- a/tests/generator/test_tunnels.py
+++ b/tests/generator/test_tunnels.py
@@ -1069,12 +1069,13 @@ ConfigureWithoutCarrier=yes
     vx0:
       mode: vxlan
       id: 1005
+      mac-learning: true
       remote: 10.0.0.5
       port-range: [100, 10]''')
         self.assertIn("swapped invalid port-range order [MIN, MAX]", out)
 
         self.assert_networkd({'vx0.netdev': ND_VXLAN % ('vx0', 1005) +
-                              'Remote=10.0.0.5\nPortRange=10-100\nIndependent=true\n',
+                              'Remote=10.0.0.5\nMacLearning=true\nPortRange=10-100\nIndependent=true\n',
                               'vx0.network': ND_EMPTY % ('vx0', 'ipv6')})
 
     def test_vxlan_ip6_multicast(self):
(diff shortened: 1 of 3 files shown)


daniloegea:
I'm not really happy with this change, but it avoids ABI changes.

There are other options that will not be emitted if they are false, but the problem doesn't really shows up when they are false by default on the system.

The proper solution would be to convert all these options to tristates.

daniloegea:
I converted three settings from gboolean to tristates. It's shouldn't change the struct size/layout as both types are actually ints.

src/parse.c
```diff
@@ -1615,12 +1615,14 @@ handle_vxlan_guint(NetplanParser* npp, yaml_node_t* node, const void* data, GErr
     return handle_generic_guint(npp, node, npp->current.vxlan, data, error);
 }
 
-static gboolean
+// LCOV_EXCL_START
+static gboolean __attribute__ ((unused))
```
daniloegea, line 1619: I didn't want to remove this function, but I feel like I should...

slyon: Yes, please remove

tests/generator/test_tunnels.py
```diff
@@ -699,6 +699,146 @@ def test_vxlan(self):
...
+    vxlan1005:
+      link: eth0
+      mode: vxlan
+      id: 1005
+      mac-learning: true
+      arp-proxy: true
+      short-circuit: true''' % {'r': self.backend})
+
+        if self.backend == 'networkd':
+            self.assert_networkd({'vxlan1005.netdev': '''[NetDev]
```
slyon, line 721: I think we should make some use of the `ND_*` templates from `generator/base.py` here, e.g. ND_VLAN, to avoid repeating ourselves too much. Maybe concatenate or str.replace()  relevant pieces of the string as needed.

slyon (approved):
lgtm overall! Let's reduce that now unused `handle_vxlan_bool` helper and try to reduce the new test's size a bit, by using some of the available templates from `base.py`

PR #312, opened 2023-01-17
doc/examples: remove unnecessary route for IPv6 on-link gateways

sbraz:
## Description
Hi,
When the on-link flag is enabled, the additional route has no effect. This extra route is only required when the on-link flag is *not* set.

Indeed, for netorkd, when "on-link" is true, the resulting systemd-networkd config file contains "GatewayOnLink=true", which translates to "flags: onlink" in its logs. The resulting route appears with this "onlink" flag in the output of "ip route".

Similarly, for NetworkManager, "route1_options=onlink=true" gets added by
https://github.com/canonical/netplan/blob/e5ff9f6dbeb20a8a3e6abc712efcf1b0008afa75/src/nm.c#L228-L230

## Checklist

- [x] Runs `make check` successfully → except for `test_with_empty_config` which also fails on the `main` branch on my Ubuntu 22.10 test server, so I'm assuming this is entirely unrelated.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -378,7 +378,7 @@ network:
                on-link: true
 ```
 
-For IPv6 the config would be very similar, with the notable difference being an additional scope: link host route to the router's address required:
+For IPv6 the config would be very similar:
 
 ```yaml
 network:
@@ -388,8 +388,6 @@ network:
         ens3:
             addresses: [ "2001:cafe:face:beef::dead:dead/64" ]
             routes:
-             - to: "2001:cafe:face::1/128"
-               scope: link
              - to: default # or "::/0"
                via: "2001:cafe:face::1"
                on-link: true
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -667,7 +667,7 @@ eth1:
   - **scope** (scalar)
 
     > The route scope, how wide-ranging it is to the network. Possible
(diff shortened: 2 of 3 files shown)


daniloegea:
Hello and thank you for your PR.

That sounds correct.

We also should mention in the docs that `scope` only applies to IPv4 routes (I can check that later). We mention that in the code but we include it in the generated configuration anyway:

https://github.com/canonical/netplan/blob/e5ff9f6dbeb20a8a3e6abc712efcf1b0008afa75/src/nm.c#L204

slyon (approved):
Thanks for your contribution. Yes, I agree with the understanding that an additional `scope: link` route is not needed for IPv6. Checking the origin of this example doesn't give any special reason for it to be included, neither: https://github.com/canonical/netplan.io/commit/ed72e3a41c7dc03b2384b815167878b23c7fdde0

And checking the relevant [systemd-networkd docs](https://systemd.network/systemd.network.html#Scope=1) for `Scope=` states the following:
> For IPv4 route, defaults to "host" if Type= is "local" or "nat", and "link" if Type= is "broadcast", "multicast", "anycast", or "unicast". In other cases, defaults to "global". The value is not used for IPv6.

LGTM!

Also, I agree with @daniloegea's assessment that we should add a comment about `scope` being effective only for IPv4. I'll add such comment before merging.

PR #313, opened 2023-01-17
Check if the interface name is too long (LP: #1988749)

daniloegea:
## Description

Netplan will allow the user to enter an interface name that's too long without issues any warning message. The backends will fail to process such an interface with errors like:

```
Jan 17 08:37:31 ubuntu2204 systemd-udevd[176]: eth0: Invalid network interface name, ignoring: anotherverylongname0
```

```
Jan 17 14:17:53 ubuntu2204 systemd-networkd[1019]: /run/systemd/network/10-netplan-notagoodnameforabridge0.netdev:2: Interface name is not valid or too long, ignoring assignment: notagoodnameforabridge0
Jan 17 14:17:53 ubuntu2204 systemd-networkd[1019]: NetDev without Name= configured in /run/systemd/network/10-netplan-notagoodnameforabridge0.netdev. Ignoring
```

Also, netplan apply will crash when trying to rename the interface:

```
oot@ubuntu2204:~# netplan apply
    netplan.main()
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 231, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 61, in run
    self.run_command()
  File "/usr/share/netplan/netplan/cli/utils.py", line 231, in run_command
    self.func()
  File "/usr/share/netplan/netplan/cli/commands/apply.py", line 245, in command_apply
    subprocess.check_call(['ip', 'link', 'set',
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ip', 'link', 'set', 'dev', 'eth0', 'name', 'anotherverylongname0']' returned non-zero exit status 255
```

Ideally we should fail on such cases, but at this point some users might have invalid names in their configuration so it would cause problems.

It addresses LP#1988749

## Checklist

### DIFF
--- a/netplan/cli/commands/apply.py
+++ b/netplan/cli/commands/apply.py
@@ -36,6 +36,7 @@ from netplan.cli.ovs import OvsDbServerNotRunning, apply_ovs_cleanup
 
 OVS_CLEANUP_SERVICE = 'netplan-ovs-cleanup.service'
 
+IF_NAMESIZE = 16
 
 class NetplanApply(utils.NetplanCommand):
 
@@ -234,6 +235,9 @@ class NetplanApply(utils.NetplanCommand):
             # rename non-critical network interfaces
             new_name = settings.get('name')
             if new_name:
+                if len(new_name) >= IF_NAMESIZE:
+                    logging.warning('Interface name {} is too long. {} will not be renamed'.format(new_name, iface))
+                    continue
                 if iface in devices and new_name in devices_after_udev:
                     logging.debug('Interface rename {} -> {} already happened.'.format(iface, new_name))
                     continue  # re-name already happened via 'udevadm test'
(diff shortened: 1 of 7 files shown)


slyon (approved):
Thanks! LGTM.

I agree we should only print a warning for now. As a hard failure could break the netplan generator at boot time and leave people without any network connectivity.

I've added a comment about this in the code and adopted `nm.c` to make use of the same validation logic, instead of using an undocumented constant there.

I'll also rebase and adopt to the `netplan_error_clear()` rename from #317

slyon:
Afterthought: Actually, there might be use cases for longer interface names, using "alternative names", once those are properly supported by netplan, cf.: https://github.com/canonical/netplan/pull/292

PR #314, opened 2023-01-18
tests: mock calls to systemctl

daniloegea:
Some tests were failing in sbuild due to calls to systemctl. It would check if networkd was running and try to start it. In my system the same happens except these tests get stuck asking for my password if I stop networkd.

## Checklist


slyon (approved):
Thanks, lgtm.

PR #317, opened 2023-01-19
api:util: rename netplan_error_clear (from _free) & adopt to nullify

slyon:
## Description
api:util: rename netplan_error_clear (from _free) & adopt to nullify

The spec was updated accordingly.
As discussed during the last netplan sync meeting

## Checklist


slyon:
This function is not (currently) used in our NetworkManager integration patch, so no need to adopt there.

daniloegea (approved):
looks good

PR #318, opened 2023-01-26
request for comment: specialized schema parse errors

blackboxsw:
## Description
@slyon thanks for the jira ping. 
Request for comment as to whether this is the right approach to take to get more detailed schema errors (we could alternatively look to unique err.contents.code differences from the C library but we'd also still have to 'unpack' some encoded line/column and path values from the structured error messages in python.

Cloud-int would like to have specialized exception types from libnetplan for schema errors. The specialization needed would be attributes on the Exception raised which point to the offending config file path, the marked error line number and column  number.

It seems that all libnetplan errors are generalized from the GERROR  raised by the C code. If we can either specialize the err.contents.code for different types or errors, or allow for message format parsing in [libnetplan:_checked_lib_call](https://github.com/canonical/netplan/blob/main/netplan/libnetplan.py#L79) then cloud-init error handling and reporting from tooling can better know how to handle schema errors vs other errors raised by netplan.

```bash
mkdir /tmp/netplan
cat > /tmp/new.yaml <<EOF
network: bogus
EOF

cat > validate_bogus_netplan.py <<EOF
from netplan import libnetplan

ROOT_DIR="/tmp/netplan"
def doit():
    parser = libnetplan.Parser()
    parser.load_yaml_hierarchy(ROOT_DIR)
    try:
        parser.load_yaml("/tmp/new.yaml")
    except LibNetplanException as e: # cloud-init wants a LibNetplanSchemaException if possible
        print(dir(e))   # cloud-init doesn't want to parse the structured e.message to tease out line and column info

if __name__ == "__main__":
    doit()
EOF

PYTHONPATH=/usr/share/netplan python3 validate_bogus_netplan.py

```

## Checklist

### DIFF
--- a/netplan/libnetplan.py
+++ b/netplan/libnetplan.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
 import tempfile
 import logging
 import ctypes
@@ -26,6 +27,21 @@ from typing import List, Union, IO
 
 class LibNetplanException(Exception):
     pass
+
+
+SCHEMA_ERROR_MESSAGE_REGEX = (
+    r'(?P<file_path>.*):(?P<error_line>\d+):(?<error_col>\d+): (?P<message>.*)'
+)
+
+class LibNetplanSchemaValidationException(LibNetplanException):
+
+    def __init__(self, message):
+        schema_error = re.match(SCHEMA_ERROR_MSG_REGEX, message).groupdict()
+        schema_error.groupdict()
+        self.file_path = schema_error["file_path"]
+        self.error_line = schema_error["error_line"]
+        self.error_col = schema_error["error_col"]
+        super().__init__(schema_error["message"])
 
 
 class _GError(ctypes.Structure):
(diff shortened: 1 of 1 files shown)


netplan/libnetplan.py
```diff
@@ -72,11 +88,22 @@ def _string_realloc_call_no_error(function):
...
+}
+
 def _checked_lib_call(fn, *args):
     err = ctypes.POINTER(_GError)()
     ret = bool(fn(*args, ctypes.byref(err)))
     if not ret:
-        raise LibNetplanException(err.contents.message.decode('utf-8'))
+        exception_cls = LibNetplanException
+        for error_format, error_type in GERROR_MESSAGE_FORMAT_EXCEPTION_MAP:
+            if re.match(error_format, err.contents.message.decode('utf-8')):
```
blackboxsw, line 103: Alternatively, we could raise different `err.contents.code` values to represent different exception types because regex matching of `err.contents.message` format matching could be brittle if message format changes or with internationalization of error messages at some point.

slyon:
Closing this in favor of #334 (which merged most this PR's concepts).

@blackboxsw We'd be glad for your feedback/review on that follow-up PR, when you find some time.

PR #320, opened 2023-01-31
parser: check for route duplicates (LP: #2003061)

daniloegea:
We currently check if the number of routes we have in the list is the same as the number of entries in the yaml structure. This cause problems when routes come from different files, see LP#2003061. In this case, the second route is not added to the netdef.

This patch changes the logic to actually check if the route being processed is already in the list. For the uniqueness check, it compares the route IP family, table ID, metric, from, to and via fields.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2003061

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -871,3 +871,33 @@ netplan_state_iterator_has_next(const NetplanStateIterator* iter)
         return FALSE;
     return _iter->next != NULL;
 }
+
+/*
+ * Returns true if a route already exists in the netdef routes list.
+ *
+ * We consider a route a duplicate if it is in the same table, has the same metric,
+ * src, to and via values.
+ *
+ * XXX: in the future we could add a route "key" to a hash set so this verification could
+ * be done faster.
+ */
+gboolean
+is_route_present(const NetplanNetDefinition* netdef, const NetplanIPRoute* route)
+{
+
+    const GArray* routes = netdef->routes;
+
+    for (int i = 0; i < routes->len; i++) {
+        const NetplanIPRoute* entry = g_array_index(routes, NetplanIPRoute*, i);
+        if (
+                entry->table == route->table &&
+                entry->metric == route->metric &&
+                g_strcmp0(entry->from, route->from) == 0 &&
+                g_strcmp0(entry->to, route->to) == 0 &&
+                g_strcmp0(entry->via, route->via) == 0
+           )
+            return TRUE;
+    }
+
+    return FALSE;
+}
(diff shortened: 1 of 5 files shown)


src/util.c
```diff
@@ -871,3 +871,33 @@ netplan_state_iterator_has_next(const NetplanStateIterator* iter)
...
+
+    for (int i = 0; i < routes->len; i++) {
+        const NetplanIPRoute* entry = g_array_index(routes, NetplanIPRoute*, i);
+        if (
+                entry->table == route->table &&
+                entry->metric == route->metric &&
+                g_strcmp0(entry->from, route->from) == 0 &&
+                g_strcmp0(entry->to, route->to) == 0 &&
+                g_strcmp0(entry->via, route->via) == 0
+           )
```
slyon, line 898: Beware that `->to` could be `default` or `0.0.0.0/0` / `::/0` which is logically the same and should fail this check.
We might want to compare `->family`, too. (IPv4 vs IPv6)

Furthermore, there can be routes which don't have the "via" field set... (e.g. when using "scope: link"), so I feel like we need to think a bit more about how to properly compare routes.

tests/generator/test_routing.py
```diff
@@ -1423,4 +1423,164 @@ def test_default_metric_v6(self):
 method=auto
 ip6-privacy=0
 route-metric=5050
+'''})
+
+    def test_add_routes_to_different_tables_from_multiple_files(self):
```
slyon, line 1428: I think we should make some use of the `ND_*` templates from `generator/base.py` here, e.g. ND_DHCP or ND_VLAN, to avoid repeating ourselves too much. Maybe concatenate or str.replace()  relevant pieces of the string as needed.

slyon (changes requested):
Thanks, I like this approach. But I think we need to work a bit more on the route matching logic. Everything else are just small suggestions inline.

While on it, we might consider refactoring some of this code, by introducing a `reset_route` helper (similar to `reset_ip_rule`). Also, what about `handle_ip_roules` in general? Wouldn't the routing-policy handling have the same problem of duplicates?

src/util.c
daniloegea: Ack. I tried to handle these cases. For fields that might be NULL, g_strcmp0() will handle it.

slyon (approved):
Thanks, lgtm. The refactoring and potential improvement of ip_rules can be done at later point in time. Let's keep this fix isolated.

PR #321, opened 2023-01-31
Memory leak action

daniloegea:
## Description

Add an script that can be used to run the netplan binaries and check for memory issues and a github action to call this script.

I added it to a directory called "tools" because it can be useful during development.

## Checklist

### DIFF
--- a/tools/run_asan.sh
+++ b/tools/run_asan.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -e
+set -x
+
+BUILDDIR="_leakcheckbuild"
+
+meson setup ${BUILDDIR} -Db_sanitize=address,undefined
+meson compile -C ${BUILDDIR} --verbose
+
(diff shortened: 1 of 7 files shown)


tools/run_asan.sh
```diff
@@ -0,0 +1,26 @@
...
+do
+    ./${test}
+done
+
+mkdir -p ${BUILDDIR}/fakeroot/{etc/netplan,run}
+export LD_LIBRARY_PATH="${BUILDDIR}/src"
+
+for yaml in examples/*.yaml
+do
+    chmod 600 ${yaml}
```
slyon, line 22: We shouldn't be modifying the file permissions inside the source directory, but rather first copy it over into the fakeroot and change permissions there before execution of `generate`

slyon (approved):
Thanks, lgtm. Just one small nitpick inline, that we should fix before merging.

slyon (approved):
+1

PR #322, opened 2023-02-01
API: update netplan_delete_connection() to avoid spawning another process

slyon:
## Description
This function has been calling into the python 'netplan set' command for historical reasons. We can now stay internal to libnetplan and execute the corresponding libnetplan functions directly.

Basically, we're now implementing inside this function in `util.c` what has previously been executed via `netplan/cli/commands/set.py`

## Checklist

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -21,6 +21,7 @@
 #include <fnmatch.h>
 #include <errno.h>
 #include <string.h>
+#include <sys/mman.h>
 
 #include <glib.h>
 #include <glib/gprintf.h>
@@ -39,6 +40,8 @@ wifi_frequency_24;
 
 NETPLAN_ABI GHashTable*
 wifi_frequency_5;
+
+const gchar* FALLBACK_FILENAME = "70-netplan-set.yaml";
 
 typedef struct netplan_state_iterator RealStateIter;
 /**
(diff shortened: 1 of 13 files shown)


src/util.c
```diff
@@ -487,55 +489,88 @@ systemd_escape(char* string)
...
+    /* Create a new parser & state to hold our output YAML, ignoring the to be
+     * deleted Netdef from the patch */
+    output_parser = netplan_parser_new();
+    output_state = netplan_state_new();
 
-    if (!np_state->netdefs) {
+    rewind(patch);
+    if (   !netplan_parser_load_nullable_fields(output_parser, patch_fd, &error)
+        || !netplan_parser_load_yaml_hierarchy(output_parser, rootdir, &error)) {
         // LCOV_EXCL_START
```
sil2100, line 542: Just out of curiosity: why are we excluding those multiple error cases from coverage? Are they being tested in integration testing?

sil2100 (commented):
Looks good from my POV, but then again I don't really know the exact logic for `netplan set` anymore so I'm not leaving any Approval/Disapproval. Is the intent to move away from calling `netplan set` from inside the compiled netplan for various cases? Since I see dbus.c also uses `netplan set`.

I took a look on how `netplan set` is written and the workflow here at least makes sense.

src/util.c
slyon: Those are things which are really hard to test. They are not being tested in integration testing either. It's basically file system errors or errors in the emitter and/or parser of libyaml, which we cannot easily influence.

I guess I could mock some of the `netplan_` functions, when adding new unit tests as part of our new C based cmoka tests in `tests/ctest/`.. Maybe I should do this to test more of the error reporting cases.

slyon:
> Is the intent to move away from calling `netplan set` from inside the compiled netplan for various cases? Since I see dbus.c also uses `netplan set`.

Yes, indeed. We want to avoid spawning new processes from inside compiled (lib-)netplan. In the past all of `netplan set` including it's YAML parser was implemented in Python, so we just called into that python code from the libnetplan API. Now the parser is actually implemented in netplan itself, so a call of this library function would call into the python code, which would in turn call back into libnetplan. We want to avoid this libnetplan -> python -> libnetplan back and forth.

`dbus.c` is a mess and should be refactored eventually consuming all of the new API. But it's also a stand-alone binary, so not that much of a problem as calling back and forth into the library.

src/util.c
```diff
@@ -487,55 +489,88 @@ systemd_escape(char* string)
...
+        || !netplan_state_import_parser_results(input_state, input_parser, &error)) {
+        g_warning("netplan_delete_connection: Cannot parse input: %s\n", error->message);
+        ret = FALSE;
+        goto cleanup;
+    }
+
+    /* find specified netdef in input state */
+    nd = netplan_state_get_netdef(input_state, id);
+    if (!nd) {
+        g_fprintf(stderr, "netplan_delete_connection: Cannot delete %s, does not exist.", id);
```
ogayot, line 514: You seem to have mix of calls to g_fprintf() with patterns [having / having no] trailing "\n".
I think at least for this occurrence, it is missing the "\n" (unless `g_fprintf` adds it automatically: I'm not familiar with glib).

I can't speak for the other occurrences where the pattern ends in "%s". You might expect that the parameter (i.e. error message) contains the the final "\n".

Also, you might want to change `"Cannot delete %s, does not exist/"` by `"Cannot delete %s: does not exist."` but this is beyond nitpick :)

src/util.c
```diff
@@ -487,55 +489,88 @@ systemd_escape(char* string)
...
+    int patch_fd = fileno(patch);
+    if (!netplan_util_create_yaml_patch(yaml_path, "NULL", patch_fd, &error)) {
         // LCOV_EXCL_START
-        g_fprintf(stderr, "%s\n", error->message);
+        g_fprintf(stderr, "netplan_delete_connection: Cannot create YAML patch: %s", error->message);
         ret = FALSE;
         goto cleanup;
         // LCOV_EXCL_STOP
     }
+    fflush(patch);
```
ogayot, line 532: I don't know if I would feel comfortable using C streams (`FILE *`) operations after passing around the underlying fd to a function.

But if you have experience with it, go for it :)

Otherwise, I would recommend dropping fflush() and using lseek() instead of rewind()

daniloegea: Is the fflush() really needed here? I think that at this point there is nothing in the stream buffer as the writer only used the file descriptor.

slyon: ACK. `netplan_util_create_yaml_patch()` is actually flushing the FD itself, so no need to do it here.

src/util.c
slyon: ACK. I was experimenting a bit with `g_warning()`/`g_error()`/g_fprintf()`, but `g_error()` actually terminates the application, which is not directly what we want here. So this was a left-over. Fixed now.

src/util.c
```diff
@@ -487,55 +489,87 @@ systemd_escape(char* string)
...
+        g_fprintf(stderr, "netplan_delete_connection: Cannot delete %s, does not exist.\n", id);
+        ret = FALSE;
+        goto cleanup;
+    }
+
+    /* Build up a tab-separated YAML path for this Netdef (e.g. network.ethernets.eth0=...) */
+    yaml_path = g_strdup_printf("network\t%s\t%s", netplan_def_type_name(nd->type), id);
+
+    /* create a temporary file in memory, to hold our YAML patch */
+    patch = tmpfile();
```
schopin-pro, line 523: issue: `tmpfile()` is disk-backed

I love the approach, but `tmpfile` can (and will, with glibc) create an actual file on the filesystem. You might want to have a look at `memfd_create` instead.

schopin-pro (commented):
I'm not approving because of the small `tmpfile` issue, but overall I really like the patch.

In an ideal world we'd have a `netplan_state_delete_netdef` function that wouldn't require us to craft a YAML payload, but that's a bit further away ;)

src/util.c
slyon: Thanks, that's a good suggestion! The comment shows that it was the original intention, too :)

I had to bump `_X_OPEN_SOURCE=700` up to `_GNU_SOURCE` to enable the `memfd_*` functions, which are part of the GNU extensions and work around some conflicts with `__USE_MISC` (needed for `GLOB_BRACE` from glob.h)

src/util.c
```diff
@@ -487,55 +490,93 @@ systemd_escape(char* string)
...
-    ret = g_spawn_sync(NULL, (gchar**)argv, NULL, 0, NULL, NULL, NULL, NULL, NULL, NULL);
 
 cleanup:
-    if (npp) netplan_parser_clear(&npp);
-    if (np_state) netplan_state_clear(&np_state);
+    if (input_parser) netplan_parser_clear(&input_parser);
+    if (input_state) netplan_state_clear(&input_state);
+    if (output_parser) netplan_parser_clear(&output_parser);
+    if (output_state) netplan_state_clear(&output_state);
+    if (patch_fd > 0) close(patch_fd);
```
daniloegea, line 579: It should be `patch_fd >= 0`. If the file descriptor 0 doesn't exist when memfd_create() is called, 0 will be used and the function would leak the descriptor (and memory).

daniloegea (changes requested):
It looks good. Only requesting a minor change to avoid a leak when stdin is closed.

daniloegea (approved):
Looks good. Just the change to _GNU_SOURCE could use another pair of eyes.

Regarding the ABI breakage, I tested Network Manager built with an old version of libnetplan and it still works after this change (without rebuilding it I mean)

schopin-pro (approved):
LGTM.

Regarding the ABI changes, since we have a tight dependency I'm OK with it. However, to be clear: this kind of ABI break wouldn't result in the binary refusing to start, and instead would lead to fun things like stack corruption ;).

I'm guessing last time I bumped the XOPEN_SOURCE value it was an ABI break too, huh? Live & learn...

slyon:
Thank you for the feedback @daniloegea and @schopin-pro! Let's get this merged.

PR #323, opened 2023-02-14
Fix FTBFS on Fedora and refresh RPM packaging

Conan-Kudo:
This pull request fixes Netplan to build on Fedora Linux and updates the RPM packaging to use the Meson build.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -36,7 +36,7 @@ executable(
     install: true)
 install_symlink(
     'netplan',
-    pointing_to: join_paths('..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
+    pointing_to: join_paths('..', '..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
     install_dir: systemd_generator_dir)
 # Install this symlink for legacy reasons, see netplan/cli/utils.py: get_generator_path()
 install_symlink(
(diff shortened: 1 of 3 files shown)


slyon (commented):
Thank you! I need to double check the systemd usrmerge situation in Debian/Ubuntu wrt. that meson.build change. Otherwise this LGTM!

src/meson.build
```diff
@@ -36,7 +36,7 @@ executable(
     install: true)
 install_symlink(
     'netplan',
-    pointing_to: join_paths('..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
+    pointing_to: join_paths('..', '..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
     install_dir: systemd_generator_dir)
```
xnox, line 40: that's not quite right. as systemd_generator_dir can and is of different length on different platforms.

maybe install_symlink accepts absolute locations which then can be "optimized" automatically, or later on by packaging itself? (i.e. in debian debhelper fixes up absolute symlinks to relative ones) Doing absolute target is better here imho.

Conan-Kudo: I'll adjust this to use absolute paths.

slyon:
Ugh.. apparently meson is not willing to install a dangling symlink, so I needed to revert this https://github.com/canonical/netplan/commit/ea5530c9664c69f7c25d09cc10c63f5ed7a89990 and we'll need come up with something better.

Conan-Kudo:
You might have to use a custom handler for installing symlinks instead of Meson's own thing. There's too many papercuts with Meson's `install_symlink()` method.

slyon:
Yeah... I'm currently investigating a few options/workarrounds, like `custom_target` or `add_install_script`, as described in https://github.com/mesonbuild/meson/issues/1602

PR #327, opened 2023-02-27
meson: use meson-make-symlink.sh helper script instead of install_symlink()

slyon:
## Description
meson: use meson-make-symlink.sh helper script instead of install_symlink()

This will allow to dynamically adopt the symlinks for usrmerged and non-usrmerged installations, especially:
root@ll:~# pkgconf --variable=systemdsystemgeneratordir systemd /lib/systemd/system-generators

[root@f37 ~]# pkgconf --variable=systemdsystemgeneratordir systemd /usr/lib/systemd/system-generators

Helper script inspired by util-linux 238.1-5 (tools/meson-make-symlink.sh)

Related to #323 

## Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,6 +15,8 @@ yaml = dependency('yaml-0.1')
 uuid = dependency('uuid')
 libsystemd = dependency('libsystemd')
 
+meson_make_symlink = meson.current_source_dir() + '/tools/meson-make-symlink.sh'
+
 systemd = dependency('systemd')
 completions = dependency('bash-completion')
 systemd_generator_dir = systemd.get_variable(pkgconfig: 'systemdsystemgeneratordir')
--- a/src/meson.build
+++ b/src/meson.build
@@ -34,10 +34,9 @@ executable(
     dependencies: [glib, gio, yaml, uuid],
     install_dir: libexec_netplan,
     install: true)
-install_symlink(
-    'netplan',
-    pointing_to: join_paths('..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
-    install_dir: systemd_generator_dir)
+meson.add_install_script(meson_make_symlink,
+    join_paths(get_option('prefix'), libexec_netplan, 'generate'),
+    join_paths(systemd_generator_dir, 'netplan'))
 # Install this symlink for legacy reasons, see netplan/cli/utils.py: get_generator_path()
 install_symlink(
     'generate',
--- a/tools/meson-make-symlink.sh
+++ b/tools/meson-make-symlink.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -eu
+
+# this is needed mostly because $DESTDIR is provided as a variable,
+# and we need to create the target directory...
+
+mkdir -vp "$(dirname "${DESTDIR:-}$2")"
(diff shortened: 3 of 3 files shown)


slyon:
This seems to be working for both cases:
```
$ tree tmproot/
tmproot/                                                   
├── lib                                                    
│   └── systemd                                            
│       └── system-generators
│           └── netplan -> ../../../usr/libexec/netplan/generate
└── usr
    ├── libexec
    │   └── netplan
    │       ├── generate

$ tree tmproot/
tmproot/                                                   
└── usr                                                    
    ├── lib
    │   ├── systemd
    │   │   └── system-generators
    │   │       └── netplan -> ../../../libexec/netplan/generate
    │   └── [...]
    ├── libexec
    │   └── netplan
    │       ├── generate
```

slyon:
I've also talked to our systemd maintainer, and we will coordinate with Debian to adopt the systemd packaging to install stuff in `/usr/lib/systemd/...` instead of `/lib/systemd/...` (+ adopting corresponding pkgconf files).

slyon:
cc @Conan-Kudo @xnox WDYT about this?

Conan-Kudo (approved):

Conan-Kudo:
> I've also talked to our systemd maintainer, and we will coordinate with Debian to adopt the systemd packaging to install stuff in `/usr/lib/systemd/...` instead of `/lib/systemd/...` (+ adopting corresponding pkgconf files).

This needs to be done anyway, since systemd is ripping out support for non-usrmerged systems soon. It'd be great to see this in place for Lunar at least.

xnox (approved):
yeah, --relative is best, as this avoids the need for distro packaging tools to fix things up later, like rpm scriplets do, or like dh_link in debian world.

slyon:
> This needs to be done anyway, since systemd is ripping out support for non-usrmerged systems soon. It'd be great to see this in place for Lunar at least.

Agreed. Lunar is post-FeatureFreeze already, so unfortunately I doubt the systemd change will still be landed in Lunar. But should be in Lunar+1

Conan-Kudo:
> > This needs to be done anyway, since systemd is ripping out support for non-usrmerged systems soon. It'd be great to see this in place for Lunar at least.
> 
> Agreed. Lunar is post-FeatureFreeze already, so unfortunately I doubt the systemd change will still be landed in Lunar. But should be in Lunar+1

I think this is honestly worthy of an exception, but my voice doesn't really matter here...

slyon:
> > > This needs to be done anyway, since systemd is ripping out support for non-usrmerged systems soon. It'd be great to see this in place for Lunar at least.
> > 
> > 
> > Agreed. Lunar is post-FeatureFreeze already, so unfortunately I doubt the systemd change will still be landed in Lunar. But should be in Lunar+1
> 
> I think this is honestly worthy of an exception, but my voice doesn't really matter here...

Your voice and input is highly appreciated! I've been told that some discussion around this usrmerge change has been happening in IRC #debian-systemd recently. But as both Debian and Ubuntu are currently in freeze, I assume the maintainers won't land such impactful change anymore (it will require changes to many more packages, like Netplan, which use systemd's `pkgconf` data). I'll try to re-ping people about, tho, and make them aware of this discussion.

PR #330, opened 2023-02-28
tests: add new spread based snapd integration test

mvo5:
## Description

This is a first draft of adding integration tests for netplan with snapd and dbus. It uses spread and the LXD backend to run inside the github runners. A first test is designed to do basic interactions between netplan and snapd.

I put this into "draft" mode because I ran into the following issues:
1. following the instruction to make crashes meson inside a 22.04 container (https://paste.ubuntu.com/p/qRnJvjyddN/)
2. when commenting out the testing the install step fails with: 
```
meson install -C build --destdir=/
...
Installing /home/tests/netplan.completions to //usr/share/bash-completion/completions

ERROR: Tried to install symlink to missing file //lib/netplan/../../usr/libexec/netplan/generate

```
I suspect (2) is a result in the most recent https://github.com/canonical/netplan/pull/323 ?

I also wonder if the build/install/build-dep step could be simplified by getting the debian packaging from somewhere but I could not figure out where it comes from (sorry for that and hints/ideas welcome).

## Testing locally via spread

```
$ GOBIN=~/go/bin go install github.com/snapcore/spread/cmd/spread@latest
$ ~/go/bin/spread -v -debug lxd:
```

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -39,8 +39,6 @@ meson.add_install_script(meson_make_symlink,
     join_paths(systemd_generator_dir, 'netplan'))
 # FIXME: Drop legacy symlink after 0.107 is released:
 # It's only around for legacy reasons, see netplan/cli/utils.py: get_generator_path()
-install_symlink(
-    'generate',
-    pointing_to: join_paths('..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
-    install_dir: join_paths('/', 'lib', 'netplan'))
-
+meson.add_install_script(meson_make_symlink,
+    join_paths(get_option('prefix'), libexec_netplan, 'generate'),
+    join_paths('/', 'lib', 'netplan', 'generate'))
--- a/tests/spread/snapd-integration/task.yaml
+++ b/tests/spread/snapd-integration/task.yaml
@@ -1,0 +1,47 @@
+summary: Check that the snapd/netplan integration works
+
+prepare: |
+    # put snapd in debug mode for good logs
+    mkdir -p /etc/systemd/system/snapd.service.d
+    cat <<EOF > /etc/systemd/system/snapd.service.d/local.conf
+    [Service]
+    Environment=SNAPD_DEBUG=1
+    EOF
+    systemctl daemon-reload
+    # install snapd if needed (should be installed on most lxd images)
+    if ! dpkg -l snapd >/dev/null; then
+        apt install -y snapd
+        echo "apt autoremove -y snapd" >> defered.cleanup
+    fi
+    # fake running on core (netplan only enabled on ubuntu core)
+    # FIXME: find a more offical way, this if fugly
+    cp /etc/os-release /etc/os-release.save
+    sed -i s/ID=ubuntu/ID=ubuntu-core/ /etc/os-release
+    M=/var/lib/snapd/assertions/asserts-v0/model/16/generic/generic-classic/active
+    cp "$M" model.save
+    sed -i 's/classic: true/architecture: amd64\ngadget: pc\nkernel: pc-kernel/g' "$M"
+    # restart snapd
+    systemctl restart snapd
+
+restore: |
+    mv /etc/os-release.save /etc/os-release
+    mv model.save /var/lib/snapd/assertions/asserts-v0/model/16/generic/generic-classic/active
+    sudo systemctl restart snapd
+    if [ -e defered.cleanup ]; then
+        sh -ex defered.cleanup
+    fi
+    rm -f model.*
+
+debug: |
+    journalctl -u snapd
+    snap get system -d
+    snap version
+
+execute: |
+  # see the basics work
+  netplan get bridges.br54.dhcp4 | MATCH true
+  snap get system system.network.netplan.network.bridges.br54.dhcp4 | MATCH true
+  # and updating netplan works
+  snap set system system.network.netplan.network.bridges.br54.dhcp4=false
+  netplan get bridges.br54.dhcp4 | MATCH false
+  snap get system system.network.netplan.network.bridges.br54.dhcp4 | MATCH false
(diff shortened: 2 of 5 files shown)


slyon:
FYI: packaging is done in: https://git.launchpad.net/~ubuntu-core-dev/netplan/+git/ubuntu

tests/spread/snapd-integration/task.yaml
```diff
@@ -0,0 +1,48 @@
...
+    EOF
+    systemctl daemon-reload
+    # install snapd if needed (should be installed on most lxd images)
+    if ! dpkg -l snapd >/dev/null; then
+        apt install -y snapd
+        echo "apt autoremove -y snapd" >> defered.cleanup
+    fi
+    # fake running on core (netplan only enabled on ubuntu core)
+    # FIXME: find a more offical way, this if fugly
+    cp /etc/os-release /etc/os-release.save
```
mvo5, line 18: This needs something like https://github.com/snapcore/snapd/pull/12607 to really work, this hack will just break

mvo5:
Hrm, also slightly sad that the new "spread" workflow seems to not get picked up :/

mvo5:
Fwiw, AFAICT the dbus test in this repo shows the issue I mentioned in the  SRU bugreport that the dbus reply is empty:
```
...
2023-02-28T19:44:22.7009429Z + dbus-send --print-reply --system --type=method_call --dest=io.netplan.Netplan /io/netplan/Netplan/config/OFRS01 io.netplan.Netplan.Config.Get
2023-02-28T19:44:22.7009921Z grep error: pattern not found, got:
2023-02-28T19:44:22.7010406Z method return time=1677613462.696987 sender=:1.23 -> destination=:1.24 serial=7 reply_serial=2
2023-02-28T19:44:22.7010926Z    string ""
...
```
and indeed, running `spread -debug` shows:
```
lxd:ubuntu-22.04 .../tests/spread/snapd-integration# busctl --system call io.netplan.Netplan /io/netplan/Netplan/config/VRAU01 io.netplan.Netplan.Config Get
s ""
lxd:ubuntu-22.04 .../tests/spread/snapd-integration# netplan get

** (process:3410): WARNING **: 20:03:14.464: Permissions for /etc/netplan/0-snapd-defaults.yaml are too open. Netplan configuration should NOT be accessible by others.

** (process:3410): WARNING **: 20:03:14.464: Permissions for /etc/netplan/50-cloud-init.yaml are too open. Netplan configuration should NOT be accessible by others.
network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
  bridges:
    br54:
      dhcp4: true
```

src/meson.build
```diff
@@ -41,6 +41,6 @@ meson.add_install_script(meson_make_symlink,
 # It's only around for legacy reasons, see netplan/cli/utils.py: get_generator_path()
 install_symlink(
     'generate',
-    pointing_to: join_paths('..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
+    pointing_to: join_paths('..', '..', '..') + join_paths(get_option('prefix'), libexec_netplan, 'generate'),
```
mvo5, line 44: This change fixed the "make install" fwiw for me

slyon:
FYI: I rebased this on top of #331 to see if the spread test is now passing.

mvo5:
> FYI: I rebased this on top of #331 to see if the spread test is now passing.

And it looks like it did! Nice job, thanks for fixing the issue :-D !

mvo5:
I pushed a small tweak now to properly setup/restore the state of the /etc/netplan dir (which makes the need to use multiple test bridge interfaces go away as spread will never run the same test in parallel in the same VM/container, when it runs in parallel it does so on multiple VMs/containers but not in the same).

I would love to know if you folks are interested in taking this PR (in principal, happy to tweak as needed) or if you prefer not to. If this has a chance to land I can https://github.com/snapcore/snapd/pull/12607 to make the setup in the snapd integration test much nicer and potentially sketch how autopktest can be integrated with spread tests.

While autopkgtest and spread are broadly doing the same our experience in snapd is that spread scales better with the (buildin) concept of suits and the more natural "prepare/execute/restore" patterns that one knows from regular unit test writing. It also scales better when many tests are used as it will use as many VMs/containers as provided (initially not a concern here though as the tests run in 2min right now ;)

src/meson.build
slyon: I'm a bit worried about this change. It makes our autopkgtest GH action fail. We need to be careful with meson installation paths, as those can be different on Fedora/Debian/Ubuntu and might need to go hand-in-hand with packaging changes. Can we somehow avoid this change?

Maybe this helps: https://github.com/canonical/netplan/commit/2cded2f2673ce90f1585c9f6f9bb16ca7d81e7a6 (i.e. rebase on top of our recent `main` branch). See https://github.com/canonical/netplan/pull/327 for context.

slyon (commented):
Hey @mvo5  Sorry for taking too long to get through my PR review backlog..
Yes, I'm very much interested in landing this! Even as a spread test in GitHub actions, as proposed. We might want to migrate that later on into an autopkgtest, so it's also run at autopkgtest.ubuntu.com (and other instances, like DebCI). 

But landing this spread test should be a good first step. It could also help to prepare infrastructure/tooling for further spread test integration, like https://github.com/snapcore/network-manager-snap/pull/15

The only thing I'm worried about is the `src/meson.build` change (see my inline comment), otherwise this LGTM!

slyon:
Actually, I think something like this should do the trick:
```
meson.add_install_script(meson_make_symlink,
    join_paths(get_option('prefix'), libexec_netplan, 'generate'),
    join_paths('/', 'lib', 'netplan', 'generate'))
```

I rebased and force pushed to your branch. Let's see the CI results.

slyon (approved):
This is providing value as-is already. So let's get it merged. We can still improve upon this once snapd landed corresponding upstream changes. Or we've prepared an autopkgtest to run those spread tests accordingly.

PR #333, opened 2023-03-07
Netplan docs rework

daniloegea:
## Description

I tried to follow the ideas from [diataxis.fr](https://diataxis.fr/) and adapted parts of our documentation to the framework.

- Tutorial: I created a tutorial to guide the user that has very little knowledge about Netplan through a number of tasks they can execute in a virtual machine. Each task increments the previous one and introduce more things one can do with Netplan.
   - Should I introduce `netplan status` as well? I didn't use it because it's only available in Lunar at the moment... 
- How-tos: I changed the sets of examples to look more like how-tos. I removed or simplified some of the explanations to go straight to the point of the problem we are trying to solve. I also added a few more how-tos to the document. The indentation was normalized to 2 spaces.
- Reference: I added a new initial section to the document that shows the general structure of a Netplan YAML file and changed the examples so they will be complete configuration files and the reader can easily see where each settings fits in the configuration. 

## Checklist


doc/netplan-yaml.md
```diff
@@ -2,8 +2,54 @@
...
+  modems: MAPPING
+  nm-devices: MAPPING
+  tunnels: MAPPING
+  vlans: MAPPING
+  vrfs: MAPPING
+  wifis: MAPPING
+```
+
+- **version** (number)
+  > Defines what version of the configuration format is used. The only value supported is `2`. Defaults to `2` is not defined.
```
slyon, line 25: This and the other new quote sections (`> `) does not render correctly in the `netplan.5` man page. (see `make && man _build/doc/netplan.5`). I think we need some newlines around the quote sections.

doc/netplan-yaml.md
```diff
@@ -2,8 +2,54 @@
 title: "YAML configuration"
 ---
 
+## General configuration structure
```
slyon, line 5: There's already a "General structure" topic in the manpage header... Maybe we should rename this to something different, like "Top-level configuration structure", to avoid confusion?

doc/examples.md
```diff
@@ -1,154 +1,184 @@
-# Examples
+# How-tos
```
slyon, line 1: I wonder if we should split out the differnt "How-to guides" into separate files.. but this will lead to plenty of tiny files, which are hard to maintain. But it would be nice if we could somehow link each individual Howto in the pimary navigation, similar to "Use DBus config API" and "Desktop integration".

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+
+Available commands:
+
+    help      Show this help message
+    apply     Apply current netplan config to running system
+    generate  Generate backend specific configuration files from /etc/netplan/*.yaml
+    get       Get a setting by specifying a nested key like "ethernets.eth0.addresses", or "all"
+    info      Show available features
+    ip        Retrieve IP information from the system
+    set       Add new setting by specifying a dotted key=value pair like ethernets.eth0.dhcp4=true
```
slyon, line 127: I'm missing the new `netplan status` here (and in the overall tutorial).

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+
+```
+default via 172.16.0.254 proto static
+172.16.0.0/24 proto kernel scope link src 172.16.0.1
+```
+
+And check the DNS configuration:
+
+```bash
+ resolvectl status enp6s0
```
slyon, line 469: Maybe we could promote `netplan status` here instead?

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+
+## Pre-requisites
+
+In order to do the exercises yourself you will need a virtual machine, preferably running Ubuntu. In this tutorial, we will use LXD to create virtual networks and launch virtual machines. Feel free to use a cloud instance or a different hypervisor. As long as you can achieve the same results, you should be fine. If you're going to use your own desktop/laptop system, some of the exercises might interrupt your network connectivity.
+
+If you already have a setup where you can do the exercises you can just skip this section.
+
+### Setting up the environment
+
+You can follow the steps below to install and create a basic LXD configuration you can use to launch virtual machines. For more information about LXD, please visit: https://linuxcontainers.org/lxd/introduction/
```
slyon, line 14: The URL should become a real markdown hyperlink, so that it becomes clickable.

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+```
+Network netplanbr0 created
+```
+
+After executing the commands above you will have a usable LXD installation with a working network bridge.
+
+Now create a virtual machine called `netplan-lab0`:
+
+```bash
+lxc init --vm ubuntu:22.04 netplan-lab0
```
slyon, line 40: I wonder if we should use something newer here, like `ubuntu-daily:devel` to make the latest Netplan features (like "status" CLI) available.

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+And start your new VM:
+
+```bash
+lxc start netplan-lab0
+```
+
+Access your new VM using `lxc exec`:
+
+```bash
+lxc exec netplan-lab0 bash
```
slyon, line 66: nitpick: `lxc shell netplan-lab0` should do the trick, too.

Maybe we should also mention the possibilty of `lxc console netplan-lab0` should somebody break their access to the VM.

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
...
+    ethernets:
+        enp5s0:
+            dhcp4: true
+```
+
+This file was automatically generated by `cloud-init` when the system was initialized. As noted in the comments, changes to this file will not persist.
+
+## Enabling your second network interface with DHCP
+
+There are basically 2 ways to create or change netplan configuration: 1) using the `netplan set` command and 2) editing the YAML files manually.
```
slyon, line 182: nitpick: "1)" and "2)" should be rendered as an actuall ordered list in markdown.

doc/netplan-tutorial.md
```diff
@@ -0,0 +1,560 @@
+
+# Getting started with Netplan
```
slyon, line 2: praise: very nice tutorial! I wonder if we should extend it for some simple bridge and/or bonding case? (As this is a typical server usecase)

slyon (commented):
This is not yet a full review, but I had an initial look and left some inline comments.

I'd like to ask some real Diataxis expert like @tmihoc to do a full review.

The docs can be built and rendered locally from Danilo's branch, using `cd doc/ && make install && make run`

doc/netplan-tutorial.md
daniloegea: Yeah I almost included it (I added a note to the PR description). As we have plans to backport Netplan 0.106 I guess I should add it.

doc/netplan-tutorial.md
daniloegea: hmm I don't know, I think the focus should be stable releases, those are the ones users will install after all. We can update it to ubuntu:23.04 in a couple of months I guess. For now I think I will mention netplan status is only available in netplan 0.106.

doc/netplan-tutorial.md
daniloegea: hmm interesting, lxc shell doesn't show up in the list of available commands for me... but it works

doc/netplan-tutorial.md
daniloegea: Thanks! I added one more example using a bond interface.

doc/examples.md
```diff
@@ -1,154 +1,184 @@
...
-            dhcp4: true
+  version: 2
+  renderer: networkd
+  ethernets:
+    enp3s0:
+      dhcp4: true
 ```
 
-To instead set a static IP address, use the addresses key, which takes a list of (IPv4 or IPv6), addresses along with the subnet prefix length (e.g. /24). DNS information can be provided as well, and the gateway can be defined via a default route:
+# How to configure a static IP address on an interface
```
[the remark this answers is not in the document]
daniloegea: As they are listed one after another the "How to" gets very repetitive... I followed the naming conventions from here though https://diataxis.fr/how-to-guides/#pay-attention-to-naming

tmihoc (commented):
Hi! I think these are all great improvements, and I would publish them right away, using any comments to plan future work. 

The comments amoung to the following recommendations:

1. In general: Help the user distinguish visually etween main content and tangents (i.e., set off notes that give links to further information, or notes with caveats about edge-cases, etc.). 
2. In general: Help the user distinguish between content that pertains to your product and other types of content (e.g., `netplan` commands vs. `lxc` commands vs. general Unix commands etc.)
3. In general (though especially in the Tutorial / How-to guides): Avoid making the user look back, ahead, or sideways (try to stick to "below" and nothing else).
5. In a Tutorial / How-to guide, always keep the user in active mode. E.g., Use verbal (doc and) section titles ("Do x", not "Doing x"), where each title should be a step in the bigger story of the Tutorial / How-to guide.  
6. Distinguish between Tutorial and How-to guides / Referenece docs. To use a culinary metaphor, even if it will inevitably involve mention of ingredients / utensils (Reference material), and procedures (How-to material), the Tutorial must always stay focused on baking the cake. 
7. In Reference docs: (a) Add a doc for every Netplan command. (b) For docs focused on giving the spec of a yaml file, structure the doc around the top-level keys. 

Happy to clarify further.

tmihoc:
Hi! I think these are all great improvements, and I would publish them right away, using any comments to plan future work.

The comments amoung to the following recommendations:

In general: Help the user distinguish visually etween main content and tangents (i.e., set off notes that give links to further information, or notes with caveats about edge-cases, etc.).
In general: Help the user distinguish between content that pertains to your product and other types of content (e.g., netplan commands vs. lxc commands vs. general Unix commands etc.)
In general (though especially in the Tutorial / How-to guides): Avoid making the user look back, ahead, or sideways (try to stick to "below" and nothing else).
In a Tutorial / How-to guide, always keep the user in active mode. E.g., Use verbal (doc and) section titles ("Do x", not "Doing x"), where each title should be a step in the bigger story of the Tutorial / How-to guide.
Distinguish between Tutorial and How-to guides / Referenece docs. To use a culinary metaphor, even if it will inevitably involve mention of ingredients / utensils (Reference material), and procedures (How-to material), the Tutorial must always stay focused on baking the cake.
In Reference docs: (a) Add a doc for every Netplan command. (b) For docs focused on giving the spec of a yaml file, structure the doc around the top-level keys.
Happy to clarify further.

daniloegea:
I really appreciate the thorough review you provided, @tmihoc 

We'll try to adopt all your suggestions in our documentation!

slyon:
Thank you very much @tmihoc and @daniloegea for your input!

I marked the "little things" that Danilo already fixed as "resolved" and will merge the PR as-is. We'll then work on a follow-up PR to fix the bigger things separately.

PR #336, opened 2023-03-23
Use controlled execution environment, to avoid failure if PATH is unset (LP: #1959570)

slyon:
## Description
Use a controlled execution environment.
This is to allow execution in constrained environments, like early boot or
Ubuntu Core 20/22 where the PATH environment variable might not be set

Also, handle `nmcli`'s "NetworkManager not running" error gracefully.
Considering `/usr/bin/nmcli` for normal NetworkManager and `/snap/bin/nmcli -> /snap/bin/network-manager.nmcli` for NetworkManager snap as valid implementations.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1959570


netplan/cli/utils.py
```diff
@@ -40,16 +42,22 @@ def get_generator_path():
 
 
 def is_nm_snap_enabled():
-    return subprocess.call(['systemctl', '--quiet', 'is-enabled', NM_SNAP_SERVICE_NAME], stderr=subprocess.DEVNULL) == 0
+    env = dict(LC_ALL='C', PATH=os.environ.get('PATH', FALLBACK_PATH))
+    return subprocess.call(['systemctl', '--quiet', 'is-enabled', NM_SNAP_SERVICE_NAME],
+                           env=env, stderr=subprocess.DEVNULL) == 0
```
daniloegea, line 47: I'm wondering if it would suffice setting these variables only once via os.environ. Something like the below in the entry point of the Python code:

`os.environ.update({'LC_ALL': 'C', 'PATH': os.environ.get('PATH', '/usr/bin:/snap/bin')})`

daniloegea (commented):
Would it be enough settings the environment variables only once via `os.environ` in the entry point of the python code?

Makefile
```diff
@@ -19,6 +19,7 @@ clean:
 	rm -f *.gcda *.gcno generate.info
 	rm -f tests/ctests/*.gcda tests/ctests/*.gcno
 	rm -rf test-coverage .coverage coverage.xml
```
ogayot, line 21: nitpick: coverage.xml would be best suited here

netplan/cli/commands/status.py
```diff
@@ -157,9 +156,10 @@ def query_nm_ssid(self, con_name: str) -> str:
 
     def query_networkctl(self, ifname: str) -> str:
         output: str = None
+        env = dict(LC_ALL='C', PATH=os.environ.get('PATH', utils.FALLBACK_PATH))
         try:
             output = subprocess.check_output(['networkctl', 'status', ifname],
```
ogayot, line 161: Not relevant to this PR so feel free to ignore.
It's very uncommon but network interfaces having a name that starts with `-` is a thing.
I would recommend doing:
```python
['networkctl', 'status', '--', ifname]
```
just to avoid any problem.

```
$ ip link add -lo type dummy
$ networkctl status -lo
networkctl: invalid option -- 'o
$ networkctl status -- -lo
● 37: -lo                                                               
                     Link File: /usr/lib/systemd/network/99-default.link
[...]
```

netplan/cli/commands/apply.py
```diff
@@ -290,17 +295,25 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
...
+                # 'nmcli' could be /usr/bin/nmcli or
+                # /snap/bin/nmcli -> /snap/bin/network-manager.nmcli
                 cmd = ['nmcli', 'general', 'status']
-                for _ in range(20):
-                    if b'\nconnected' in subprocess.check_output(cmd, env=env):
+                # wait a bit for 'connected (site/local-only)' or
+                # 'connected' to appear in 'nmcli general' STATE
+                for _ in range(10):
+                    out = subprocess.run(cmd, capture_output=True,
+                                         universal_newlines=True, env=env)
```
ogayot, line 308: Up to you to use it here, not use it, or use it everywhere ; but it does improve readability IMO:

> Changed in version 3.7: Added the text parameter, as a more understandable alias of universal_newlines.
https://docs.python.org/3/library/subprocess.html#subprocess.run

netplan/cli/commands/apply.py
```diff
@@ -290,17 +295,25 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
                 utils.ip_addr_flush(iface)
             # clear NM state, especially the [device].managed=true config, as that might have been
             # re-set via an udev rule setting "NM_UNMANAGED=1"
-            subprocess.check_call(['rm', '-rf', '/run/NetworkManager/devices'])
+            subprocess.check_call(['rm', '-rf', '/run/NetworkManager/devices'], env=env)
```
ogayot, line 298: alternative approach is to use `shutil.rmtree` but it's not exactly equivalent so take it with a grain of salt :)

slyon: ACK. This is the cleaner approach and also being used elsewhere in the codebase.

netplan/cli/commands/apply.py
slyon: ACK. I've been holding back on this one consciously, for backwards compatibility reasons. But we're mostly targeting Python 3.10 (Jammy) nowadays, so should be fine to do this refactoring.

It should be done in an independent PR, though.

Makefile
slyon: Moved into #338 338

netplan/cli/commands/apply.py
slyon: see #338

ogayot (approved):
LGTM

netplan/cli/utils.py
slyon: Yes, moving it all into `cli/core.py` in the `class Netplan` constructor makes lots of sense. Let me test if this still works as expected (it should).

daniloegea (approved):
lgtm

slyon:
All tests green.

I've re-built a package in the above-mentioned PPA + core22 base snap + UC22 image and run the reproducer from the bug report:
=> It's still working as expected with the new `os.environ.update()` approach.

Merging.

PR #339, opened 2023-04-04
Reimplement optional-addresses (LP: #1880029)

daniloegea:
## Description

The Netplan's key "optional-addresses" only generates configuration for systemd-networkd. Although, the generated "OptionalAddresses" key doesn't exist so this property never worked on Netplan.

This PR re-implements the key "optional-addresses" to emit configuration for network manager, using the "may-fail" key, and for networkd, using the "RequiredFamilyForOnline" key.

Note that RequiredFamilyForOnline is actually the apposite of "optional addresses". It defaults to "any", meaning that either ipv4 or ipv6 will be enough to set the interface as online. Because of that, the only time when Netplan will generate RequiredFamilyForOnline is when the user enters "none", meaning that none of the addresses are optional, so RequiredFamilyForOnline=both will be emitted. 

For Network Manager, optional-addresses never generated configuration. Now, Netplan will emit the key "may-fail" for ipv4 and/or ipv6 according to the configuration.

All the old values accepted by optional-addresses are no-ops, so if any users out there have this configuration in their YAML files, nothing will break.

These changes seem to not be a perfect fit based on our documentation, but I guess it's close :sweat_smile: 

## Checklist

### DIFF
--- a/src/abi.h
+++ b/src/abi.h
@@ -30,6 +30,9 @@ typedef enum {
     NETPLAN_OPTIONAL_DHCP4   = 1<<2,
     NETPLAN_OPTIONAL_DHCP6   = 1<<3,
     NETPLAN_OPTIONAL_STATIC  = 1<<4,
+    NETPLAN_OPTIONAL_IPV4    = 1<<5,
+    NETPLAN_OPTIONAL_IPV6    = 1<<6,
+    NETPLAN_OPTIONAL_NONE    = 1<<7,
 } NetplanOptionalAddressFlag;
 
 /* Fields below are valid for dhcp4 and dhcp6 unless otherwise noted. */
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -739,15 +739,12 @@ netplan_netdef_write_network_file(
         is_optional = TRUE;
     }
 
-    if (is_optional || def->optional_addresses) {
-        if (is_optional) {
-            g_string_append(link, "RequiredForOnline=no\n");
-        }
-        for (unsigned i = 0; NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name != NULL; ++i) {
-            if (def->optional_addresses & NETPLAN_OPTIONAL_ADDRESS_TYPES[i].flag) {
-            g_string_append_printf(link, "OptionalAddresses=%s\n", NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name);
-            }
-        }
+    if (is_optional) {
+        g_string_append(link, "RequiredForOnline=no\n");
+    }
+
+    if (def->optional_addresses & NETPLAN_OPTIONAL_NONE) {
+        g_string_append(link, "RequiredFamilyForOnline=both\n");
     }
 
     if (def->mtubytes)
--- a/src/parse.c
+++ b/src/parse.c
@@ -1530,6 +1530,10 @@ NETPLAN_OPTIONAL_ADDRESS_TYPES[] = {
     {"dhcp4",   NETPLAN_OPTIONAL_DHCP4},
     {"dhcp6",   NETPLAN_OPTIONAL_DHCP6},
     {"static",  NETPLAN_OPTIONAL_STATIC},
+    /* All above are deprecated */
+    {"ipv4",    NETPLAN_OPTIONAL_IPV4},
+    {"ipv6",    NETPLAN_OPTIONAL_IPV6},
+    {"none",    NETPLAN_OPTIONAL_NONE},
     {NULL},
 };
 
@@ -1543,6 +1547,14 @@ handle_optional_addresses(NetplanParser* npp, yaml_node_t* node, const void* _, 
 
         for (unsigned i = 0; NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name != NULL; ++i) {
             if (g_ascii_strcasecmp(scalar(entry), NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name) == 0) {
+
+                /* Values below NETPLAN_OPTIONAL_STATIC (including it) are not valid and
+                 * considered deprecated. See LP: #1880029
+                 */
+                if (NETPLAN_OPTIONAL_ADDRESS_TYPES[i].flag <= NETPLAN_OPTIONAL_STATIC)
+                    g_warning("Flag \"%s\" in optional-addresses is deprecated. Valid values are: "
+                            "\"ipv4\", \"ipv6\" and \"none\"\n", NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name);
+
                 npp->current.netdef->optional_addresses |= NETPLAN_OPTIONAL_ADDRESS_TYPES[i].flag;
                 found = TRUE;
                 break;
--- a/tests/generator/test_common.py
+++ b/tests/generator/test_common.py
@@ -58,15 +58,26 @@ UseMTU=true
 
     def test_optional_addresses(self):
         eth_name = self.eth_name()
-        self.generate(self.config_with_optional_addresses(eth_name, '["dhcp4"]'))
-        self.assertEqual(self.get_optional_addresses(eth_name), set(["dhcp4"]))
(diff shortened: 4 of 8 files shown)


tests/generator/test_common.py
```diff
@@ -58,15 +58,26 @@ def config_with_optional_addresses(self, eth_name, optional_addresses):
...
         eth_name = self.eth_name()
-        self.generate(self.config_with_optional_addresses(eth_name, '["dhcp4"]'))
-        self.assertEqual(self.get_optional_addresses(eth_name), set(["dhcp4"]))
+        self.generate(self.config_with_optional_addresses(eth_name, '["ipv4"]'))
+        self.assertEqual(self.get_optional_addresses(eth_name), set())
 
     def test_optional_addresses_multiple(self):
         eth_name = self.eth_name()
-        self.generate(self.config_with_optional_addresses(eth_name, '[dhcp4, ipv4-ll, ipv6-ra, dhcp6, dhcp4, static]'))
+        self.generate(self.config_with_optional_addresses(eth_name, '["ipv4", "ipv6", "ipv4"]'))
```
daniloegea, line 66: "ipv4" is duplicated here for the same reason "dhcp4" was duplicated before: to have code coverage for the entire `normalize_yaml_tree()` method in generator/base.py :sweat_smile:

src/abi.h
```diff
@@ -30,6 +30,9 @@ typedef enum {
     NETPLAN_OPTIONAL_DHCP4   = 1<<2,
     NETPLAN_OPTIONAL_DHCP6   = 1<<3,
     NETPLAN_OPTIONAL_STATIC  = 1<<4,
+    NETPLAN_OPTIONAL_IPV4    = 1<<5,
+    NETPLAN_OPTIONAL_IPV6    = 1<<6,
```
slyon, line 34: IMO we shouldn't introduce new IPV4 & IPv6 flags, but rather extend "_STATIC" to "_STATIC4" and "_STATIC6". But the "static4" and "static6" filters should just be internally and should not propagate to the YAML.

The "_IPv4" filter is then just a shortcut for IPV4_LL + DHCP4 + STATIC4.
The "_IPv6" filter is then just a shortcut for IPV6_RA (should this actually be called IPV6_LL?!) + DHCP6 + STATIC6.

src/networkd.c
```diff
@@ -739,15 +739,12 @@ netplan_netdef_write_network_file(
...
-            g_string_append(link, "RequiredForOnline=no\n");
-        }
-        for (unsigned i = 0; NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name != NULL; ++i) {
-            if (def->optional_addresses & NETPLAN_OPTIONAL_ADDRESS_TYPES[i].flag) {
-            g_string_append_printf(link, "OptionalAddresses=%s\n", NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name);
-            }
-        }
+    if (is_optional) {
+        g_string_append(link, "RequiredForOnline=no\n");
+    }
```
slyon, line 744: ACK. This is the easiest case, and basically a shortcut for IPV4_LL + DHCP4 + STATIC4 + IPV6_RA/LL + DHCP6 + STATIC6.

src/networkd.c
```diff
@@ -739,15 +739,12 @@ netplan_netdef_write_network_file(
...
+
+    if (def->optional_addresses & NETPLAN_OPTIONAL_NONE) {
```
slyon, line 746:
```suggestion
    if (def->optional_addresses == 0x0) {
```

src/parse.c
```diff
@@ -1543,6 +1547,14 @@ handle_optional_addresses(NetplanParser* npp, yaml_node_t* node, const void* _,
 
         for (unsigned i = 0; NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name != NULL; ++i) {
             if (g_ascii_strcasecmp(scalar(entry), NETPLAN_OPTIONAL_ADDRESS_TYPES[i].name) == 0) {
+
+                /* Values below NETPLAN_OPTIONAL_STATIC (including it) are not valid and
+                 * considered deprecated. See LP: #1880029
+                 */
+                if (NETPLAN_OPTIONAL_ADDRESS_TYPES[i].flag <= NETPLAN_OPTIONAL_STATIC)
+                    g_warning("Flag \"%s\" in optional-addresses is deprecated. Valid values are: "
```
slyon, line 1555: We shouldn't deprecate the original keywords, but rather try to get them implemented in the corresponding backend renderers.

That is especially:
* Waiting LL configuration only
  * This can already be achieved on networkd, by using `RequiredForOnline=degraded` (instead of `routable/yes`). Can be in combination with `RequiredFamilyForOnline=ipv4/ipv6`. – But what happens if we want to have static or dynamic IP4 but only LL IP6?
* Waiting for static IP configuration, but not dynamic/DHCP.

slyon (changes requested):
A few remarks to start with:
* We should probably re-organize the internal flags/filters a little:
  - "ipv4-ll" = NETPLAN_OPTIONAL_IPV4_LL
  - "ipv6-ll" = NETPLAN_OPTIONAL_IPv6_LL
  - "dhcp4" = NETPLAN_OPTIONAL_DHCP4
  - "dhcp6" = NETPLAN_OPTIONAL_DHCP6
  - "static" = NETPLAN_OPTIONAL_STATIC4 & NETPLAN_OPTIONAL_STATIC6
* "ipv6-ra" should probably be re-named to "ipv6-ll". RA doesn't really make sense IMO, as we also have "dhcp6", which is kind of related.
* "static" should activate two internal filters, for finer control: NETPLAN_OPTIONAL_STATIC4 and NETPLAN_OPTIONAL_STATIC6 (in favor of the _STATIC flag)

Input variables:
We have multiple input variables that we need to consider when making a decision about "optional-addresses". Any address defined in the YAML is non-optional by default.
* IPv4 & IPv6 LL addresses, as defined in the `link-local:` setting. (`[ ipv6 ]` by default)
* IPv4 & IPv6 static addresses, as defined in the `addresses:` setting (IP address family is auto-detected).
* IPv4 & IPv6 dynamic addresses, as defined in the `dhcp4:` and `dhcp6:` (possibly in combination with `accept-ra:`) settings.

Considering the `optional-addresses:` schema:
Depending on the input variables (address types defined in the YAML), optional-addresses can be used to "filter out" some types, if we want to have them defined, but not block on them.

Basically, we have 4 levels of filters:
* internal flags: NETPLAN_OPTIONAL_IPV4_LL, NETPLAN_OPTIONAL_IPv6_LL, NETPLAN_OPTIONAL_DHCP4, NETPLAN_OPTIONAL_DHCP6, NETPLAN_OPTIONAL_STATIC4, NETPLAN_OPTIONAL_STATIC6
* YAML flags: ipv4-ll, ipv6-ra/ll, dhcp4, dhcp6, static
* (YAML shortcuts: ipv4, ipv6)
  * ipv4 == IPV4_LL + DHCP4 + STATIC4
  * ipv6 == IPV6_RA/LL + DHCP6 + STATIC6
  * I'm not yet sure if we should introduce those additional shortcuts, it might make things more complicated, but we can keep them here during implementation as a "reminder".
* YAML shortcut: `optional: true`
  * activates all the filters for this interface

So, overall I think we should keep the YAML schema for `optional-addresses` as-is, but split up the internal flags, to enable more fine-grained control. We should then introduce some helpers:
* wants_ipv4_static(addresses, optional_addresses_flags) -> bool
* wants_ipv6_static(addresses, optional_addresses_flags) -> bool
* wants_ipv4_ll(link_local, optional_addresses_flags) -> bool
* wants_ipv6_ll(link_local, optional_addresses_flags) -> bool
* wants_ipv4_dynamic(dhcp4, optional_addresses_flags) -> bool
* wants_ipv6_dynamic(dhcp6, [accept_ra,] optional_addresses_flags) -> bool

Building upon those filters we should then try to generate the corresponding output values for `RequiredForOnline=routable/yes/degraded/no` / `RequiredFamilyForOnline=ipv6/ipv6/both/any` (networkd) and `ipv4.may-fail` + `ipv6.may-fail` (NetworkManager) wherever possible, and print a warning for the cases that cannot be covered yet, due to backend limitations.

PR #340, opened 2023-04-14
netplan: adjust the maximum buffer size to 1MB

daniloegea:
It says it's 1MB but the value is actually 1GB

## Checklist


slyon (approved):
+1 good catch!

PR #341, opened 2023-04-17
vrfs: skip policies parsing if list is NULL (LP: #2016427)

daniloegea:
It addresses LP: #2016427

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2016427


daniloegea:
Actually, I suppose we should allow routing-policies without routes as well, right?

PR #343, opened 2023-04-20
parse: don't point to the wrong node on validation

daniloegea:
The function validate_netdef_grammar() checks several properties from the netdef that are not related to the current context (YAML node). This leads to error messages like these ones:

```
foo.yaml:5:7: Error ...: eth0: 'set-name:' requires 'match:' properties
      macaddress: c0:f3:7e:35:8e:1b
      ^
```
```
foo.yaml:5:7: Error in ...: wlan0: No access points defined
      macaddress: c0:f3:7e:35:8e:1b
      ^
```

Using the YAML tree node here doesn't seem to make sense, but removing it would break the ABI.

## Checklist

### DIFF
--- a/src/validation.c
+++ b/src/validation.c
@@ -351,59 +351,59 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, yaml
 
     /* set-name: requires match: */
     if (nd->set_name && !nd->has_match)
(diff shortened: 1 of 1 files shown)


slyon:
Could you please give some more context on *how* this would break ABI? Our ABI checker CI seems to be happy..

slyon (approved):
lgtm!

And the ABI checker seems to be happy, too :)

PR #344, opened 2023-04-20
networkd: plug a memory leak

daniloegea:
## Description

Believe or not, it was found with this https://github.com/daniloegea/netplan-fuzz :eyes: 

## Checklist


slyon (approved):
+1 nice find!

We should get your "netplan-fuzz" tool (and other fuzzing) integrated into a CI workflow.

PR #345, opened 2023-04-21
nm: return if write_routes() fails

daniloegea:
When routes validation fails we should return. It's also hitting an assertion later because the "error" object is already built:
```
(generate:246468): GLib-CRITICAL **: 09:45:27.526: g_key_file_save_to_file: assertion 'error == NULL || *error == NULL' failed
```

## Checklist


slyon (approved):
ACK, lgtm.

That's the same we do further down when writing the `[ipv4]`/`[ipv6]` sections.

PR #346, opened 2023-04-21
docs: fix bridge parameters types and add examples

daniloegea:
port-priority and path-cost are mappings.

## Checklist

### DIFF
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -1238,11 +1238,29 @@ The specific settings for bridges are defined below.
     > priority. The bridge with the higher priority will be elected as
     > the root bridge.
 
-  - **port-priority** (scalar)
+  - **port-priority** (mapping)
 
     > Set the port priority to <priority>. The priority value is
     > a number between `0` and `63`. This metric is used in the
     > designated port and root port selection algorithms.
+
+    Example:
+
+    ```yaml
+    network:
+      ethernets:
+        eth0:
+          dhcp4: false
+        eth1:
+          dhcp4: false
+      bridges:
+        br0:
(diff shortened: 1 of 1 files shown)


doc/netplan-yaml.md
```diff
@@ -1268,12 +1286,30 @@ The specific settings for bridges are defined below.
     > renderer is used. If no time suffix is specified, the value will be
     > interpreted as seconds.
 
-  - **path-cost** (scalar)
+  - **path-cost** (mapping)
 
     > Set the cost of a path on the bridge. Faster interfaces should have
     > a lower cost. This allows a finer control on the network topology
     > so that the fastest paths are available whenever possible.
```
slyon, line 1293:
```suggestion
    > Set the per-interface cost of a path on the bridge. Faster interfaces
    > should have a lower cost. This allows a finer control on the network
    > topology so that the fastest paths are available whenever possible.
```

doc/netplan-yaml.md
```diff
@@ -1238,12 +1238,30 @@ The specific settings for bridges are defined below.
     > priority. The bridge with the higher priority will be elected as
     > the root bridge.
 
-  - **port-priority** (scalar)
+  - **port-priority** (mapping)
 
     > Set the port priority to <priority>. The priority value is
     > a number between `0` and `63`. This metric is used in the
     > designated port and root port selection algorithms.
```
slyon, line 1245:
```suggestion
    > Set the port priority per interface. The priority value is
    > a number between `0` and `63`. This metric is used in the
    > designated port and root port selection algorithms.
```

slyon (approved):
+1, some tiny suggestions inline.

PR #349, opened 2023-04-24
parse: set the backend on nm-devices to NM by default

daniloegea:
By definition, the renderer on nm-devices must be NetworkManager. Not defining it will default to networkd. Apart from the fact it doesn't make sense, it is causing a memory leak because the function g_datalist_clear() will no be called in reset_backend_settings() if the backend is not NM.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -3034,6 +3034,11 @@ handle_network_type(NetplanParser* npp, yaml_node_t* node, const char* key_prefi
             case NETPLAN_DEF_TYPE_NM:
                 g_warning("netplan: %s: handling NetworkManager passthrough device, settings are not fully supported.", npp->current.netdef->id);
                 handlers = ethernet_def_handlers;
+                if (npp->current.netdef->backend != NETPLAN_BACKEND_NM) {
+                    g_warning("nm-device: %s: the renderer for nm-devices must be NetworkManager, it will be used instead of the defined one.",
+                            npp->current.netdef->id);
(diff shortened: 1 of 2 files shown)


daniloegea:
The memory leak above is fixed by https://github.com/canonical/netplan/pull/348

src/parse.c
```diff
@@ -3034,6 +3034,11 @@ handle_network_type(NetplanParser* npp, yaml_node_t* node, const char* key_prefi
             case NETPLAN_DEF_TYPE_NM:
                 g_warning("netplan: %s: handling NetworkManager passthrough device, settings are not fully supported.", npp->current.netdef->id);
                 handlers = ethernet_def_handlers;
+                if (npp->current.netdef->backend != NETPLAN_BACKEND_NM) {
+                    g_warning("nm-device: %s: the renderer for nm-devices must be NetworkManager, it will be used instead of the defined one.",
+                            npp->current.netdef->id);
```
slyon, line 3039: Formatting:
```suggestion
                    g_warning("nm-device: %s: the renderer for nm-devices must be NetworkManager, it will be used instead of the defined one.",
                              npp->current.netdef->id);
```

slyon (approved):
+1

I was pondering if it makes sense to downgrade this to a `g_debug(...)` message, maybe only if renderer is NETPLAN_BACKEND_NONE.

But it's better to have this as an explicit warning. People are not supposed to be using `nm-devices` manually at all. It's fine for them to see a warning should they choose a wrong (or non at all) renderer.

PR #350, opened 2023-04-24
Fix a memory leak, an assert and an error message

daniloegea:
## Checklist

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -94,8 +94,8 @@ type_str(const NetplanNetDefinition* def)
             return "ip-tunnel";
         case NETPLAN_DEF_TYPE_NM:
             /* needs to be overriden by passthrough "connection.type" setting */
-            g_assert(def->backend_settings.nm.passthrough);
-            GData *passthrough = def->backend_settings.nm.passthrough;
+            g_assert(def->backend_settings.passthrough);
+            GData *passthrough = def->backend_settings.passthrough;
             return g_datalist_get_data(&passthrough, "connection.type");
         // LCOV_EXCL_START
         default:
@@ -517,7 +517,7 @@ write_vxlan_parameters(const NetplanNetDefinition* def, GKeyFile* kf)
 
 /**
  * Special handling for passthrough mode: read key-value pairs from
- * "backend_settings.nm.passthrough" and inject them into the keyfile as-is.
+ * "backend_settings.passthrough" and inject them into the keyfile as-is.
  */
 static void
 write_fallback_key_value(GQuark key_id, gpointer value, gpointer user_data)
@@ -600,10 +600,10 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
     }
 
     kf = g_key_file_new();
-    if (ap && ap->backend_settings.nm.name)
-        g_key_file_set_string(kf, "connection", "id", ap->backend_settings.nm.name);
-    else if (def->backend_settings.nm.name)
-        g_key_file_set_string(kf, "connection", "id", def->backend_settings.nm.name);
+    if (ap && ap->backend_settings.name)
+        g_key_file_set_string(kf, "connection", "id", ap->backend_settings.name);
+    else if (def->backend_settings.name)
+        g_key_file_set_string(kf, "connection", "id", def->backend_settings.name);
     else {
         /* Auto-generate a name for the connection profile, if not specified */
         if (ap)
@@ -617,10 +617,10 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
     if (nm_type && def->type != NETPLAN_DEF_TYPE_NM)
         g_key_file_set_string(kf, "connection", "type", nm_type);
 
-    if (ap && ap->backend_settings.nm.uuid)
-        g_key_file_set_string(kf, "connection", "uuid", ap->backend_settings.nm.uuid);
-    else if (def->backend_settings.nm.uuid)
-        g_key_file_set_string(kf, "connection", "uuid", def->backend_settings.nm.uuid);
+    if (ap && ap->backend_settings.uuid)
+        g_key_file_set_string(kf, "connection", "uuid", ap->backend_settings.uuid);
+    else if (def->backend_settings.uuid)
+        g_key_file_set_string(kf, "connection", "uuid", def->backend_settings.uuid);
     /* VLAN/VXLAN devices refer to us as their parent; if our ID is not a name
      * but we have matches, parent= must be the connection UUID, so put it into
      * the connection */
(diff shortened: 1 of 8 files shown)


src/nm.c
```diff
@@ -545,6 +545,11 @@ write_fallback_key_value(GQuark key_id, gpointer value, gpointer user_data)
         group = g_strjoinv(".", group_key); //re-combine group parts
     }
 
+    if (!g_strcmp0(group, "")) {
+        g_warning("NetworkManager: keyfile group is empty for key '%s'.", key);
+        goto cleanup;
+    }
```
slyon, line 551: The assertion seems to stem from GLib's `g_key_file_is_group_name()` check: https://github.com/GNOME/glib/blob/main/glib/gkeyfile.c#L4158

But we're already in the output stage here in `nm.c`. The data that we have in our datastructures should be valid. So this seems more like a workaround. I feel like this should rather be fixed in `parse-nm.c` to avoid getting an invalid `group.key` value into the datastructures in the first place.

src/nm.c
```diff
@@ -365,6 +365,7 @@ write_wireguard_params(const NetplanNetDefinition* def, GKeyFile *kf, GError** e
         if (peer->preshared_key) {
             if (peer->preshared_key[0] == '/' && !is_wireguard_key(peer->preshared_key)) {
                 g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT, "%s: shared key needs to be base64 encoded when using the NM backend\n", def->id);
+                g_free(tmp_group);
```
slyon, line 368: This is fine!

suggestion: How about using `g_autofree gchar* tmp_group = NULL;` in the definition above, instead of this? That way we wouldn't need to take as much care in the future.

slyon (changes requested):
Thanks for spotting and fixing more papercuts!

I put one tiny suggestion inline (up to you if you want it that way!), and one comment about your fix in `nm.c` which I'd rather like to see fixed at its origin in `parse-nm.c`.

src/nm.c
daniloegea: Yeah, that works. I will just need to move the variable declaration to inside the for loop. Because it will be overridden with a new allocation for each iteration.

src/nm.c
daniloegea: In this case we actually read the passthrough data in handle_generic_datalist(). It doesn't do any parsing. If we want to not load invalid configuration I guess that would be the place.

daniloegea: But well, handle_generic_datalist() is generic so it doesn't need to know about the structure of the data it's reading...

src/parse.c
```diff
@@ -768,11 +768,33 @@ handle_netdef_map(NetplanParser* npp, yaml_node_t* node, const char* key_prefix,
...
+    g_strfreev(group_key);
+}
+
 static gboolean
-handle_netdef_datalist(NetplanParser* npp, yaml_node_t* node, const char* key_prefix, const void* data, GError** error)
+handle_passthrough_datalist(NetplanParser* npp, yaml_node_t* node, const char* key_prefix, const void* data, GError** error)
 {
     g_assert(npp->current.netdef);
-    return handle_generic_datalist(npp, node, key_prefix, npp->current.netdef, data, error);
+    gboolean ret = handle_generic_datalist(npp, node, key_prefix, npp->current.netdef, data, error);
```
daniloegea, line 792: I didn't want to touch handle_generic_datalist() or write a more specific version of it to not duplicate code. Although it would be more efficient that way.

My solution validated the passthrough words that were loaded from the YAML and remove the ones that doesn't have at least 2 components separated by a dot. According to the glib documentation, it is safe to remove the key inside a foreach. https://docs.gtk.org/glib/func.datalist_foreach.html

Let me know what you think!

slyon: Your approach wfm! It's good to warn the user when an entry is deleted/ignored.

But I think the function should be called `handle_netdef_passthrough_datalist` and we'd need a similar modification to `handle_access_point_datalist`. Also, we should think about what happens if the invalid entry was the only entry in the datalist. In this case, we should probably clear the datalist alltogether?

Consider extending the testcase like this:
```diff
--- a/tests/generator/test_passthrough.py
+++ b/tests/generator/test_passthrough.py
@@ -332,6 +332,14 @@ method=auto
 
     def test_passthrough_empty_keyfile_group(self):
         out = self.generate('''network:
+  wifis:
+    wlan0:
+      access-points:
+        "SSID":
+          networkmanager:
+            name: connection_name
+            passthrough:
+              itsmissingadot: abc
   nm-devices:
     device0:
       networkmanager:
```

slyon (changes requested):
Commented inline about the datalist approach.

src/parse.c
daniloegea: So, your new test case revealed another memory leak... I did something I was trying to avoid: I changed `reset_backend_settings()` to reset all the settings without checking what is the backend.

In your test for example, the key `networkmanager.name` would never be freed if the backend is not NetworkManager. So I just removed the `switch (){case}` block.

src/types.c
```diff
@@ -164,20 +164,12 @@ reset_ip_rule(NetplanIPRule* ip_rule)
...
-            FREE_AND_NULLIFY(settings->nm.uuid);
-            FREE_AND_NULLIFY(settings->nm.stable_id);
-            FREE_AND_NULLIFY(settings->nm.device);
-            g_datalist_clear(&settings->nm.passthrough);
-            break;
-        default:
-            break;
-    }
+    FREE_AND_NULLIFY(settings->networkd.unit);
+    FREE_AND_NULLIFY(settings->nm.name);
```
slyon, line 168: I was pondering if this could lead to double-frees or corrupt data, as `backend_settings` is a union of `struct NetplanNMSettings` and `struct NetplanNetworkdSettings`, therefore `settings->networkd.unit` and `settings.nm->name` are basically the same pointer (both `char *`).

But we're nullifying the pointer, so the `FREE_AND_NULLIFY(settings->nm.name);` call is basically a no-op here. But we should probably it as-is for clarity.

Looking further.. I don't think `backend_settings->networkd.unit` is being used anywhere. I wonder why it's there at all?? Maybe we should just replace that union with just the `NetplanNMSettings` struct, which should be ABI-safe, as we won't change memory size or layout. We could then add an additional `NetplanNetworkdSettings` struct in the future (separately, not as a union), should we ever need it. This should make the situation much clearer. WDYT`

slyon (commented):
Two more inline comments:
* We cannot be sure the `renderer` already got set in the `parse.c` stage, we need to check that in `validation.c` stage, maybe using a `has_backend_settings_nm` helper variable.
* We might want to drop the `NetplanBackendSettings` union all together. `backend_settings.networkd.union` doesn't seem to be used anywhere. Replacing the union by just `struct NetplanNMSettings` should be ABI-safe IMO.

src/types.c
daniloegea: I did just that, it made things a bit simpler.

daniloegea:
I moved the wireguard part to another PR (#352).

I converted the NetplanBackendSettings union to a struct as suggested, I think it makes sense.

src/validation.c
```diff
@@ -402,12 +403,19 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, GErr
...
         return yaml_error(npp, NULL, error, "%s: network type 'nm-devices:' needs to provide a 'connection.type' via passthrough", nd->id);
 
     if (npp->current.netdef)
         validate_interface_name_length(npp->current.netdef);
 
+    if (backend == NETPLAN_BACKEND_NONE)
+        backend = npp->global_backend;
+
+    if (nd->has_backend_settings_nm && backend != NETPLAN_BACKEND_NM) {
+            return yaml_error(npp, NULL, error, "%s: networkmanager.passthrough settings found but renderer is not NetworkManager.", nd->id);
```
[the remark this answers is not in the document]
daniloegea: Done :)

PR #353, opened 2023-04-27
ovs: don't allow peers with the same name

daniloegea:
I found it thanks to a memory leak. If both peers have the same name, the netdef will end up with a `netdef->peer_link` pointing to itself and `netdef->link` strdup'ed twice. 

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -603,7 +603,7 @@ handle_netdef_id_ref(NetplanParser* npp, yaml_node_t* node, const void* data, GE
         *dest = ref;
 
         if (netdef->type == NETPLAN_DEF_TYPE_VLAN && ref->backend == NETPLAN_BACKEND_OVS) {
-            g_debug("%s: VLAN defined for openvswitch interface, choosing OVS backend", netdef->id);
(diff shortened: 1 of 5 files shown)


src/parse.c
```diff
@@ -2926,6 +2926,9 @@ handle_network_ovs_settings_global_ports(NetplanParser* npp, yaml_node_t* node,
         peer = yaml_document_get_node(&npp->doc, *(item+1));
         assert_type(npp, peer, YAML_SCALAR_NODE);
 
+        if (!g_strcmp0(scalar(port), scalar(peer)))
+            return yaml_error(npp, peer, error, "Openvswitch peer ports must be different");
```
slyon, line 2930: nitpick: improve error message a bit.
```suggestion
            return yaml_error(npp, peer, error, "Open vSwitch patch ports must be of different name");
```

We use the "patch port" term in our documentation, so should be using it here, too (and probably in the other OVS error messages). And we should also use the official "Open vSwitch" writing for OVS (same with the other error messages).

slyon (approved):
lgtm.

But we might want to improve this error message and the other OVS messages a bit.

slyon (approved):
+1

PR #361, opened 2023-05-22
Add support for "dummy" interfaces (LP: #1774203)

daniloegea:
## Description

This is a resurrection of #115. I didn't reuse the PR because it's old and incomplete.

It implements support for parsing YAMLs and Keyfiles so it works with netplan-everywhere (Network Manager + Netplan).

Related LP bug: https://bugs.launchpad.net/netplan/+bug/1774203

Unfortunately the Linux kernel name for this type of interface conflicts with our sensitive-wording  policy so I had to add some annotations in the code to ignore the errors from Woke. Our words list doesn't catch the word 'dummies' though (flaw I exploited to not insert the wokeignore annotation on some places), so I don't know, maybe we could just drop "dummy" from our denylist...

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1774203

### DIFF
--- a/.woke.yaml
+++ b/.woke.yaml
@@ -103,14 +103,6 @@ rules:
       - foreperson
     severity: warning
 
-  - name: dummy
-    terms:
-      - dummy
-    alternatives:
-      - placeholder
-      - sample
-    severity: warning
-
   - name: grandfathered
     terms:
       - grandfathered
@@ -178,3 +170,4 @@ rules:
     severity: warning
   # Ignore rules
   - name: he
+  - name: dummy
--- a/abi-compat/suppressions.abignore
+++ b/abi-compat/suppressions.abignore
@@ -7,3 +7,8 @@
 [suppress_variable]
   name = global_state
   type_name = NetplanState
+
+[suppress_type]
+  type_kind = typedef
+  name = NetplanDefType
+  changed_enumerators = NetplanDefType
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -12,6 +12,7 @@ network:
   renderer: STRING
   bonds: MAPPING
   bridges: MAPPING
+  dummy-devices: MAPPING
   ethernets: MAPPING
   modems: MAPPING
   tunnels: MAPPING
@@ -36,6 +37,10 @@ network:
 - [**bridges**](#properties-for-device-type-bridges) (mapping)
 
   > Creates and configures bridge devices.
+
+- [**dummy-devices**](#properties-for-device-type-dummy-devices) (mapping)
+
+  > Creates and configures virtual devices.
 
 - [**ethernets**](#properties-for-device-type-ethernets) (mapping)
 
@@ -1315,6 +1320,31 @@ The specific settings for bridges are defined below.
     > Define whether the bridge should use Spanning Tree Protocol. The
     > default value is "true", which means that Spanning Tree should be
     > used.
+
+
+## Properties for device type `dummy-devices:`
+
+**Status**: Optional.
+
+**Purpose**: Use the `dummy-devices` key to create virtual interfaces.
+
+**Structure**: The key consists of a mapping of interface names.
+Dummy devices are virtual devices that can be used to route packets to
+without actually transmitting them.
+
+```yaml
+network:
+  dummy-devices:
+    dm0:
+      addresses:
+        - 192.168.0.123/24
+      ...
+```
+
+When applied, a virtual interface called `dm0` will be created in the system.
+
+[See this section](#properties-for-all-device-types) for the list of properties that can be
+used with this type of interface.
 
 
 ## Properties for device type `bonds:`
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -1343,7 +1343,7 @@ network:
 
 When applied, a virtual interface called `dm0` will be created in the system.
 
-[See this section](#properties-for-all-device-types) for the list of properties that can be
+See the ["Properties for all device types"](#properties-for-all-device-types) section for the list of properties that can be
 used with this type of interface.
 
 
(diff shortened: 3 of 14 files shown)


doc/netplan-yaml.md
```diff
@@ -1317,6 +1322,31 @@ The specific settings for bridges are defined below.
...
+    dm0:
+      addresses:
+        - 192.168.0.123/24
+      ...
+```
+
+When applied, a virtual interface called `dm0` will be created in the system.
+
+[See this section](#properties-for-all-device-types) for the list of properties that can be
+used with this type of interface.
```
slyon, line 1347: How does this look inside the rendered manpage? We don't really have hyperlinks/anchors there.

```suggestion
See the ["Properties for all device types"](#properties-for-all-device-types) section for the list of properties that can be
used with this type of interface.
```

tests/integration/dummies.py
```diff
@@ -0,0 +1,89 @@
...
+            f.write('''network:
+  renderer: %(r)s
+  version: 2
+  dummies:
+    dm0:
+      addresses:
+        - 192.168.123.123/24
+        - 1234:FFFF::42/64
+''' % {'r': self.backend})
+        self.generate_and_settle(['dm0'])
```
slyon, line 73: We could consider waiting for `state_up('dm0')` here.

abi-compat/suppressions.abignore
```diff
@@ -7,3 +7,8 @@
 [suppress_variable]
   name = global_state
   type_name = NetplanState
+
+[suppress_type]
+  type_kind = typedef
+  name = NetplanDefType
```
slyon, line 13: I agree we can allow this ABI change to the NetplanDeFType enum, as we're only changing the internal `NETPLAN_DEF_TYPE_MAX_` value, which does not affect existing consumers.

slyon (commented):
LGTM, some small nitpicks inline.

The new "dummies"/"dummy" schema is non-inclusive language, but I agree it's the common wording, also used by iproute2, systemd-networkd and NetworkManager. We should not diverge from that common understanding.

Let's try to get some more opinions on that.

Also we should keep in mind this old comment from **cyphermox**:
> I'm not fundamentally opposed to dummy devices, but I wonder if they wouldn't better fit elsewhere (like under ethernet, simply) since dummies are indeed a thing, but once you move into network hardware, you'll much more likely want to create loopback devices, which is something else that is not currently straightforward.

doc/netplan-yaml.md
daniloegea: Oops, good point!

slyon (approved):
Thanks.

I think `examples/*.yaml` is what's being used for `tools/run_asan.sh` in our memory leak CI, so it would be nice to have one dummy example in there, too. But not required.

Let's wait on consensus for the schema change (& adding "schema ok" label), before merging.

daniloegea:
That's a good point. I added the examples file and rebased.

varCharlie:
Hopping on here to say I will be very happy when netplan allows the use of dummy interfaces...  anyone doing anycast routing will use this feature (lots of big network operators).

It will be a great way to make netplan more useful for managing network services on servers

doc/netplan-yaml.md
```diff
@@ -12,6 +12,7 @@ network:
   renderer: STRING
   bonds: MAPPING
   bridges: MAPPING
+  dummies: MAPPING
```
vorlonofportland, line 15: Hmm, this is difficult.  While `dummy` is clearly correct, the plural `dummies` evokes for me images of mannequins or crash test dummies or martial arts training dummies.  At the risk of being too long, what do you think of `dummy-interfaces` here?

daniloegea: oh `dummy-interfaces` would deviate from the naming pattern... `ethernets`, `bonds`, `bridges`... `dummy-interfaces`... as a non-native English speaker I wasn't seeing the type of dummies you mentioned but now I can't unsee them :D

If we are not going with `dummies`, maybe we could go with `dummy-devices` then? Because we already have a `nm-devices` category.

vorlonofportland: works for me

slyon: Same for me... I didn't think about cash test dummies and thought "dummies" would be most in line with what we have so far for the other types. But I see Steve's point. Considering we also have "nm-devices", I think I'd go with "dummy-devices" instead of "dummy-interfaces".

daniloegea:
I replaced `dummies` by `dummy-devices` (and also rebased the branch). I also had to do some changes in our woke ruleset because of the word `dummy`...

.woke.yaml
```diff
@@ -1,5 +1,8 @@
 ignore_files:
   - abi-compat/*.xml
+  - doc/netplan-yaml.md
```
slyon, line 3: This is a big hammer. But I guess there's no better way to exclude "dummy*" from verbatim code examples inside  the docs... Or maybe we should really just allow the word "dummy" globally.

This looks relevant here: https://github.com/get-woke/woke/discussions/159

I wonder if we can do something like this in our `.woke` configuration?
```
  - name: dummy
    terms:
      - dummy
    alternatives:
      - placeholder
      - sample
    severity: warning
    options:
      word_boundary_end: true
```

Well.. after some experimentation, this doesn't seem to do the trick.

Maybe instead we should change our `.woke` to use `severity: error` by default but keep it at `info` for "dummy" and use `fail-on-error: true` inside our woke GH action.

Well.. some more experimentation and this doesn't seem to work either, as `woke` doesn't differentiate between info/warning/error severity in their return value. (Really woke?!)

So in the end I think our best option is to comment out the woke rules for "dummy" (even though I rejected that idea initially). We could still keep the "wokeignore" comments around in the code, I guess. WDYT?

slyon (approved):
One small inline question about `.woke` vs our documentatio wrt "dummy" words. Otherwise this is good to be merged. Thanks!

.woke.yaml
daniloegea: I agree. If one day woke allows the use of exceptions we can put it back.

PR #362, opened 2023-05-24
CI: add DebCI integration test

slyon:
## Description
This is running on Debian testing inside a raw LXC container and should resemble the ci.debian.net migration pretty closely.

## Checklist

### DIFF
--- a/.github/workflows/debci.yml
+++ b/.github/workflows/debci.yml
@@ -46,7 +46,10 @@ jobs:
           dget -u "https://deb.debian.org/debian/pool/main/n/netplan.io/netplan.io_$V.dsc"
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
-          dch -v $(git describe --long --tags $(git rev-list --tags --max-count=1)) "Autopkgtest CI testing (Debian testing)"
+          TAG=$(git describe --tags $(git rev-list --tags --max-count=1))  # find latest (stable) tag
+          REV=$(git rev-parse --short HEAD)  # get current git revision
+          VER="$TAG+git~$REV"
+          dch -v "$VER" "Autopkgtest CI testing (Debian testing)"
       - name: Run autopkgtest (incl. build)
         run: |
           # using --setup-commands='apt -y install ...' temporarily to install
--- a/.github/workflows/debci.yml
+++ b/.github/workflows/debci.yml
@@ -1,0 +1,54 @@
+name: Autopkgtest DebCI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the main branch
+on:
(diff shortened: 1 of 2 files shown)


.github/workflows/debci.yml
```diff
@@ -0,0 +1,52 @@
...
+      #      (LP: #1950787, LP: #1998943)
+      - name: Preparing autopkgtest-build-lxc
+        run: |
+          # Fix Docker blocking LXC networking:
+          # https://discuss.linuxcontainers.org/t/9953/4
+          sudo iptables -I DOCKER-USER  -j ACCEPT
+          sudo apparmor_parser -R /etc/apparmor.d/usr.bin.lxc-start
+          sudo ln -s /etc/apparmor.d/usr.bin.lxc-start /etc/apparmor.d/disable/
+          echo "lxc.apparmor.profile = unconfined" | sudo tee -a /etc/lxc/default.conf
+          sudo debci setup -s testing -a amd64 -b lxc
```
daniloegea, line 41: Should we test against unstable? On Debian it is tested against unstable first right? I think it would be more realistic because the dependencies in unstable might be different... although we might run into problems we are not interested in, like when some dependency is temporarily broken... yeah maybe for us it's better to use testing...

.github/workflows/debci.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Prepare test
+        run: |
+          pull-debian-source netplan.io
+          cp -r netplan.io-*/debian .
+          rm -r debian/patches/  # clear any distro patches
+          dch -v $(git describe --long --tags $(git rev-list --tags --max-count=1)) "Autopkgtest CI testing (Debian testing)"
+      - name: Run autopkgtest (incl. build)
+        run: |
+          # using --setup-commands='apt -y install ...' temporarily to install
```
daniloegea, line 50: I guess these comments were left behind

daniloegea (approved):

.github/workflows/debci.yml
slyon: Debian builds the package in `unstable` but then DebCI does run the autopkgtest against a `testing` environment, to check the migration to testing. Similarly how Ubuntu builds in `-proposed` then runs the autopkgtests against a `devel` environment. So we want to replicate this as closely as possible.

.github/workflows/debci.yml
slyon: I let them sit there on purpose, so we can use it as a template when we need to mangle with the dependencies next time.

slyon:
Also see https://github.com/canonical/netplan/commit/ec27ce966b54006e559411f7fdb86b3d01f6ea96 for some basic RPM (Fedora & EL9) build & test CI

PR #364, opened 2023-05-26
CI: fix dch after adding the new 0.106.1 tag

daniloegea:
After adding the new tag from the stable/0.106 branch, dch is failing with:

  New version specified (0.106-78-g3074070) is less than
  the current version number (0.106.1-0ubuntu1)!  Use -b to force.

Basically, git describe is still describing the tag 0.106 and not 0.106.1.
With this change, we'll first find the lastest tag before calling describe.

## Checklist


schopin-pro (approved):

schopin-pro:
ACK. I haven't gone through all the scenarios in which this could fail down the line, and in the case of our CI config I figure a quickfix now is better than an hypothetical "correct" fix that doesn't yet exist.

PR #365, opened 2023-05-30
Improve the coverage of the memory leak tests

daniloegea:
## Description

Now the run_asan.sh script will call the generator twice for each yaml file, once with renderer=networkd and again with renderer=NetworkManager.

When the renderer is set to NetworkManager, it will also call a new tool (keyfile_to_yaml.c) to load the keyfile.

Also, plug two memory leaks :)

## Checklist


schopin-pro (approved):
I like it :)

Good call on using the system netplan to change the renderer!

PR #367, opened 2023-06-13
routes: fix metric rendering (LP: #2023681)

bengentil:
Routes metric are rendered as signed integer instead of unsigned integer. This means all metric > 2147483647 is rendered as negative and is breaking the generated configuration.

## Description

https://bugs.launchpad.net/netplan/+bug/2023681

## Checklist

- [-] New/changed keys in YAML format are documented.
- [-] \(Optional\) Adds example YAML for new feature.
- [-] \(Optional\) Closes an open bug in Launchpad.

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -591,7 +591,7 @@ write_route(NetplanIPRoute* r, GString* s)
     if (r->onlink)
         g_string_append_printf(s, "GatewayOnLink=true\n");
     if (r->metric != NETPLAN_METRIC_UNSPEC)
-        g_string_append_printf(s, "Metric=%d\n", r->metric);
+        g_string_append_printf(s, "Metric=%u\n", r->metric);
     if (r->table != NETPLAN_ROUTE_TABLE_UNSPEC)
         g_string_append_printf(s, "Table=%d\n", r->table);
     if (r->mtubytes != NETPLAN_MTU_UNSPEC)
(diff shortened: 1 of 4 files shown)


daniloegea:
Thanks so much for you contribution, @bengentil!

Your fix looks good. Do you think you can write a couple of unit test for it though?

You could add one for networkd and another for Network Manager in `tests/generator/test_routing.py`. Maybe calling them something like `test_route_metric_rendering_lp2023681`. It will ensure we'll not reintroduce this issue in the future.

Thanks again!

daniloegea (approved):
Approved. Check my other comment, please. Maybe we could add a couple of unit tests as well.

bengentil:
Hi @daniloegea, I've pushed a couple of tests.

The tests are testing against 4294967294, which is not strictly the max value as the max value is used internally to set the metric as undefined (NETPLAN_METRIC_UNSPEC == G_MAXUINT == 4294967295)

For more details see:
* https://github.com/canonical/netplan/blob/012ce6a9fda89d7f4cded8d8195bab9b3fcca6e2/src/netplan.c#L570
*
https://github.com/canonical/netplan/blob/012ce6a9fda89d7f4cded8d8195bab9b3fcca6e2/src/netplan.c#LL64C9-L64C26

This means technically 4294967295 can't be set as metric with netplan but this is another issue, here the goal is to avoid signed values in metrics.

slyon (approved):
Thank you very much for your contribution. LGTM!

I'll leave merging to Danilo, when he feels his comments are resolved.

PR #368, opened 2023-06-15
Support for VETH interfaces

daniloegea:
# Design: Netplan support for VETH

## Rationale 

Veths are Virtual Ethernet interfaces that are commonly used to
connect bridges or network namespaces. They act like a link between
two networks, simply bypassing traffic from one veth peer to the other.

Veths work in peers, meaning that users will always create two of them.

This implementation allows users to create peers of veths, configure them with
IPs, routes and etc, change their MAC addresses and use them as member
interfaces for bridges and bonds for example. It doesn't support adding veths
to different network namespaces, although it could be done by an external tool.

## Changes in the schema

As both peers are configurable with IPs and etc, they must be defined in
separate stanzas.

A new type of network called `virtual-ethernets` was created and one extra key called 
`peer` was introduced.

```yaml
network:
  version: 2
  renderer: networkd
  virtual-ethernets:
    veth0-peer1:
      peer: veth0-peer2
    veth0-peer2:
      peer: veth0-peer1

  bridges:
    br0:
      interfaces:
        - veth0-peer1
    br1:
      interfaces:
        - veth0-peer2
```

Even though `veth` only work in pairs, it's not mandatory to define the peer
stanza (the `peer` key is mandatory) when using NetworkManager as renderer. The `peer` interface is not required
because Network Manager requires the creation of two connections, one for each
peer. As our keyfile parser maps a keyfile to a YAML file (a 1:1 mapping), if
we forced the creation of both peers, we'd had problems with the
Netplan-NetworkManager integration.

## systemd-networkd backend

Networkd requires a single `.netdev` file to create both veths.
So the configuration above will generate the `.netdev` file below:

```ini
[NetDev]
Name=veth0-peer1
Kind=veth

[Peer]
Name=veth0-peer2
```

If further configuration is required, one `.network` file per veth peer must be
created.

As only one `.netdev` is needed, we sort both peer names and use the first one.
If the peer stanza is not defined, it will not generate any files.

## Network Manager backend

On Network Manager, one keyfile per peer is required. So, creating a `veth` pair
requires 2 connections.

`veth-peer1.nmconnection`:
```ini
[connection]
id=veth-peer1
uuid=64cec75d-1f76-4379-bd82-684bc77484b9
type=veth
interface-name=veth-peer1

[veth]
peer=veth-peer2
```

`veth-peer2.nmconnection`:
```ini
[connection]
id=veth-peer2
uuid=a5d3f350-15f7-4e21-a2d4-be549531fba3
type=veth
interface-name=veth-peer2

[veth]
peer=veth-peer1
```

## Netplan Everywhere
One of the implications of these changes is that Network Manager will
create YAML files for `virtual-ethernets` automatically.

`veth` devices will be seen as unmanaged by Network Manager until the file
`/var/run/NetworkManager/conf.d/10-globally-managed-devices.conf` is
created by the generator and Network Manager is restarted. The implication
is that, `veth` pairs created with NM on a system that is not already
managed by NM will not come up online until the user either calls
`netplan apply` or restart NM.

### Keyfile parser

The keyfile parser will create a 1:1 mapping between the keyfile 
and the interface in the YAML file it generates. That's the main
reason why it's not mandatory to defined both peers at once.

That approach enable Netplan Everywhere to work out of the box
with `virtual-ethernets`.

## Alternative solution

$NOTE$: this was the original approach implemented. But due to complications
with the integration with Network Manager, we used the approach described
above.

We could make it mandatory to define both peers at once. It makes sense
because `virtual-ethernets` only work in pairs.

When creating the `veth` interface from the keyfile, a stanza for the peer
is also created and linked to the `veth`. That basically means that a keyfile
for a `veth` interface will generate 2 interfaces in the YAML file, the
interface itself and its peer. The same will happen when the second `veth`
(the peer) is created in a second API call to Network Manager. In the end,
2 Netplan YAML files will be created. When the generator is called, it will
merge both files, creating the full `veth` configuration.

This approach will require some changes in the Network Manager integration with Netplan.
Currently, Network Manager assumes a single netdef is present in the
Netplan state, but now we will have 2 netdefs. Both netdefs must be part of
the same YAML file so, instead of writing the netdef to the file, we will
need to write the entire netplan state (which will contain 2 netdefs).

After some changes to the Network Manager's integration patch, 
that's the results:

Adding the first `veth` interface:

```
nmcli con add \
  con-name veth-nm-peer1 \
  type veth \
  peer veth-nm-peer2 \
  connection.interface-name veth-nm-peer1
```

This will create the file
`/etc/netplan/90-NM-d4afd034-9634-4a67-a3ee-0c52154eefa3.yaml` with the
following content:

```yaml
network:
  version: 2
  virtual-ethernets:
    veth-nm-peer1:
      renderer: NetworkManager
      dhcp4: true
      dhcp6: true
      peer: "veth-nm-peer2"
      networkmanager:
        uuid: "d4afd034-9634-4a67-a3ee-0c52154eefa3"
        name: "veth-nm-peer1"
        passthrough:
          ipv6.addr-gen-mode: "default"
          ipv6.ip6-privacy: "-1"
          proxy._: ""
    veth-nm-peer2:
      renderer: NetworkManager
      peer: "veth-nm-peer1"
```

Now, adding the `veth` peer with the following command:

```
nmcli con add \
  con-name veth-nm-peer2 \
  type veth \
  peer veth-nm-peer1 \
  connection.interface-name veth-nm-peer2
```

This will create the file
`/etc/netplan/90-NM-505fc532-a910-4346-8bd8-19e849b696e6.yaml` with the
following content:

```yaml
network:
  version: 2
  virtual-ethernets:
    veth-nm-peer2:
      renderer: NetworkManager
      dhcp4: true
      dhcp6: true
      peer: "veth-nm-peer1"
      networkmanager:
        uuid: "505fc532-a910-4346-8bd8-19e849b696e6"
        name: "veth-nm-peer2"
        passthrough:
          ipv6.addr-gen-mode: "default"
          ipv6.ip6-privacy: "-1"
          proxy._: ""
    veth-nm-peer1:
      renderer: NetworkManager
      peer: "veth-nm-peer2"
```

Network Manager now has both connections:

```
NAME           UUID                                  TYPE  DEVICE
veth-nm-peer1  d4afd034-9634-4a67-a3ee-0c52154eefa3  veth  --
veth-nm-peer2  505fc532-a910-4346-8bd8-19e849b696e6  veth  --
```

`netplan get` will then produce the following result:

```yaml
network:
  version: 2
  virtual-ethernets:
    veth-nm-peer2:
      renderer: NetworkManager
      dhcp4: true
      dhcp6: true
      peer: "veth-nm-peer1"
      networkmanager:
        uuid: "505fc532-a910-4346-8bd8-19e849b696e6"
        name: "veth-nm-peer2"
        passthrough:
          ipv6.addr-gen-mode: "default"
          ipv6.ip6-privacy: "-1"
          proxy._: ""
    veth-nm-peer1:
      renderer: NetworkManager
      dhcp4: true
      dhcp6: true
      peer: "veth-nm-peer2"
      networkmanager:
        uuid: "d4afd034-9634-4a67-a3ee-0c52154eefa3"
        name: "veth-nm-peer1"
        passthrough:
          ipv6.addr-gen-mode: "default"
          ipv6.ip6-privacy: "-1"
          proxy._: ""
```
### Deleting connections

Netplan's `netplan_delete_connection()` will try to remove one of the peers
from the Netplan's state. As a `veth` peer cannot exist without the other,
`netplan_delete_connection()` will need to remove both connections when
any one of them is deleted.

The problem with this approach is that, after deleting one of the veths,
both of them will disappear from the YAML files but one connection will be kept
in the Network Manager's state.

So, deleting an interface, would require the deletion of both of them. If
we delete a single interface, Netplan will fail to parse the one that's left.
If we delete both of them from netplan, after running `nmcli con del`, one would
still be in the Network Manager's state so it would be out of sync with netplan.

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -524,6 +524,21 @@ write_netdev_file(const NetplanNetDefinition* def, const char* rootdir, const ch
             g_string_append_printf(s, "Kind=dummy\n");      /* wokeignore:rule=dummy */
             break;
 
+        case NETPLAN_DEF_TYPE_VETH:
+            /*
+             * Only one .netdev file is required to create the veth pair.
+             * To select what netdef we are going to use, we sort both names, get the first one,
+             * and, if the selected name is the name of the netdef being written, we generate
+             * the .netdev file. Otherwise we skip the netdef.
+             */
+            gchar* first = g_strcmp0(def->id, def->veth_peer_link->id) < 0 ? def->id : def->veth_peer_link->id;
+            if (first != def->id) {
+                g_string_free(s, TRUE);
+                return;
+            }
+            g_string_append_printf(s, "Kind=veth\n\n[Peer]\nName=%s\n", def->veth_peer_link->id);
+            break;
+
         case NETPLAN_DEF_TYPE_TUNNEL:
             switch(def->tunnel.mode) {
                 case NETPLAN_TUNNEL_MODE_GRE:
--- a/src/parse.c
+++ b/src/parse.c
@@ -810,6 +810,35 @@ handle_netdef_passthrough_datalist(NetplanParser* npp, yaml_node_t* node, const 
     return ret;
 }
 
+static gboolean
+handle_veth_peer(NetplanParser* npp, yaml_node_t* node, __unused const void* data, GError** error)
+{
+    NetplanNetDefinition* netdef = npp->current.netdef;
+
+    if (!g_strcmp0(netdef->id, scalar(node)))
+        return yaml_error(npp, node, error, "%s: virtual-ethernet peer cannot be itself", netdef->id);
+
+    NetplanNetDefinition* link = g_hash_table_lookup(npp->parsed_defs, scalar(node));
+    if (link) {
+        if (link->type != NETPLAN_DEF_TYPE_VETH && link->type != NETPLAN_DEF_TYPE_NM_PLACEHOLDER_)
+            return yaml_error(npp, node, error, "%s: virtual-ethernet peer '%s' is not a virtual-ethernet interface", netdef->id, link->id);
+
+        if (link->veth_peer_link && link->veth_peer_link != netdef)
+            return yaml_error(npp, node, error, "%s: virtual-ethernet peer '%s' is another virtual-ethernet's (%s) peer already",
+                              netdef->id, link->id, link->veth_peer_link->id);
+
+        netdef->veth_peer_link = link;
+        link->veth_peer_link = netdef;
+
+        return TRUE;
+    }
+
+    add_missing_node(npp, node);
+
+    return TRUE;
+}
+
+
 /****************************************************
  * Grammar and handlers for network config "match" entry
  ****************************************************/
@@ -2789,6 +2818,14 @@ static const mapping_entry_handler ethernet_def_handlers[] = {
     {NULL}
 };
 
+static const mapping_entry_handler veth_def_handlers[] = {
+    COMMON_LINK_HANDLERS,
+    COMMON_BACKEND_HANDLERS,
+    {"peer", YAML_SCALAR_NODE, {.generic=handle_veth_peer}, netdef_offset(veth_peer_link)},
+    {NULL}
+};
+
+
 static const mapping_entry_handler wifi_def_handlers[] = {
     COMMON_LINK_HANDLERS,
     COMMON_BACKEND_HANDLERS,
@@ -3126,6 +3163,7 @@ handle_network_type(NetplanParser* npp, yaml_node_t* node, const char* key_prefi
             case NETPLAN_DEF_TYPE_VRF: handlers = vrf_def_handlers; break;
             case NETPLAN_DEF_TYPE_WIFI: handlers = wifi_def_handlers; break;
             case NETPLAN_DEF_TYPE_DUMMY: handlers = dummy_def_handlers; break;      /* wokeignore:rule=dummy */
+            case NETPLAN_DEF_TYPE_VETH: handlers = veth_def_handlers; break;
             case NETPLAN_DEF_TYPE_NM:
                 g_debug("netplan: %s: handling NetworkManager passthrough device, settings are not fully supported.", npp->current.netdef->id);
                 handlers = ethernet_def_handlers;
@@ -3216,6 +3254,7 @@ static const mapping_entry_handler network_handlers[] = {
     {"wifis", YAML_MAPPING_NODE, {.map={.custom=handle_network_type}}, GUINT_TO_POINTER(NETPLAN_DEF_TYPE_WIFI)},
     {"modems", YAML_MAPPING_NODE, {.map={.custom=handle_network_type}}, GUINT_TO_POINTER(NETPLAN_DEF_TYPE_MODEM)},
     {"dummy-devices", YAML_MAPPING_NODE, {.map={.custom=handle_network_type}}, GUINT_TO_POINTER(NETPLAN_DEF_TYPE_DUMMY)},    /* wokeignore:rule=dummy */
+    {"virtual-ethernets", YAML_MAPPING_NODE, {.map={.custom=handle_network_type}}, GUINT_TO_POINTER(NETPLAN_DEF_TYPE_VETH)},
     {"nm-devices", YAML_MAPPING_NODE, {.map={.custom=handle_network_type}}, GUINT_TO_POINTER(NETPLAN_DEF_TYPE_NM)},
     {"openvswitch", YAML_MAPPING_NODE, {.map={.handlers=ovs_network_settings_handlers}}, NULL},
     {NULL}
@@ -3255,6 +3294,18 @@ process_missing_ids(NetplanParser* npp, __unused GError** error)
          */
         if (netdef->type == NETPLAN_DEF_TYPE_VLAN && backend == NETPLAN_BACKEND_NM) {
             netdef->vlan_link = netplan_netdef_new(npp, scalar(missing->node), NETPLAN_DEF_TYPE_NM_PLACEHOLDER_, NETPLAN_BACKEND_NM);
+            g_hash_table_iter_remove(&iter);
+        }
+
+        /* VETH case: NetworkManager doesn't enforce the existence of the veth peer.
+         * NM will create one connection for each veth in the pair. In this case, due to our integration with
+         * NM (netplan-everywhere), we can't enforce the existence of both peers at a given moment because
+         * they might be created one after the other.
+         * When we find that the peer is missing, we create a temporary one using the placeholder type. That is necessary
+         * so we can generate the keyfile referring to the correct peer name, even though it still doesn't exist.
+         */
+        if (netdef->type == NETPLAN_DEF_TYPE_VETH && backend == NETPLAN_BACKEND_NM) {
+            netdef->veth_peer_link = netplan_netdef_new(npp, scalar(missing->node), NETPLAN_DEF_TYPE_NM_PLACEHOLDER_, NETPLAN_BACKEND_NM);
             g_hash_table_iter_remove(&iter);
         }
     }
--- a/src/validation.c
+++ b/src/validation.c
@@ -393,6 +393,11 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, GErr
             goto netdef_grammar_error;
     }
 
+    if (nd->type == NETPLAN_DEF_TYPE_VETH) {
+        if (!nd->veth_peer_link)
+            return yaml_error(npp, NULL, error, "%s: virtual-ethernet missing 'peer' property", nd->id);
+    }
+
     if (nd->ip6_addr_gen_mode != NETPLAN_ADDRGEN_DEFAULT && nd->ip6_addr_gen_token)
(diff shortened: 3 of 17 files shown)


slyon:
Schema wise, I like the proposal from the PR description. I think it makes sense to introduce a new interface type for `veths`, as networkd is using a top-level `[NetDev].Kind=veth` setting and NetworkManager is also using a top-level `[connection].type=veth` setting. I'm a bit undecided about the new `peer` setting, though. Calling correlated veths "peers" is the common term across the networking stack. OTOH, in Netplan we're using the term "link" for SRIOV, VLAN & VXLAN, and we could use it here, too, for consistency.
I'm leaning towards still calling it `peer`, as `link` is more of a one-sided relationship between interfaces (parent-child), while "peer" is linking interfaces on the same level.

@vorlonofportland What's your stance on this?

```yaml
network:
  veths:
    IFACE1:
      peer: IFACE2
      ...
    IFACE2:
      peer: IFACE1
      ...

File: tests/generator/test_veths.py [hunk 1]
```diff
@@ -0,0 +1,222 @@
...
+  renderer: networkd
+  veths:
+    veth0:
+      peer: veth1
+    veth1:
+      peer: abc''', expect_fail=True)
+
+        self.assertIn('veth1: interface \'abc\' is not defined', out)
+
+    def test_veth_peer_is_anothers_interface_peer_already(self):
```
slyon, line 210: nitpick (non-blocking): I guess we could combine some of those tests into a `_CommonTests()` class (where the behavior does not differ between the backend renderers), to make equal tests run on both backends. Cf. `tests/generator/test_tunnels.py`

tests/integration/veths.py
```diff
@@ -0,0 +1,85 @@
...
+import subprocess
+import unittest
+
+from base import IntegrationTestsBase, test_backends
+
+
+class _CommonTests():
+
+    def test_create_veth_pair(self):
+        self.addCleanup(subprocess.call, ['ip', 'link', 'delete', 'veth0'], stderr=subprocess.DEVNULL)
```
slyon, line 32: todo: We want to cleanup the `veth1`, too. Here and in the test below.

tests/integration/veths.py
```diff
@@ -0,0 +1,85 @@
...
+      peer: veth1
+      addresses:
+        - 192.168.123.123/24
+        - 1234:FFFF::42/64
+    veth1:
+      dhcp4: false
+      dhcp6: false
+      peer: veth0''' % {'r': self.backend})
+        self.generate_and_settle(['veth0', 'veth1'])
+        self.assert_iface_up('veth0', ['inet 192.168.123.123/24', 'inet6 1234:ffff::42/64'])
```
slyon, line 69: suggestion: We might want to use the new `self.iface_json()` method here to check for the IP addresses, as that is the more future-proof way of doing it (iproute2's JSON interface is expected to stay stable, the CLI output not necessarily). See `tests/integration/tunnels.py:test_tunnel_vxlan()`.

tests/parser/test_keyfile.py
```diff
@@ -1894,3 +1894,39 @@ def test_keyfile_ethernet_random_cloned_mac_address(self):
...
+      dhcp4: true
+      peer: "veth-peer2"
+      networkmanager:
+        uuid: "{}"
+        name: "veth-peer1"
+        passthrough:
+          connection.interface-name: "veth-peer1"
+'''.format(UUID, UUID)})
+
+    def test_veth_without_key_pair(self):
```
slyon, line 1924: nitpick (non-blocking): I assume this should be called `veth_peer` instead of `key_pair`?

src/parse.c
```diff
@@ -3257,6 +3295,12 @@ process_missing_ids(NetplanParser* npp, __unused GError** error)
             netdef->vlan_link = netplan_netdef_new(npp, scalar(missing->node), NETPLAN_DEF_TYPE_NM_PLACEHOLDER_, NETPLAN_BACKEND_NM);
             g_hash_table_iter_remove(&iter);
         }
+
+        /* VETH case: NetworkManager doesn't enforce the existence of the veth peer */
```
slyon, line 3299: suggestion: We might want to extend that comment a bit, to describe in more detail the relaxation of validation for this `NETPLAN_DEF_TYPE_NM_PLACEHOLDER_` special case. Future-us would be happy about that context :)

src/networkd.c
```diff
@@ -524,6 +524,21 @@ write_netdev_file(const NetplanNetDefinition* def, const char* rootdir, const ch
...
             break;
 
+        case NETPLAN_DEF_TYPE_VETH:
+            /*
+             * Only one .netdev file is required to create the veth pair.
+             * To select what netdef we are going to use, we sort both names, get the first one,
+             * and, if the selected name is the name of the netdef being written, we generate
+             * the .netdev file. Otherwise we skip the netdef.
+             */
+            gchar* first = g_strcmp0(def->id, def->veth_peer_link->id) < 0 ? def->id : def->veth_peer_link->id;
```
slyon, line 534: praise: That's a nice approach!

src/validation.c
```diff
@@ -644,3 +649,31 @@ validate_default_route_consistency(__unused const NetplanParser* npp, GHashTable
     g_slist_free_full(defroutes, g_free);
     return ret;
 }
+
+gboolean
+validate_veth_pair(__unused const NetplanState* np_state, const NetplanNetDefinition* netdef, GError** error)
+{
+
+    NetplanNetDefinition* veth_peer = netdef->veth_peer_link;
+    if (veth_peer && veth_peer->type != NETPLAN_DEF_TYPE_NM_PLACEHOLDER_) {
```
slyon, line 658: suggestion: Some brief comment/reference about the `NETPLAN_DEF_TYPE_NM_PLACEHOLDER_` quirk would probably be helpful in the future.

slyon (approved):
Thank you, this is looking very good implementation wise!

I added a few inline remarks, but no major changes required, IMO.

+1 (but let's block merging util we have the new YAML schema approved)

vorlonofportland:
`veth` is short for `virtual ethernet`.  Elsewhere in the schema we spell out `ethernets`.  Should this be `virtual-ethernets` instead?

What if they were just called `ethernets`, and ethernets with the mandatory `peer` key were automatically veths, and ethernets without this key were physical devices?

daniloegea:
Regarding the name, `veth` is the term used by both backends and, as far as I know, it's the most common term people will use to refer to this type of devices. But apart from that, I don't have a strong opinion against using `virtual-ethernets`.

One argument against generalizing `ethernets` and allowing it to also create `veths` is that it would allow users to try to use configuration intended to be used only for physical devices with virtual devices, such as `match` for example.

tests/integration/veths.py
daniloegea: That's not necessary, deleting one of the veths makes the other one go away.

slyon:
> Regarding the name, `veth` is the term used by both backends and, as far as I know, it's the most common term people will use to refer to this type of devices. But apart from that, I don't have a strong opinion against using `virtual-ethernets`.

I'd also be fine with `virtual-ethernets`, as indeed we're usually using the more verbose spelling in Netplan.

> One argument against generalizing `ethernets` and allowing it to also create `veths` is that it would allow users to try to use configuration intended to be used only for physical devices with virtual devices, such as `match` for example.

Exactly. This is a problem Danilo and I discussed previously. `ethernets` are pure physical devices, allowing for settings such as `match`, `set-name`, `*-offload`, etc, while `virtual-ethernets`/`veths` are pure virtual devices, where the NetplanID is the interface name and it can have the `peer` setting. `peer` and `match` (or other physical-devices properties) are mutually exclusive, and it's hard to check this condition at the YAML parsing stage, as those settings might be defined in arbitrary order (could be checked in the validation stage, though). But mixing those physical vs virtual properties could be confusing to our users.

So I'd suggest going with `virtual-ethernets`. Are you fine with that @vorlonofportland ?

tests/integration/veths.py
```diff
@@ -0,0 +1,85 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+import subprocess
+import unittest
+
+from base import IntegrationTestsBase, test_backends
```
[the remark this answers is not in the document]
daniloegea: hm I tried that and apparently it will not work because the way we run the integration tests. We basically call `python3 veths.py` so the script is the `__main__` module and relative imports behave differently. But I might be missing something here...

tests/integration/veths.py
slyon: Alright, wfm.

vorlonofportland:
Yes, +1 from me for `virtual-ethernets` then.

daniloegea:
I replaced `veths` by `virtual-ethernets`.

The `debci` test is failing because api.ftp-master.debian.org is having problems apparently...
```
curl: (7) Failed to connect to api.ftp-master.debian.org port 443 after 177 ms: Connection refused
```

slyon:
+1 Thanks for the renaming. I've adopted the PR description accordingly for `virtual-ethernets`. The DebCi test failure looks unrelated.

LGTM.

PR #369, opened 2023-06-16
WPA3 support (LP: #2023238)

daniloegea:
Related to https://bugs.launchpad.net/netplan/+bug/2023238

## Description

The following wpa_supplicant configuration is required to connect to a WPA3 network:

```
network={
	ssid="WPA3-Network"
	psk="abcdefgh"
	key_mgmt=SAE
	ieee80211w=2
}
```
From the wpa_supplicant.conf manual:

```
# SAE = Simultaneous authentication of equals; pre-shared key/password -based
#	authentication with stronger security than WPA-PSK especially when using
#	not that strong password; a.k.a. WPA3-Personal
...
...
# ieee80211w: whether management frame protection is enabled
# 0 = disabled (default unless changed with the global pmf parameter)
# 1 = optional
# 2 = required
# The most common configuration options for this based on the PMF (protected
# management frames) certification program are:
# PMF enabled: ieee80211w=1 and key_mgmt=WPA-EAP WPA-EAP-SHA256
# PMF required: ieee80211w=2 and key_mgmt=WPA-EAP-SHA256
# (and similarly for WPA-PSK and WPA-PSK-SHA256 if WPA2-Personal is used)
# WPA3-Personal-only mode: ieee80211w=2 and key_mgmt=SAE
```

It was tested with a WPA3 only network configured with `hostapd` using both renderers.

The integration with Network Manager (netplan everywhere) was also tested.

## Checklist


src/abi.h
```diff
@@ -106,6 +106,7 @@ typedef enum {
     NETPLAN_AUTH_KEY_MANAGEMENT_NONE,
     NETPLAN_AUTH_KEY_MANAGEMENT_WPA_PSK,
     NETPLAN_AUTH_KEY_MANAGEMENT_WPA_EAP,
+    NETPLAN_AUTH_KEY_MANAGEMENT_WPA_SAE,
```
slyon, line 109: ABI checker does not mention any issue here... But this might be changing the `NETPLAN_AUTH_KEY_MANAGEMENT_8021X` value implicitly.

Can we move the new `_WPA_SAE` value down below `_8201X`, this way only the `_MAX` value would be affected in the ABI.

slyon (approved):
Thanks, LGTM!

I left one little inline comment, which we might want to resolve before merging.

src/abi.h
daniloegea: Done

PR #371, opened 2023-06-23
NetworkManager: Wireguard private key flag support

daniloegea:
## Description

Network Manager supports storing the Wireguard private key in an external keychain agent, like the kdewallet. In this cases, it will emit the following configuration in the keyfile: `wireguard.private-key-flags=1` and omit the private key, which will be stored somewhere else.

The new key will look like this:

```yaml
network:
  version: 2
  tunnels:
    wg-tunnel:
      renderer: NetworkManager
      addresses:
      - "10.20.30.1/24"
      mode: "wireguard"
      port: 51820
      keys:
        private-key-flags:
        - agent-owned
```

Adding this options is one way of solving the issue described here https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2024661

## Checklist

### DIFF
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -1642,6 +1642,39 @@ The specific settings for tunnels are defined below.
     > `systemd-networkd` backend (v242+) is used, this can also be an
     > absolute path to a file containing the private key.
 
+  - **private-key-flags** (sequence of scalars) – since **0.107**
+
+    > Private key flags used by Network Manager. Possible values are:
+    > `agent-owned`, `not-saved` and `not-required`.
+    >
+    > `agent-owned`: a user-session secret agent is responsible for
+    > providing and storing this secret.
+    >
+    > `not-saved`: this secret should not be saved but should be
+    > requested from the user each time it is required.
+    >
+    > `not-required`: this flag hints that the secret is not required
+    > and should not be requested from the user.
+
+    Example:
+
+    ```yaml
+    network:
+      renderer: NetworkManager
+      tunnels:
+        wg0:
+          mode: wireguard
+          port: 5182
+          key:
+            private-key-flags:
+              - agent-owned
+          peers:
+            - keys:
+                public: rlbInAj0qV69CysWPQY7KEBnKxpYCpaWqOs/dLevdWc=
+              allowed-ips: [0.0.0.0/0, "2001:fe:ad:de:ad:be:ef:1/24"]
+              keepalive: 23
+              endpoint: 1.2.3.4:5
+      ```
 - **keys** (scalar or mapping)
 
   > Alternate name for the `key` field. See above.
--- a/src/abi.h
+++ b/src/abi.h
@@ -132,6 +132,14 @@ typedef struct authentication_settings {
     char* phase2_auth;  /* netplan-feature: auth-phase2 */
 } NetplanAuthenticationSettings;
 
+typedef enum {
+    NETPLAN_KEY_FLAG_NONE           = 0,
+    NETPLAN_KEY_FLAG_AGENT_OWNED    = 1<<0,
+    NETPLAN_KEY_FLAG_NOT_SAVED      = 1<<1,
+    NETPLAN_KEY_FLAG_NOT_REQUIRED   = 1<<2,
+    NETPLAN_KEY_FLAG_MAX_
+} NetplanKeyFlags;
+
 typedef struct ovs_controller {
     char* connection_mode;
     GArray* addresses;
@@ -383,4 +391,6 @@ struct netplan_net_definition {
 
     /* True if "networkmanager" settings are present */
     gboolean has_backend_settings_nm;
+
+    guint tunnel_private_key_flags;
 };
--- a/src/nm.c
+++ b/src/nm.c
@@ -344,6 +344,9 @@ write_wireguard_params(const NetplanNetDefinition* def, GKeyFile *kf, GError** e
             g_key_file_set_string(kf, "wireguard", "private-key", def->tunnel.private_key);
     }
 
+    if (def->tunnel_private_key_flags != NETPLAN_KEY_FLAG_NONE)
+        g_key_file_set_uint64(kf, "wireguard", "private-key-flags", def->tunnel_private_key_flags);
+
     if (def->tunnel.port)
         g_key_file_set_uint64(kf, "wireguard", "listen-port", def->tunnel.port);
     if (def->tunnel.fwmark)
--- a/src/parse.c
+++ b/src/parse.c
@@ -1043,6 +1043,29 @@ handle_access_point_band(NetplanParser* npp, yaml_node_t* node, __unused const v
     return TRUE;
 }
 
+static gboolean
+handle_tunnel_key_flags(NetplanParser* npp, yaml_node_t* node, __unused const void* _, GError** error)
+{
+    for (yaml_node_item_t *i = node->data.sequence.items.start; i < node->data.sequence.items.top; i++) {
+        yaml_node_t *entry = yaml_document_get_node(&npp->doc, *i);
+        gboolean found = FALSE;
+        assert_type(npp, entry, YAML_SCALAR_NODE);
+
+        for (int i = 1; i < NETPLAN_KEY_FLAG_MAX_; i <<= 1) {
+            if (!g_ascii_strcasecmp(scalar(entry), netplan_key_flags_name(i))) {
+                npp->current.netdef->tunnel_private_key_flags |= i;
+                found = TRUE;
+            }
+        }
+
+        if (!found)
+            return yaml_error(npp, node, error,
+                              "Key flag '%s' is not supported. Valid values are \"agent-owned\", \"not-saved\" and \"not-required\"",
+                              scalar(entry));
+    }
+    return TRUE;
+}
+
 /* Keep in sync with ap_nm_backend_settings_handlers */
 static const mapping_entry_handler nm_backend_settings_handlers[] = {
     {"name", YAML_SCALAR_NODE, {.generic=handle_netdef_backend_settings_str}, netdef_offset(backend_settings.name)},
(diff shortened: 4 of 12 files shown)


doc/netplan-yaml.md
```diff
@@ -1771,6 +1771,39 @@ WireGuard specific keys:
...
+
+    ```yaml
+    network:
+      renderer: NetworkManager
+      tunnels:
+        wg0:
+          mode: wireguard
+          port: 5182
+          private-key-flags:
+            - agent-owned
```
slyon, line 1798: Regarding new schema:
This seems to be a rather NetworkManager specific settings, I feel like it should therefore live in the `nm_backend_settings_handlers` instead. Also, I wonder if we could/should make this a bit more generic, by introducing a generic "networkmanager.flags" setting, which might contain other (future) flags, then using the "validation" stage to verify those flags make sense for the given interface definition implicitly (c.f. `wakeonwlan` setting). E.g.:

   ```yaml
    network:
      renderer: NetworkManager
      tunnels:
        wg0:
          mode: wireguard
          networkmanager:
            flags: [private-key-agent-owned]
          port: 5182
          peers:
            - keys:
                public: rlbInAj0qV69CysWPQY7KEBnKxpYCpaWqOs/dLevdWc=
              allowed-ips: [0.0.0.0/0, "2001:fe:ad:de:ad:be:ef:1/24"]
              keepalive: 23
              endpoint: 1.2.3.4:5
      ```

What do you think?

daniloegea: That works for me. It could be moved to `backend_settings`, but it's used in the middle of the netdef struct, so adding a new variable there will change the struct layout. Is that fine?

daniloegea: On a second thought, I think having a generic `flags` key might be tricky to maintain. We'll have to have our own encoding. What if a flag for a given context have the same value as another flag from a different context in Network Manager?

I think it's fine to move the private-key-flags to the backend specific data though.

daniloegea:
I moved the new key to the `networkmanager` section and rebased.

daniloegea:
Having the flags in the `peers.keys` path is not accurate. Those flags are about the private key used by the local connection, not its peers.

After checking nm-settings(5), I've found many more places where the same set of flags (NMSettingSecretFlags) are used apart from `wireguard.private-key-flags` (I missed that before):
`802-1x.ca-cert-password-flags`, `802-1x.client-cert-password-flags`, `802-1x.password-flags`, `802-1x.password-raw-flags`, `802-1x.phase2-ca-cert-password-flags`, `802-1x.phase2-client-cert-password-flags`, `802-1x.phase2-private-key-password-flags`, `802-1x.pin-flags`, `802-1x.private-key-password-flags`, `adsl.password-flags`, `cdma.password-flags`, `gsm.password-flags`, `gsm.pin-flags`, `macsec.mka-cak-flags`, `pppoe.password-flags`, `802-11-wireless-security.leap-password-flags`, `leap-password-flags.psk-flags` and `leap-password-flags.wep-key-flags`.

For reference, these are the flags:

```c
/**
 * NMSettingSecretFlags:
 * @NM_SETTING_SECRET_FLAG_NONE: the system is responsible for providing and
 * storing this secret (default)
 * @NM_SETTING_SECRET_FLAG_AGENT_OWNED: a user secret agent is responsible
 * for providing and storing this secret; when it is required agents will be
 * asked to retrieve it
 * @NM_SETTING_SECRET_FLAG_NOT_SAVED: this secret should not be saved, but
 * should be requested from the user each time it is needed
 * @NM_SETTING_SECRET_FLAG_NOT_REQUIRED: in situations where it cannot be
 * automatically determined that the secret is required (some VPNs and PPP
 * providers don't require all secrets) this flag indicates that the specific
 * secret is not required
 *
 * These flags indicate specific behavior related to handling of a secret.  Each
 * secret has a corresponding set of these flags which indicate how the secret
 * is to be stored and/or requested when it is needed.
 *
 **/
typedef enum /*< flags >*/ {
    NM_SETTING_SECRET_FLAG_NONE         = 0x00000000,
    NM_SETTING_SECRET_FLAG_AGENT_OWNED  = 0x00000001,
    NM_SETTING_SECRET_FLAG_NOT_SAVED    = 0x00000002,
    NM_SETTING_SECRET_FLAG_NOT_REQUIRED = 0x00000004

    /* NOTE: if adding flags, update nm-core-internal.h as well */
} NMSettingSecretFlags;
```

That said, I think it makes sense to have a more generic key for setting these flags. Maybe something like `networkmanager.secret-flags` so we could use it to set flags for all those types of connections I listed above.

slyon:
> That said, I think it makes sense to have a more generic key for setting these flags. Maybe something like `networkmanager.secret-flags` so we could use it to set flags for all those types of connections I listed above.

Thanks for looking up the details! So that would be back to my original suggestion of `networkmanager.(secret-)flags` and be pretty much a 1:1 representation of the setting in NetworkManager, but it would still keep the secret-flags far away from the actual secrets in Netplan's YAML schema. I wonder if we can do better than that?

We don't need to keep a 1:1 representation of Netplan <-> NetworkManager settings. Can we think of any (hypothetical/future) usecase where such flags could be used for another backend renderer, like networkd or OVS? If so that would be an argument against moving it into the `networkmanager.` backend settings.

I feel like those secret-flags should be tightly bundled with the secrets itself, which are bundled under the `key[s]` setting in Netplan. You're right that it shouldn't be in `peers.keys.flags`, as this is about the private key for the local connection. Rather, it should be in `tunnels.wg0.keys.flags`, which uses the same datastructure/schema as `peers.keys`.

slyon:
> Should we call it `private-key-flags` though?

That works for me. It's close to the actual secret (if any) and could also be used to define (future) flags for other backends. So let's try to get @vorlonofportland review of this schema proposal:
```yaml
 network:
   renderer: NetworkManager
   tunnels:
     wg0:
       mode: wireguard
       key:
         private-key-flags: [agent-owned]
       peers:
         [...]
```

vorlonofportland:
This looks good to me, thanks.

slyon:
> This looks good to me, thanks.

Thanks! I've added the "schema ok" tag and will do a review of the implementation details next.

slyon:
I wonder if we can use the same schema for WPA-PSK credentials stored in an agent?
See https://discourse.ubuntu.com/t/call-for-testing-networkmanager-yaml-settings/32420/8 (point 3) and https://developer-old.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html => `psk-flags`

src/abi.h
```diff
@@ -155,6 +163,7 @@ typedef struct netplan_backend_settings {
     char *stable_id;
     char *device;
     GData* passthrough; /* See g_datalist* functions */
+    guint private_key_flags;
```
slyon, line 166: This should probably move into `NetplanAuthenticationSettings`

src/nm.c
```diff
@@ -344,6 +344,9 @@ write_wireguard_params(const NetplanNetDefinition* def, GKeyFile *kf, GError** e
             g_key_file_set_string(kf, "wireguard", "private-key", def->tunnel.private_key);
     }
 
+    if (def->backend_settings.private_key_flags != NETPLAN_PRIVATE_KEY_FLAG_NONE)
+        g_key_file_set_uint64(kf, "wireguard", "private-key-flags", def->backend_settings.private_key_flags);
```
slyon, line 348: thought: I wonder if the same data-structure could be used for `[wifi-security].psk-flags`?

src/validation.c
```diff
@@ -202,7 +202,7 @@ validate_tunnel_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, yaml
         return yaml_error(npp, node, error, "%s: missing or invalid 'mode' property for tunnel", nd->id);
 
     if (nd->tunnel.mode == NETPLAN_TUNNEL_MODE_WIREGUARD) {
-        if (!nd->tunnel.private_key)
+        if (!nd->tunnel.private_key && (nd->backend_settings.private_key_flags & NETPLAN_PRIVATE_KEY_FLAG_AGENT_OWNED) == 0)
```
slyon, line 205: question: Doesn't that apply to all current values: `agent-owned`, `not-saved` and `not-required`? IIUC, we're just checking for `agent-owned` here...

src/parse.c
```diff
@@ -1043,12 +1043,36 @@ handle_access_point_band(NetplanParser* npp, yaml_node_t* node, const void* _, G
     return TRUE;
 }
 
+static gboolean
+handle_wireguard_private_key_flags(NetplanParser* npp, yaml_node_t* node, const void* _, GError** error)
```
slyon, line 1047: thought: Maybe we can move some of the logic into a `handle_generic_key_flags` function, so we can handle wireguard-private-key-flags and wifi-psk-flags alike.

slyon (changes requested):
Thanks, this is looking good overall. The docs & implementation need to be adopted to the agreed upon YAML schema. And I wonder if we could use the same data-structure for wifi-security.psk-flags, too?

I added some remarks inline.

src/validation.c
daniloegea: yeah that makes sense, all the private key flags suggest that the private key definition is not required in the file...

src/abi.h
daniloegea: `NetplanAuthenticationSettings`only has Wifi related configuration. I don't think it should be moved there, it's more like we'll need different `private_key_flags` variables in different places for each key we want to correlated the flags with.

This one in particular should probably be moved to `netplan_net_definition.tunnel`.

src/nm.c
daniloegea: If only wireguard and the the psk password will have flags we could use the same variable, as the netdef will never be wifi and wireguard at the same time. But if we want to extend it to all the other secrets we will need multiple variables...

The case for wifi é covered by the passthrough data, only the wireguard one was causing issues due to how NM parses the keyfile and how we were generating it. Do we want to extend it to wifi right now?

src/abi.h
slyon: Yes, you're right!
Also, we need to take special care about the `netplan_net_definition.tunnel` struct (and `struct NetplanAuthenticationSettings`, too) as they are part of the bigger `struct netplan_net_definition`. So adding new fields in-between would break ABI.

We probably need to use some quirk as done with `tunnel_ttl`, adding a new field to the end of the top-level struct (+ a relevant XXX/TODO comment), and refactor it eventually, once we bump the library's SONAME.

PR #373, opened 2023-06-24
Documentation: make dhcp4 & dhcp6 docs match website homepage

triatic:


slyon (changes requested):
Hey @triatic thank you for your contribution to Netplan.

Could you please give some more context on this change? AFAIU the `dhcp4` & `dhcp6` settings are off by default on the networkd AND NetworkManager backends. so I wonder where this change comes from? What do you mean by "match website homepage"?

TIA.

slyon:
> "Off by default, unless NetworkManager is the renderer and no interface definitions are given." would work.

Well, strictly speaking yes. But OTOH this snippet is a special case, which should be described elsewhere in the documentation (maybe it is already? I need to search), instead of leaking that information into the `dhcp4`/`dhcp6` settings, as it might confuse people. I'd rather like to keep those as brief as possible.

```yaml
network:
  version: 2
  renderer: NetworkManager
```

I'll try to create a PR for the netplan.io website to update the working and search for a place where we could describe that "NetworkManager manages everything" special case.

slyon:
I'm closing this in favour of https://github.com/canonical/netplan/pull/378

PR #374, opened 2023-06-28
CI: initial NetworkManager autopkgtests

slyon:
## Description
We're testing NetworkManager from mantic-proposed against a local build of netplan.io from this repository.

So far we can only run NetworkManager's autopkgtests which allow for 'isolation-container'.

FR-4415

## Checklist

### DIFF
--- a/.github/workflows/network-manager.yml
+++ b/.github/workflows/network-manager.yml
@@ -1,0 +1,55 @@
+name: NetworkManager Autopkgtest
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the main branch
+on:
+  push:
+    branches: [ main, 'stable/**' ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


daniloegea (approved):
I think it's fine hard coding `mantic` there. I though of using `ubuntu-daily:devel` so we wouldn't need to update the yaml, but I don't know if it's safe to use it...

slyon:
> I think it's fine hard coding `mantic` there. I though of using `ubuntu-daily:devel` so we wouldn't need to update the yaml, but I don't know if it's safe to use it...

Yes, that was my first attempt. But `autopkgtest-build-lxd` will then still create an `autopkgtest/ubuntu/mantic/amd64` alias for the image, so we need to hard-code "mantic" anyway. Therefore, I decided to keep it the same in both places.

PR #375, opened 2023-07-06
Added pciid match support

fatihusta:
## Description

Netplan doesn't match mac address of ethernet when using bond. So member ethernet doesn't up. 

This [PR](https://github.com/canonical/netplan/pull/278) fixes ethernet up problem, but still prints warning message(below).

I try to add PCI ID match support for better renaming to the ethernet.

```
[]
Cannot find unique matching interface for enp152s0f0np0
[]
Cannot find unique matching interface for eno12399np0
[]
Cannot find unique matching interface for enp152s0f1np1
[]
Cannot find unique matching interface for eno12409np1
```

## Checklist

Also may fixes this issue: https://bugs.launchpad.net/netplan/+bug/1608447

### DIFF
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -105,6 +105,12 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
     > A sequence of globs is supported, any of which must match.
     > Matching on driver is *only* supported with networkd.
 
+  - **pciid** (scalar) – since **0.107**
+
+    > The PCI ID, corresponding to the `ID_PATH` udev properity. (`0000:98:00.1`)
+    > Matching on pci id is *only* supported with networkd.
+    > `udevadm info /sys/class/net/DEVICE_NAME`
+
   Examples:
 
   - All cards on second PCI bus:
@@ -143,6 +149,15 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
           match:
             driver: ["bcmgenet", "smsc*"]
             name: en*
+    ```
+
+  - Matching with PCI ID `0000:98:00.1`:
+    ```yaml
+    network:
+      ethernets:
+        nic0:
+          match:
+            pciid: 0000:98:00.1
     ```
 
 - **set-name** (scalar)
--- a/include/netplan.h
+++ b/include/netplan.h
@@ -125,7 +125,7 @@ NETPLAN_PUBLIC gboolean
 netplan_netdef_has_match(const NetplanNetDefinition* netdef);
 
 NETPLAN_PUBLIC gboolean
-netplan_netdef_match_interface(const NetplanNetDefinition* netdef, const char* name, const char* mac, const char* driver_name);
+netplan_netdef_match_interface(const NetplanNetDefinition* netdef, const char* name, const char* mac, const char* driver_name, const char* pciid);
 
 /********** Old API below this ***********/
 
--- a/netplan/cli/utils.py
+++ b/netplan/cli/utils.py
@@ -176,7 +176,6 @@ def get_interface_macaddress(interface):
 
 def get_interface_pciid(interface):  # pragma: nocover (covered in autopkgtest)
     devdir = os.path.join('/sys/class/net', interface)
-    pciid = None
     try:
         uevent = os.path.realpath(os.path.join(devdir, 'device', 'uevent'))
         if os.path.exists(uevent):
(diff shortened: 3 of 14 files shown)


slyon:
Sorry, this is taking much longer than anticipated. I like the idea of having `pciid` matching capabilities. But it touches core components of Netplan and I need to find enough time to do a proper review on this.

include/netplan.h
```diff
@@ -125,7 +125,7 @@ NETPLAN_PUBLIC gboolean
 netplan_netdef_has_match(const NetplanNetDefinition* netdef);
 
 NETPLAN_PUBLIC gboolean
-netplan_netdef_match_interface(const NetplanNetDefinition* netdef, const char* name, const char* mac, const char* driver_name);
+netplan_netdef_match_interface(const NetplanNetDefinition* netdef, const char* name, const char* mac, const char* driver_name, const char* pciid);
```
slyon, line 128: todo: We cannot just change public API like this, but would rather need to introduce a new `netplan_netdef_match_interface2()` function. Keeping the original function backwards compatible.

src/abi.h
```diff
@@ -249,6 +249,7 @@ struct netplan_net_definition {
         char* driver;
         char* mac;
         char* original_name;
+        char* pciid;
```
slyon, line 252: todo: This has the potential of breaking ABI and we'd need to move the new field to the very end of the `netplan_net_definition` struct, e.g. called `match_pciid`.

doc/netplan-yaml.md
```diff
@@ -105,6 +105,12 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
     > A sequence of globs is supported, any of which must match.
     > Matching on driver is *only* supported with networkd.
 
+  - **pciid** (scalar) – since **0.107**
+
+    > The PCI ID, corresponding to the `ID_PATH` udev properity. (`0000:98:00.1`)
+    > Matching on pci id is *only* supported with networkd.
+    > `udevadm info /sys/class/net/DEVICE_NAME`
```
slyon, line 112: nitpick (non-blocking): Version needs to be bumped.

question: Also, is there any possibility of doing the same for the NetworkManager backend? The `match` stanza is a very central piece of Netplan's configuration and we should try not to diverge between the backends here.

src/parse.c
```diff
@@ -857,6 +857,7 @@ static const mapping_entry_handler match_handlers[] = {
     {"driver", YAML_NO_NODE, {.variable=handle_match_driver}},
     {"macaddress", YAML_SCALAR_NODE, {.generic=handle_netdef_mac}, netdef_offset(match.mac)},
     {"name", YAML_SCALAR_NODE, {.generic=handle_netdef_id}, netdef_offset(match.original_name)},
+    {"pciid", YAML_SCALAR_NODE, {.generic=handle_netdef_str}, netdef_offset(match.pciid)},
```
slyon, line 860: todo: We'd probably want to have a more specific handler here, that validates the input string is an actual PCI ID.

tests/generator/test_ethernets.py
```diff
@@ -164,6 +164,27 @@ def test_eth_match_by_mac_rename(self):
         self.assert_nm(None)
         self.assert_nm_udev(NM_UNMANAGED % 'lom1' + NM_UNMANAGED_MAC % '11:22:33:44:55:66')
 
+    def test_eth_match_by_pciid_rename(self):
```
slyon, line 167: nitpick: Thanks for providing some test cases! I'd love to see an integration test along the lines of `tests/integration/ethernets.py:test_rename_interfaces()`, too (if possible).

netplan/cli/utils.py
```diff
@@ -174,14 +174,42 @@ def get_interface_macaddress(interface):
     return link.get('addr', '')
 
 
+def get_interface_pciid(interface):  # pragma: nocover (covered in autopkgtest)
```
slyon, line 177: question: Can we apply some test-mocking to get this net method covered by unit tests?

slyon (changes requested):
My apologies, this PR fell through the cracks. I've now done an initial high-level round of review, as this is touching quite some core parts of Netplan and the original issue seems to be fixed as of recently (https://github.com/canonical/netplan/pull/503).

Please let me know if you're still interested in driving this forward, so I can do a more detailed review and talk to our architect about the new YAML schema.

PR #376, opened 2023-07-06
parse-nm: handle cloned-mac-address special cases

daniloegea:
The property "cloned-mac-address" might contain one of these special values: "preserve", "permanent", "random" and "stable". Currently we accept only a valid MAC address as value.

This change will leave the cloned-mac-address property in the networkmanager.passthrough section if it contains one of these values and will try to validate it otherwise.

Related to LP: #2026230

## Description

This is a quick fix for the problem reported here https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/2026230

*TODO*: accept these values as part of Netplan's grammar and create an autopkgtest for network-manager.

## Checklist


slyon (approved):
Thank you Danilo! LGTM :eyes:

PR #377, opened 2023-07-10
Improve autopkgtest stability with systemd 253 & iproute 6.4

slyon:
## Description
    Starting with the systemd v253 update NetworkManager started taking
    implicit control of certain interfaces, blocking itself (after a restart)
    from managing those interface through its configure netplan-* connection
    profile.

https://bugs.debian.org/1039071

    iproute2 v6.4 changed its CLI output, the JSON output is expected to
    stay stable, so we should migrate to using iface_json() over time.

https://bugs.debian.org/1040004

Also, resolves a race condition where the following text is already gone from the buffer at the time when we check the assertions. We see such failures a lot in the Ubuntu autopkgtests:
```
Do you want to keep these settings?
Press ENTER before the timeout to accept the new configuration
Changes will revert in {} seconds
```

## Checklist

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -31,6 +31,7 @@ import unittest
 import shutil
 import gi
 import glob
+import json
 
 # make sure we point to libnetplan properly.
 os.environ.update({'LD_LIBRARY_PATH': '.:{}'.format(os.environ.get('LD_LIBRARY_PATH'))})
@@ -277,6 +278,15 @@ class IntegrationTestsBase(unittest.TestCase):
             self.poll_text(dnsmasq_log, 'IPv6 router advertisement enabled')
         else:
             self.poll_text(dnsmasq_log, 'DHCP, IP range')
+
+    def iface_json(self, iface: str) -> dict:
+        '''Return iproute2's (detailed) JSON representation'''
+        out = subprocess.check_output(['ip', '-j', '-d', 'a', 'show', 'dev', iface],
+                                      text=True)
+        json_dict = json.loads(out)
+        if json_dict:
+            return json_dict[0]
+        return {}
 
     def assert_iface(self, iface, expected_ip_a=None, unexpected_ip_a=None):
         '''Assert that client interface has been created'''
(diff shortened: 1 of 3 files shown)


slyon:
The RPM build CI failure on fedora:rawhide seems unrelated.

tests/integration/base.py
```diff
@@ -278,6 +279,15 @@ def start_dnsmasq(self, ipv6_mode, iface):
         else:
             self.poll_text(dnsmasq_log, 'DHCP, IP range')
 
+    def iface_json(self, iface: str) -> dict:
+        '''Return iproute2's (detailed) JSON representation'''
+        out = subprocess.check_output(['ip', '-j', '-d', 'a', 'show', 'dev', iface],
```
daniloegea, line 284: question: should we handle a possible call against an interface that doesn't exist?

tests/integration/base.py
```diff
@@ -278,6 +279,15 @@ def start_dnsmasq(self, ipv6_mode, iface):
...
+                                      text=True)
+        json_dict = json.loads(out)
+        if len(json_dict) == 1:
```
daniloegea, line 287: suggestion: as empty arrays are evaluated to False, this could be simplified to `if json_dict:`

tests/integration/base.py
```diff
@@ -278,6 +279,15 @@ def start_dnsmasq(self, ipv6_mode, iface):
         else:
             self.poll_text(dnsmasq_log, 'DHCP, IP range')
 
+    def iface_json(self, iface: str) -> dict:
```
daniloegea, line 282: question: does this change affect other parts of netplan, like netplan status?

daniloegea (approved):
lgtm, just left some non-blocking questions/suggestions.

My main question was: does the iproute2 change affect other parts of the code, like netplan status?

tests/integration/base.py
slyon: I don't think that's necessarily needed. This is only part of the integration tests, so all we could do is: bail out.
If the interface doesn't exist, this will crash with an error message, too, as-is.

IMHO, no need to over-engineer this, but rather keep it small and simple in the test framework. All the input is controlled, as it originates from our own test scripts, not random user input.

tests/integration/base.py
slyon: No. `netplan status` has been using iproute2's JSON integration from the start. This change only affects the base class of Netplan's integration tests.

But we want to migrate more parts of our integration tests to use this JSON interface. As other issues of the same type might occur in the future when we keep parsing the CLI output.

slyon:
Thank you! I rebased and addressed your comments. I think this is ready to be merged, once CI is green.

PR #380, opened 2023-07-17
Fixes for minor issues

daniloegea:
## Description

Some of this issues were found when I tried to compile netplan as c++, some were found by increasing the warning level with `-Wextra` and some were found by me while fixing the issues found by the compiler.

Some changes might seem unnecessary, specially the one annotating unused parameters with `__unused`. But I think they add some value.

My motivation to increase the warning_level to 2 is to avoid problems like this:

```c
int main() {
    int a = -1;
    unsigned int b = 1;
    if(a > b) {
        printf("According to my calculations %d IS greater than %d\n", a, b);
    }
}
```
```
$ cc main.c
$ ./a.out
According to my calculations -1 IS greater than 1
``` 

It's unfortunate that it's only caught by the compiler by using `-Wextra` or explicitly using `-Wsign-compare` (which is enabled by `-Wextra`).

Annotating the unused parameters with `__unused` at least makes us aware about the big number of function parameters we never use and, because it's annoying having this annotations, we might decide to get rid of them in the future.

Some changes were made to improve consistency, for example, the g++ compiler found a couple of instances of the code below (which is accepted by a C compiler):

```c
enum Values {
    V1, V2, V3,
};
int main() {
    enum Values v = V1 | V2 | V3;
}
```
As the result of the expression is not a variant of the Values enum, it seems to me it doesn't make sense to store it in a variable of type enum Values.

## Checklist

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project('netplan', 'c',
         license: 'GPL3',
         default_options: [
             'c_std=c99',
-            'warning_level=1',
+            'warning_level=2',
             'werror=true',
         ],
         meson_version: '>= 0.61.0',
--- a/src/abi.h
+++ b/src/abi.h
@@ -252,7 +252,7 @@ struct netplan_net_definition {
     } match;
     gboolean has_match;
     gboolean wake_on_lan;
-    NetplanWifiWowlanFlag wowlan;
+    gint wowlan;
     gboolean emit_lldp;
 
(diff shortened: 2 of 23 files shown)


slyon (approved):
Thank you, those are some very good catches.
It is tedious work to fix all those paper cuts, kudos for doing it and increasing the compiler warning level, to avoid such issues going forward!

lgtm.

PR #381, opened 2023-07-18
tests:integration: Adopt for systemd v254 (Closes: #1041310)

slyon:
## Description
systemd-networkd v245 seems to be more picky about moving an interface from "configuring" to "configured" state. It waits for an IPv6 RA to be received. So we need to define "accept-ra: false" in Netplan tests to ignore that, or make our test environment (dnsmasq) to actually provide RAs. I went with the latter.

We still see weird error messages from 'networkctl', which seems new, but they do not fail tests. E.g.: Failed to query link bit rates: Invalid argument
Failed to query link DHCP leases: Invalid argument

I have the impression that those error messages are thrown while the interface is still in "configuring" state, i.e. while waiting for the RA to be received/processed. We're calling 'networkctl status IFACE' in a loop and the error message vanishes when calling the command after a few seconds.

https://bugs.debian.org/1041310

## Checklist

### DIFF
--- a/tests/integration/ethernets.py
+++ b/tests/integration/ethernets.py
@@ -346,7 +346,7 @@ class TestNetworkd(IntegrationTestsBase, _CommonTests):
 
     # TODO: implement link-local handling in NetworkManager backend and move this test into CommonTests()
     def test_link_local_ipv6(self):
-        self.setup_eth(None)
+        self.setup_eth('ra-only')
         with open(self.config, 'w') as f:
             f.write('''network:
   renderer: %(r)s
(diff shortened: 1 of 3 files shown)


slyon:
cc @bluca

bluca:
> We still see weird error messages from 'networkctl', which seems new, but they do not fail tests. E.g.: Failed to query link bit rates: Invalid argument
> Failed to query link DHCP leases: Invalid argument

Those are due to a bug that is going to be fixed by: https://github.com/systemd/systemd/pull/28447

daniloegea (approved):
lgtm.

PR #384, opened 2023-07-20
parse: Downgrade NM passthrough warning to debug

slyon:
## Description
With the NetworkManager integration in place there will be many more cases where "nm-devices" are used, and we don't want to spam system logs with our warnings.

## Checklist


slyon:
@daniloegea WDYT?

PR #385, opened 2023-07-26
Add Python bindings for libnetplan

slyon:
## Description
Implementation of public Python bindings for `libnetplan.so`, using [CFFI](https://cffi.readthedocs.io/) in API mode, as described in spec FO0114. The diff is fairly big, but it contains lots of refactorings, effectively moving 700 LOC of our old/internal ctypes-bindings over into a public module and porting them to make use of CFFI. Going through the commits one by one should make the review manageable:

8184c2b5 parse-nm: improve some error passing
59ee33cf CI: Add python3-cffi build-dependency
032e3568 include: Extend public API functions
9a84892a src: make internal API functions/structures available
4bfe29be bindings: python-cffi build skeleton
fdad4e08 bindings: Initial implementation of CFFI Python bindings
40c4f47f cli:tools:test: Make use of new cffi python bindings
a21d9f43 cli: drop legacy/internal ctypes bindings, now unused
8fd5adf2d1541adb8bd7a30fa3fea72e78ed4bf9 examples: Add an example of how to use Netplan's CFFI bindings

The most interesting bits are porbably those commits, where the new module is introduced, to be installed into `usr/lib/python3/dist-packages/netplan/`:
* https://github.com/canonical/netplan/commit/4bfe29be6b00f6ecf7b6dec6dc6351122a2a83ec bindings: python-cffi build skeleton
* https://github.com/canonical/netplan/commit/fdad4e08956bb0e29edf4d5ade9338d4d2c91312 bindings: Initial implementation of CFFI Python bindings

We're using the file in `python-cffi/netplan/_build_cffi.py` to build a `_netplan_cffi.cpython-311-x86_64-gnu.so` (internal) binary module, that is linked to `libnetplan.so` itself. In the `ffibuilder.cdef(...)` we declare any functions/constants/structures that we want to make available from libnetplan.so to the Python bindings. The compiler will match those against Netplan's headers at build time. In `python-cffi/netplan/{parser,state,netdef}.py` we're then making use of those declarations to define our high-level, Python-native bindings.

We're still making some limited use of ctypes in our tests (`test_libnetplan.py`) in order to validate some raw library functionality, which are not necessarily exposed through the Python bindings.

## Checklist

- [x] \(Optional\) Closes an open bug ~~in Launchpad~~: FR-2780


python-cffi/netplan/state.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @property
+    def backend(self) -> str:
+        return ffi.string(lib.netplan_backend_name(lib.netplan_state_get_backend(self._ptr))).decode('utf-8')
+
+    @property
+    def netdefs(self) -> NetDefinitionIterator:
+        return dict((nd.id, nd) for nd in NetDefinitionIterator(self, None))
+
+    # TODO: veths
```
daniloegea, line 89: suggestion: `virtual-ethernets` and `dummy-devices` could be added now.

examples/cffi-bindings.py
```diff
@@ -0,0 +1,58 @@
...
+# the Free Software Foundation; version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
daniloegea, line 17: suggestion: it would be nice to have a short comment explaining what this script does.

python-cffi/netplan/netdef.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        next_value = lib._netplan_netdef_address_iter_next(self.iterator)
+        if not next_value:
+            raise StopIteration
+        content = next_value
+        # XXX: Introduce getters for .address/.lifetime/.label, to avoid
```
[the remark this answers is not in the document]
slyon: Yes.. we should build proper API around it, eventually. And keep it internal until that's done.

slyon:
Thanks @daniloegea! I've rebased and addressed most of your remarks. I still need to improve my comments/description in  `examples/cffi-bindings.py` and `python-cffi/netplan/meson.build` but maybe you could already double-check the logical changes I did so far.

gcovr.cfg
```diff
@@ -0,0 +1 @@
+exclude = _build-cov/python-cffi/netplan/_*
```
daniloegea, line 1: I just realized that the test coverage-c is failing for me. The reason is because my build directory is called `build`, and not `_build-cov`... we will probably need to invert the logic here to not depend on the build directory name.

I tested this and it worked:

```
filter = src/*
filter = tests/ctests/*
```

daniloegea (approved):
Looks good!

Only one thing I'd like to be changed before merging: the new `gcovr.cfg` file probably should use the `filter` configuration to include the source code instead of `exclude` with a hardcoded build directory. The test coverage-c will fail if you use a different build directory.

python-cffi/netplan/meson.build
```diff
@@ -0,0 +1,55 @@
...
+        join_paths(meson.project_source_root(), 'include'),
+        join_paths(meson.current_build_dir(), 'src'),
+    ],
+    output: '_netplan_cffi.c',
+)
+
+# See https://github.com/grimme-lab/xtb-python/blob/main/meson.build
+cffi_pyext = python.extension_module(
+    '_netplan_cffi',
+    link_whole: static_library(
```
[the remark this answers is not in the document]
slyon: After further investigation, it looks like we don't need that quirk at all, but can just build the binary Python module (shared object) directly from the source (see https://github.com/mesonbuild/meson/issues/5797).

slyon:
Thanks, I've addressed all of your comments. CI failures about Azure kernels are unrelated:
`E: Package 'linux-modules-extra-5.15.0-1042-azure' has no installation candidate`

Let's merge this!

PR #388, opened 2023-08-01
cli:status: Make rich pretty printing optional

slyon:
## Description
cli:status: Make rich pretty printing optional

To avoid a hard dependency on python3-rich

E.g. needed for stable series SRUs (like Jammy), which don't support python3-rich in `main`

## Checklist

### DIFF
--- a/netplan_cli/cli/commands/status.py
+++ b/netplan_cli/cli/commands/status.py
@@ -18,22 +18,30 @@
 '''netplan status command line'''
 
 import json
+import logging
+import re
 
 import yaml
-from rich.console import Console
-from rich.highlighter import RegexHighlighter
-from rich.theme import Theme
 
 from .. import utils
 from ..state import SystemConfigState, JSON
 
 
-class NetplanHighlighter(RegexHighlighter):
-    base_style = 'netplan.'
-    highlights = [
-        r'(^|[\s\/])(?P<int>\d+)([\s:]?\s|$)',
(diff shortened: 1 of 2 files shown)


netplan_cli/cli/commands/status.py
```diff
@@ -57,20 +65,31 @@ def run(self):
         self.parse_args()
         self.run_command()
 
+    def plain_print(self, *args, **kwargs):
+        if len(args):
+            pattern = r'\[\/?\w+\]'
```
daniloegea, line 70: thought: I think it's ok in general. But, of course, interfaces with names like `int[external]0` are possible in Linux so it will end up removing part of the name if someone for some reason uses `[]` in interface names out there... but apparently it is possible without this patch as well, as rich will interpret `[]` as formatting.

even funnier, one can affect the rendering by creating an interface called, for example, `[red]`...

daniloegea (approved):
Just a note about interface names, but lgtm.

netplan_cli/cli/commands/status.py
slyon: @daniloegea Thanks, I improved the matching pattern a bit, to match only if there's also a corresponding closing tag. This might have other edge cases (e.g. nested tags...), but maybe those are less common or more under our control. WDYT?

daniloegea: nice, it looks like it solves the problem for the plain_print() case.

PR #392, opened 2023-08-10
github: only run the coverity workflow on our repository

daniloegea:
Also, get the project version dynamically using meson introspect.

## Checklist

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   coverity:
+    if: github.repository == 'canonical/netplan'
     runs-on: ubuntu-22.04
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/coverity.yml
```diff
@@ -29,4 +30,4 @@ jobs:
           tar czf netplan.tar.gz cov-int
       - name: Upload results
         run: |
-          curl --form token=${{ secrets.COVERITY_TOKEN }} --form email=${{ secrets.COVERITY_EMAIL }} --form file=@netplan.tar.gz --form version="0.106" --form description="Coverity scan" https://scan.coverity.com/builds?project=Netplan
+          curl --form token=${{ secrets.COVERITY_TOKEN }} --form email=${{ secrets.COVERITY_EMAIL }} --form file=@netplan.tar.gz --form version="$(meson introspect coveritybuild --projectinfo | jq -r .version)" --form description="Coverity scan" https://scan.coverity.com/builds?project=Netplan
```
slyon, line 33: thought (non-blocking): I wonder if we can get a more specific version for those coverity scans, using the git revision?

Something along the lines of this?
```bash
TAG=$(git describe --tags $(git rev-list --tags --max-count=1))  # find latest (stable) tag
REV=$(git rev-parse --short HEAD)  # get current git revision
VER="$TAG+git~$REV"
```

slyon (approved):
LGTM!
One little inline thought for consideration.

.github/workflows/coverity.yml
daniloegea: sounds like a good idea

slyon:
> `E: Package 'linux-modules-extra-5.15.0-1042-azure' has no installation candidate`
> 
> Apparently some old kernel versions were purged again...

Yeah.. I think the CI failures are related to https://ubuntu-archive-team.ubuntu.com/nbs.html cleaning. Should resolve itself eventually, once the CI runners are upgraded to a newer kernel.

PR #393, opened 2023-08-14
Addressing a few issues found by coverity

daniloegea:
## Description

Addressing CIDs 398341, 398340 and 398338.

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -2097,6 +2097,7 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, __unused const void* _, GEr
 
 err:
     route_clear(&npp->current.route);
+    npp->current.route = NULL;
     return FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


src/parse.c
```diff
@@ -2097,6 +2097,7 @@ handle_routes(NetplanParser* npp, yaml_node_t* node, __unused const void* _, GEr
 
 err:
     route_clear(&npp->current.route);
+    npp->current.route = NULL;
```
slyon, line 2100: comment: FWIW, I think this nullification is already being done, using the `CLEAR_FROM_FREE` macro. But I guess it doesn't hurt to do it manually again in the error case.

daniloegea: hmm yeah, that's true. Coverity (and I) wasn't smart enough to see that I guess... well, if nothing, this line will eliminate an issue from the Coverity report.

slyon (approved):
+1 CI failures seem unrelated.

PR #397, opened 2023-08-16
libnetplan: expose the routes list in the netdef

daniloegea:
## Description

I created the NetplanRoute class as a dataclass but I'm not convinced it's a good idea... I feel like it's better to assign each property separately without an `__init__` method...

## Checklist

### DIFF
--- a/tests/test_libnetplan.py
+++ b/tests/test_libnetplan.py
@@ -293,6 +293,50 @@ class TestNetdefNameserverSearchDomainIterator(TestBase):
         self.assertSetEqual(expected_domains, set(domain for domain in netdef.nameserver_search))
 
 
+class TestNetdefRouteIterator(TestBase):
+    def test_with_empty_routes(self):
+        state = state_from_yaml(self.confdir, '''network:
+  ethernets:
+    eth0: {}''')
+
+        netdef = next(netplan.netdef.NetDefinitionIterator(state, "ethernets"))
+        self.assertTrue(len([ip for ip in netdef.routes]) == 0)
+
+    def test_iter_routes(self):
+        state = state_from_yaml(self.confdir, '''network:
+  ethernets:
+    eth0:
+      routes:
+        - to: default
+          via: 192.168.0.1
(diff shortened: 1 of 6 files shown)


tests/test_libnetplan.py
```diff
@@ -293,6 +293,50 @@ def test_iter_ethernets_nameservers_and_domains(self):
...
+          on-link: true
+          type: local
+          scope: host
+          mtu: 1500
+          congestion-window: 123
+          advertised-receive-window: 321
+          from: 192.168.0.0/24''')
+
+        netdef = next(netplan.netdef.NetDefinitionIterator(state, "ethernets"))
+        routes = [route for route in netdef.routes]
```
slyon, line 330:
```suggestion
        routes = list(netdef.routes)
```

suggestion (non-blocking): I guess this way it would be easier to read and according to https://www.curs-ml.com/post/fastest-way-to-convert-an-iterator-to-a-list it would even be faster. But OTOH this is a test case, so we can keep it as-is if you want to. The same applies to `ip in netdef.routes` (and others) above.

slyon (approved):
Thanks, lgtm. I left one non-blocking inline comment.

What's your reasoning against the `@dataclass`, I think it's looking good.

tests/test_libnetplan.py
daniloegea: oh, yeah I think at some point I was calling str() on route and ended up leaving the list comprehension there.

daniloegea:
So, I think the only thing I didn't like is passing all those objects of the same type to the constructor. One can easily get confused and put them in the wrong place. It's harder to do that using the pattern `route.to = to, route.via = via`...

But I think the dataclass will be beneficial in other ways, such as comparing the objects for example... let's see I guess

slyon:
> But I think the dataclass will be beneficial in other ways, such as comparing the objects for example... let's see I guess

We could consider calling it `class _NetplanRoute` for now, so it's easier for us to change the structure in the future. We're our only consumer of that class currently anyway.

PR #398, opened 2023-08-16
parse-nm: add support for VRF devices

daniloegea:
## Checklist

### DIFF
--- a/src/parse-nm.c
+++ b/src/parse-nm.c
@@ -53,6 +53,8 @@ type_from_str(const char* type_str)
         return NETPLAN_DEF_TYPE_VETH;
     else if (!g_strcmp0(type_str, "vlan"))
         return NETPLAN_DEF_TYPE_VLAN;
+    else if (!g_strcmp0(type_str, "vrf"))
+        return NETPLAN_DEF_TYPE_VRF;
     else if (   !g_strcmp0(type_str, "wireguard")
              || !g_strcmp0(type_str, "vxlan")
              || !g_strcmp0(type_str, "ip-tunnel"))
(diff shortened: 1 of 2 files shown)


src/parse-nm.c
```diff
@@ -697,6 +699,14 @@ netplan_parser_load_keyfile(NetplanParser* npp, const char* filename, GError** e
         }
     }
 
+    /* Handle VRFs */
+    if (nd_type == NETPLAN_DEF_TYPE_VRF) {
+        if (g_key_file_has_key(kf, "vrf", "table", NULL)) {
+            nd->vrf_table = g_key_file_get_integer(kf, "vrf", "table", NULL);
```
slyon, line 705: todo: According to https://developer-old.gnome.org/NetworkManager/stable/settings-vrf.html the keyfile's `vrf.table` field is of type `uint32`, so we should probably use something like this, to avoid truncation:

```suggestion
            nd->vrf_table = g_key_file_get_uint64(kf, "vrf", "table", NULL);
```

src/parse-nm.c
```diff
@@ -697,6 +699,14 @@ netplan_parser_load_keyfile(NetplanParser* npp, const char* filename, GError** e
...
+        }
+        _kf_clear_key(kf, "vrf", "table");
```
slyon, line 707: nitpick (non-blocking): This can probably move into the `if (g_key_file_has_key(kf, "vrf", "table", NULL))` case above, as we don't need to clear a non-existing field.

slyon (approved):
I have one inline TODO, which I'd like to see addressed. Other than that, lgtm.

PR #399, opened 2023-08-17
Netplan state refactoring

daniloegea:
## Description

It's basically #386 without the diff part.

## Checklist


netplan_cli/cli/state.py
```diff
@@ -193,26 +224,9 @@ def down(self) -> bool:
...
-            'wwan': 'modem',
-            'vlan': 'vlan',
-            'vrf': 'vrf',
-            'vxlan': 'tunnel',
-            })
         nd_type = self.nd.get('Type') if self.nd else None
-        if nd_type in match:
-            return match[nd_type]
+        if nd_type := DEVICE_TYPES.get(nd_type, nd_type):
+            return nd_type
```
slyon, line 229: I think this doesn't fully match up (I guess I was confused earlier). We're now returning the old `nd_type` value, but we actually want to return `None` if it's not found, but still log the old `nd_type` value in the warning.

Maybe we should just keep the original logic here, without the `:=` shortcut?

slyon (approved):
Just one tiny TODO inline, that we need to resolve. Otherwise, lgtm!
Thank you very much!

PR #400, opened 2023-08-21
Post 0.107 cleanup & dropping API/ABI compat

slyon:
### Note
Please review commit-by-commit (and ignore the huge "ABI: regenerate compat check for dropped symbols" commit in `abi-compat/*.xml` -- it's autogenerated)

### Unrelated changes
This PR contains some unrelated "cleanup" changes for post 0.107 tasks.

### Primary changes
The primary changes, of dropping legacy Netplan API and ABI (in preparation for a breaking 1.0 release) start from commit "test:generator:base: Refactor to use new API".

## Checklist


slyon:
> * Maybe comment out these asserts to unblock autopackage tests https://github.com/canonical/netplan/blob/main/tests/integration/routing.py#L316 until we properly address the problem.

see https://github.com/canonical/netplan/pull/401

tests/generator/base.py
```diff
@@ -269,31 +272,33 @@ def validate_generated_yaml(self, yaml_input):
...
         for input in yaml_input:
-            lib.netplan_clear_netdefs()  # clear previous netdefs
-            lib.netplan_parse_yaml(input.encode(), None)
-            lib.write_netplan_conf_full(output.encode(), self.workdir.name.encode())
+            parser = netplan.Parser()
+            parser.load_yaml(input)
+            state = netplan.State()
+            state.import_parser_results(parser)
+
+            # TODO: Allow handling of full hierarchy overrides,
```
daniloegea, line 282: I think this is fine for now, but we should create a task to keep track of it as it will not be possible to write tests that rely on merging without skipping this validation.

tests/ctests/test_netplan_misc.c
```diff
@@ -177,6 +189,107 @@ test_netplan_netdef_get_output_filename_invalid_backend(__unused void** state)
...
+
+    /* Check file exists */
+    struct stat st = {0};
+    char output_yaml[80] = {};
+    snprintf(output_yaml, 80, "%s/90-NM-990548be-01ed-42d7-9f9f-cd4966b25c08.yaml", etc_netplan);
+    assert_true(stat(output_yaml, &st) == 0);
+
+    /* Check file contents */
+    FILE *fd = fopen(output_yaml, "r");
+    char file_buffer[600];
```
daniloegea, line 280: The unittest is failing with ASAN because this buffer is not being zeroed (on the previous test as well but it's not failing...). There is garbage in the end of string (see below). Initializing it with zeros (` = {0}`) fixes the test. The memory leak happens because the `assert_string_equal` assertion prevents `np_state` to be cleared.

```
[  ERROR   ] --- "network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
      networkmanager:
        uuid: "990548be-01ed-42d7-9f9f-cd4966b25c08"" != "network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
      networkmanager:
        uuid: "990548be-01ed-42d7-9f9f-cd4966b25c08"IH"
[   LINE   ] --- ../tests/ctests/test_netplan_misc.c:282: error: Failure!
```

daniloegea: The RPM build is failing for the same reason (see the end of the buffer below). But there both tests are failing without ASAN, which is interesting.

```
[  ERROR   ] --- "network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
      networkmanager:
        uuid: "990548be-01ed-42d7-9f9f-cd4966b25c08"" != "network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
      networkmanager:
        uuid: "990548be-01ed-42d7-9f9f-cd4966b25c08"oÏwBV"
[   LINE   ] --- ../tests/ctests/test_netplan_misc.c:282: error: Failure!
```

src/util-internal.h
```diff
@@ -34,7 +34,7 @@ wifi_frequency_24;
 extern GHashTable*
 wifi_frequency_5;
 
-NETPLAN_ABI void
+NETPLAN_ABI void // XXX: still used in generate.c
```
daniloegea, line 37: Do we need to keep the `NETPLAN_ABI` macro? We could replace it with `NETPLAN_INTERNAL`...

daniloegea (changes requested):
It's looking good so far. The only thing we must fix are the tests that are failing. They just need to be zeroed because the `\0` is missing when the strings are compared (see my comment in the test).

src/util-internal.h
slyon: Very true!

tests/generator/base.py
slyon: :ballot_box_with_check:  FR-6338

slyon:
Thanks for your review @daniloegea ! I've addressed all of your comments and got the CI to be green (except for the known tunnel/GRE failures on Jammy autopkgtest, which are unrelated).

PTAL.

.github/workflows/autopkgtest.yml
```diff
@@ -45,7 +45,12 @@ jobs:
           pull-lp-source netplan.io
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
-          sed -i 's|rm debian/tmp/lib/netplan/generate|# DELETED|' debian/rules
+          # usrmerge-fix
+          mkdir debian/extra
```
daniloegea, line 49: I believe this will start to fail when your changes get synced with Ubuntu but we can fix the CI again later.

daniloegea (approved):
lgtm!

.github/workflows/autopkgtest.yml
slyon: Yes. And I just uploaded 0.107.1-2 into Debian unstable. I'll use my next PR to adopt CI accordingly, once that landed.

PR #404, opened 2023-08-25
wifi: support WPA2 and WPA3 Personal simultaneously

daniloegea:
When using WPA-PSK and defining the password without the "auth" section, enable support for both WPA2 and 3 Personal. This will set the support for Protected Management Frames to optional implicitly.

## Checklist


slyon (approved):
Thank you! LGTM.

This should keep simple things simple. Allowing the casual user to "just connect" to any WPA2/3 personal network, providing only their password, as discussed in https://github.com/canonical/netplan/pull/402

PR #407, opened 2023-09-01
tunnels/validation: do not error out if "local" is not defined

daniloegea:
The "local" IP property is not required by backends so we shouldn't error out if it's not defined.

Tracking LP bug https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2034067

## Checklist


slyon (approved):
This is relaxing the validation a bit, in accordance with the backend's rules. LGTM.

Could you please create a bug report about it on Launchpad, so we can get it landed in Ubuntu post feature freeze?

PR #409, opened 2023-09-19
Extend KeepConfiguration= functionality

chr4:
## Description

With systemd-v243 (which is the default since focal), the "CriticalConnection" attribute was
revamped into "KeepConfiguration" [1], addressing several issues, especially in high-availability
environents [2][3].

The new "KeepConfiguration" setting can take multiple arguments, according to the documentation
[4]:

> Takes a boolean or one of "static", "dhcp-on-stop", "dhcp". When "static", systemd-networkd will
> not drop static addresses and routes on starting up process. When set to "dhcp-on-stop",
> systemd-networkd will not drop addresses and routes on stopping the daemon. When "dhcp", the
> addresses and routes provided by a DHCP server will never be dropped even if the DHCP lease
> expires. This is contrary to the DHCP specification, but may be the best choice if, e.g., the
> root filesystem relies on this connection. The setting "dhcp" implies "dhcp-on-stop", and "yes"
> implies "dhcp" and "static". Defaults to "no".

If I understand the code correctly, it still uses "CriticalConnection" when setting "critical: true" in the netplan configuration.

- [1] https://github.com/systemd/systemd/pull/12511/files
- [2] https://github.com/systemd/systemd/issues/12050
- [3] https://chr4.org/blog/2019/01/21/make-keepalived-play-nicely-with-netplan-slash-systemd-network/
- [4] https://systemd.network/systemd.network.html#KeepConfiguration=

Closes: https://bugs.launchpad.net/netplan/+bug/1896799

## Checklist


chr4:
This is a first draft - not tested yet. Feedback welcome!

daniloegea:
Hello @chr4,

Thank you for bringing this to our attention and also for your PR, it's much appreciated.

Yes, we are still emitting the following configuration when `critical: true` is used in Netplan:
```
[DHCP]
CriticalConnection=true
```
As you already said,  [According to systemd's NEWS file](https://github.com/systemd/systemd/blob/main/NEWS#L6751C36-L6751C36) this option was deprecated some time ago:

```
The CriticalConnection= setting in .network files is now deprecated,
and replaced by a new KeepConfiguration= setting which allows more
detailed configuration of the IP configuration to keep in place.
```

So yes, we need to get rid of `CriticalConnection` in favor of `KeepConfiguration`.

The problem is that we can't get rid of the `critical` key in Netplan, because it would break Netplan for everybody using this option.

Instead of replacing it with `keep-configuration`, I'd suggest a much simpler approach: adapt `critical` to accept all the new values supported by `KeepConfiguration`, including `true` and `false`.

This would make the patch simpler and wouldn't break Netplan. Also, all those functions and variables renaming wouldn't be necessary.

The second approach would be to keep `critical` as it is, deprecate it and introduce a new key, such as `keep-configuration`.

Keeping `critical` untouched would still work because systemd still accepts `CriticalConnection`, even though it's not even documented anymore. See https://github.com/systemd/systemd/blob/main/src/network/networkd-network.c#L245

In my opinion we could reuse `critical` and just change it to accept all the new values accepted by `KeepConfiguration`.

@slyon @vorlonofportland any thoughts on this?

src/abi.h
```diff
@@ -211,7 +211,7 @@ struct netplan_net_definition {
     /* status options */
     gboolean optional;
     NetplanOptionalAddressFlag optional_addresses;
-    gboolean critical;
+    char* keep_configuration;
```
daniloegea, line 214: To make things simpler, we should use an `enum` for this option. It has a well defined set of possible values, so it would naturally fit well in an `enum`:

```c
enum {
    NETPLAN_CRITICAL_TRUE,
    NETPLAN_CRITICAL_FALSE,
    NETPLAN_CRITICAL_STATIC,
    ...,
} CriticalOptions;
```

Please, check my comment in the summary.

chr4: Good idea! Let me revamp

chr4: I've been trying to rework this, but I'm not sure if I'm on the right track, as quite a few tests are failing or even segfaulting.

src/types.c
```diff
@@ -225,7 +225,7 @@ reset_netdef(NetplanNetDefinition* netdef, NetplanDefType new_type, NetplanBacke
 
     netdef->optional = FALSE;
     netdef->optional_addresses = 0;
-    FREE_AND_NULLIFY(netdef->keep_configuration);
+    netdef->critical = NETPLAN_CRITICAL_FALSE;
```
daniloegea, line 228: We probably shouldn't initialize `critical` with false.

According to the documentation, KeepConfiguration `defaults to "dhcp-on-stop" when systemd-networkd is running in initrd, "yes" when the root filesystem is a network filesystem, and "no" otherwise.`

Initializing it to false would imply setting it to false even when the default is something else. I suggest adding a new entry in the enum, something like `NETPLAN_CRITICAL_UNSET`, and use it as the default value. Then in the switch-case you added, you'd also have an option for the false case and and an option that does nothing for the unset case.

src/parse.c
```diff
@@ -2785,7 +2785,7 @@ static const mapping_entry_handler dhcp6_overrides_handlers[] = {
     {"accept-ra", YAML_SCALAR_NODE, {.generic=handle_accept_ra}, netdef_offset(accept_ra)}, \
     {"activation-mode", YAML_SCALAR_NODE, {.generic=handle_activation_mode}, netdef_offset(activation_mode)}, \
     {"addresses", YAML_SEQUENCE_NODE, {.generic=handle_addresses}, NULL}, \
-    {"keep-configuration", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(keep_configuration)}, \
+    {"critical", YAML_SCALAR_NODE, {.generic=handle_netdef_bool}, netdef_offset(critical)}, \
```
daniloegea: As the new critical option is not a boolean anymore, we'll need to update the parser. You'll need to create a new function that checks what option the user set and store one of the enum variants in the variable. You can base your implementation on the `handle_auth_methods` function for example https://github.com/chr4/netplan/blob/feat/keep-configuration/src/parse.c#L947

And as we want to still accept the boolean values so we'll not break existing configurations out there, we'll need to accept all the boolean values accepted by netplan. See https://github.com/chr4/netplan/blob/feat/keep-configuration/src/parse.c#L402

src/networkd.c
```diff
@@ -916,13 +916,26 @@ netplan_netdef_write_network_file(
...
+        case NETPLAN_CRITICAL_STATIC:
+            g_string_append(s, "KeepConnection=static\n");
+            break;
+        case NETPLAN_CRITICAL_DHCP:
+            g_string_append(s, "KeepConnection=dhcp\n");
+            break;
+        case NETPLAN_CRITICAL_DHCP_ON_STOP:
+            g_string_append(s, "KeepConnection=dhcp-on-stop\n");
+            break;
+        default: g_assert_not_reached(); // LCOV_EXCL_LINE
```
daniloegea, line 937: One of the reasons it's failing tests is because the case for the false option is missing here. See my comment below about the enum.

python-cffi/netplan/netdef.py
```diff
@@ -75,8 +75,8 @@ def set_name(self) -> str:
         return _string_realloc_call_no_error(lambda b: lib.netplan_netdef_get_set_name(self._ptr, b, len(b)))
 
     @property
-    def keep_configuration(self) -> str:
-        return _string_realloc_call_no_error(lambda b: lib.netplan_netdef_get_keep_configuration(self._ptr, b, len(b)))
+    def critical(self) -> bool:
```
daniloegea, line 78: I wonder if we could keep the Python binding untouched and return False if critical is false (or unset) and True if it's set to any of the other values...

Although we probably should change it to return the enum values. You can see an example of enums in the Python code here https://github.com/chr4/netplan/blob/feat/keep-configuration/python-cffi/netplan/_utils.py#L27

daniloegea (changes requested):
Hi there! Sorry for taking so long to reply. I added a few comments in your most recent commits. I hope it helps.

May I ask you to squash your commits for the next iteration? It will make it a bit easier to read the changes.

Thank you! :)

src/types.c
slyon: +1 for using `NETPLAN_CRITICAL_UNSET` here.

doc/netplan-yaml.md
```diff
@@ -374,11 +374,19 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
   > (networkd backend only) Allow the specified interface to be configured even
   > if it has no carrier.
 
-- **critical** (bool)
+- **critical** (scalar)
```
slyon, line 377: We'd need to keep this as "(bool or scalar)", making use of `handle_generic_bool()` internally in `parse.c`, checking for the error condition in the `GError** error` parameter. Then falling back to "scalar" (and validating the input to be one of the accepted strings) if a bool cannot be parsed.

src/abi.h
```diff
@@ -32,6 +32,14 @@ typedef enum {
     NETPLAN_OPTIONAL_STATIC  = 1<<4,
 } NetplanOptionalAddressFlag;
 
+typedef enum {
+    NETPLAN_CRITICAL_FALSE,
+    NETPLAN_CRITICAL_TRUE,
+    NETPLAN_CRITICAL_STATIC,
+    NETPLAN_CRITICAL_DHCP,
+    NETPLAN_CRITICAL_DHCP_ON_STOP,
+} NetplanCriticalOption;
```
slyon, line 41: nitpick: we might just want to call that `NetplanCritical` to stay more in line with the other enums. Everything is an "Option", so that suffix feels kind of redundant.

PS: please also add and use a `NETPLAN_CRITICAL_UNSET` option, as suggested by @daniloegea 

Note: changing the `gboolean` type to an enum should not break our ABI, as both will be mapped to a `gint` in the end.

src/networkd.c
slyon: +1

src/util-internal.h
```diff
@@ -102,7 +102,7 @@ _netplan_state_get_vf_count_for_def(const NetplanState* np_state, const NetplanN
 NETPLAN_INTERNAL gboolean
 _netplan_netdef_get_sriov_vlan_filter(const NetplanNetDefinition* netdef);
 
-NETPLAN_INTERNAL gboolean
+NETPLAN_INTERNAL NetplanCriticalOption
 _netplan_netdef_get_critical(const NetplanNetDefinition* netdef);
```
slyon, line 106: Note: This is `NETPLAN_INTERNAL` API, as also desribed by the `_netplan` prefix. It's supposed to be used by Netplan tools/CLI only, not by external consumers. Although, changing the signature is an API breaking change, I think it's fine as we're the only ones affected.

doc/netplan-yaml.md
```diff
@@ -374,11 +374,19 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
...
+  > This option sets the systemd KeepConfiguration option.
+  > (not recognized by NetworkManager)
+  >
+  > When set to "static", static addresses and routes won't be dropped on
+  > starting up process.
+  > When set to "dhcp-on-stop", addresses and routes won't be dropped when
+  > stopping the daemon.
+  > When set to "dhcp", addresses and routes provided by a DHCP server will
+  > never be dropped even if the DHCP lease expires, implies "dhcp-on-stop".
+  > When set to "yes", "dhcp" and "static" is implied.
```
slyon, line 388: question: Did you double-check if something similar to `KeepConfiguration=` exists for NetworkManager these days? I'd like to align the wording between the two backends. If it does not exist it might be fine re-using the systemd-networkd terminology here. I need to think about this a bit more.

src/parse.c
slyon: It looks like any changes to `src/parse.c` have been lost in a rebase?

src/networkd.c
slyon: +1

python-cffi/netplan/netdef.py
slyon: That's an interesting thought. I feel like we should not diverge the bindings from the native API behavior, though. But we need to think about how we can handle that in a backwards compatible way.

slyon (changes requested):
Thanks a lot for your contribution to Netplan @chr4! And for the initial review from @daniloegea 

Overall, I think this is going into the right direction, but could you please have a look at #424 ? Maybe we should make it a two-step approach to land that first (and fix the bug at hand), then work on the extension of the new setting as an additional step.

Danilo's suggestions are good and should be addressed. I like the approach of keeping the `critical:` setting in Netplan and just extending it to accept "bool or scalar" eventually. So we can keep backwards compatibility and support the new usecase going forward.

I left some additional remarks inline, which need some more work. Especially:
* Any changes to `src/parse.c` seem to been lost. We need those.
* unit-test and CI failures. We want it to be (mostly) green. Danilo gave some good advice of how to approach this. You might also want to run `make check-coverage` locally, to test on your own, outside the CI.

zakthan:
Hi all,
we are using netplan with calico for our k8s clusters.
We would love to have this option so we could use `netplan generate && networkctl reload`

`KeepConfiguration=static` is critical for environments running BIRD/BGP alongside systemd-networkd. When `networkctl reload` is triggered:

`KeepConfiguration=true` causes networkd to attempt reconciliation of foreign (BIRD) routes, triggering a full reconfigure cycle that deletes and re-adds all routes (disruption)
`KeepConfiguration=static` precisely preserves only `proto static` routes and ignores foreign routes, enabling clean delta-only reloads with zero route disruption


Not sure if this PR can still be merged so I have created new PR https://github.com/canonical/netplan/pull/599

JegadishD:
Fixed as part of commit https://github.com/canonical/netplan/pull/599

PR #411, opened 2023-09-27
added mii-monitor-interval

Sudo-Whodo:
…needed

## Description
I added mii-monitor-interval to the example so the active-backup connection would failover. The default value is 0 and doesn't monitor the carrier connection. 

## Checklist

### DIFF
--- a/examples/bonding.yaml
+++ b/examples/bonding.yaml
@@ -13,3 +13,4 @@ network:
       parameters:
         mode: active-backup
         primary: enp3s0
+        mii-monitor-interval: 1


daniloegea:
Hi there, thank you for your PR.

The default time unit, when not specified, is milliseconds. I wonder if 1 millisecond is a too small value... Was it your intention setting it to 1 second instead? If so, it must be defined as `1s`.

Sudo-Whodo:
Hey, My main goal with the pr is to enable the mii monitoring for the active-backup example. I used this as a template and went to test the failover and it didn't happen. Once I added the mii-monitor-interval it failed-over as intended. I've updated the value to 100ms. This is from the Mikrotik default value I saw online in their [docs](https://help.mikrotik.com/docs/display/ROS/Bonding), but the desire is to add a positive int value.

daniloegea:
Yes, I got it :)
But the thing with example files is that people will copy and paste them as they are and some settings might cause problems. I'm not really sure if the 1ms mii monitor would cause any issues but I feel like polling something every millisecond might cause unnecessary load.

Anyway, thanks for your patch ;)

daniloegea (approved):

PR #412, opened 2023-10-04
Ignore bad NetDefs and files via parser flags

daniloegea:
## Description

This is an implementation of what I called "parser flags". The idea is to enable the user to change some parsing decisions.

The main application at the moment is to support ignoring parsing errors. Currently Netplan will not generate any configuration if a little mistake is made in one of its YAML files. It can be really bad if you're operating a remote system and end up rebooting the system with a syntax issue in one of your files. Your system will basically boot without any network configuration.

The IGNORE_ERRORS flag will allow the Netplan generator to ignore bad files and bad netdefs so it will still generate some configuration.

**TESTS**

Integration tests will be executed twice now, once without the ignore_errors flags and once with it enabled. It's intended to increase the confidence that the parser behaves as expected when the flag is enabled and the configuration is good.

I prepared a PPA for Ubuntu Noble with this patch: https://launchpad.net/~danilogondolfo/+archive/ubuntu/netplan-parser-flags

Suggestions of tests:

It can be easily tested in a LXD VM. You can add some broken configuration and reboot the VM or call `/usr/libexec/netplan/generate -i` to see the parsing process when it's ignoring errors

1. Break the default `50-cloud-init.yaml` and reboot the VM. For example, the config below should still bring enp5s0 up and with DHCP working.

```yaml
network:
  version: 2
  ethernets:
    enp5s0:
      dhcp4: true
      dhcp: true
```

2. Add a second file that is not a valid YAML
3.  Add a third broken file with some complex configuration and dependencies between netdefs. For example, with the configuration below `eth2` should still have a backend configuration file and it should contain `Bond=bond0`.
```yaml
network:
  ethernets:
    eth0: {}
    eth1: {}
    eth2:
      a: b

  bonds:
    bond0:
      interfaces:
        - eth0
        - eth1
        - eth2
```
4. Add a fourth broken file with missing dependencies. For example, with the config below you still should have a `br0` interface created:
```yaml
network:
  bridges:
    br0:
      addresses:
        - 10.20.30.40/24

      interfaces:
        - aaaa
        - bbbb
        - dddd
```

## Checklist

### DIFF
--- a/include/parse.h
+++ b/include/parse.h
@@ -54,6 +54,26 @@ NETPLAN_PUBLIC void
 netplan_parser_clear(NetplanParser **npp);
 
 /**
+ * @brief   Set @ref NetplanParser flags.
+ * @details Parser flags are used to change the default behavior of the parser.
+ * @param[in] npp       The @ref NetplanParser to set the flags.
+ * @param[in] flags     The value of the flags. The possible values are defined in @ref NETPLAN_PARSER_FLAGS
+ * @param[out] error    Filled with a @ref NetplanError in case of failure
+ * @return              Indication of success or failure
+ */
+NETPLAN_PUBLIC gboolean
+netplan_parser_set_flags(NetplanParser* npp, unsigned int flags, NetplanError** error);
+
+/**
+ * @brief   Get @ref NetplanParser flags.
+ * @details Parser flags are used to change the default behavior of the parser.
+ * @param[in] npp   The @ref NetplanParser to get the flags from.
+ * @return          The current flags set in the parser.
+ */
+NETPLAN_PUBLIC unsigned int
+netplan_parser_get_flags(const NetplanParser* npp);
+
+/**
  * @brief Parse a given YAML file and create or update the list of @ref NetplanNetDefinition inside @p npp.
  * @param[in]  npp      The @ref NetplanParser object that should contain the parsed data
  * @param[in]  filename Full path to a Netplan YAML configuration file
--- a/include/parse.h
+++ b/include/parse.h
@@ -74,6 +74,15 @@ NETPLAN_PUBLIC unsigned int
 netplan_parser_get_flags(const NetplanParser* npp);
 
 /**
+ * @brief   Get @ref NetplanParser error count.
+ * @details The number of errors that were ignored when the IGNORE_ERRORS is used.
+ * @param[in] npp   The @ref NetplanParser to get the error count from.
+ * @return          The current error count.
+ */
+NETPLAN_PUBLIC unsigned int
+netplan_parser_get_error_count(const NetplanParser* npp);
+
+/**
  * @brief Parse a given YAML file and create or update the list of @ref NetplanNetDefinition inside @p npp.
  * @param[in]  npp      The @ref NetplanParser object that should contain the parsed data
  * @param[in]  filename Full path to a Netplan YAML configuration file
--- a/include/types.h
+++ b/include/types.h
@@ -132,7 +132,8 @@ enum NETPLAN_ERROR_DOMAINS {
  */
 enum NETPLAN_PARSER_ERRORS {
     NETPLAN_ERROR_INVALID_YAML,
-    NETPLAN_ERROR_INVALID_CONFIG
+    NETPLAN_ERROR_INVALID_CONFIG,
+    NETPLAN_ERROR_INVALID_FLAG,
 };
 
 /**
@@ -163,4 +164,12 @@ enum NETPLAN_EMITTER_ERRORS {
  */
 enum NETPLAN_FORMAT_ERRORS {
     NETPLAN_ERROR_FORMAT_INVALID_YAML,
+};
+
+/**
+ * @brief   Flags used to change the parser behavior.
+ */
+enum NETPLAN_PARSER_FLAGS {
+    NETPLAN_PARSER_IGNORE_ERRORS = 1 << 0, ///< Ignore parsing errors such as bad YAML files and definitions.
+    NETPLAN_PARSER_FLAGS_MAX_,
 };
--- a/python-cffi/netplan/_build_cffi.py
+++ b/python-cffi/netplan/_build_cffi.py
@@ -68,6 +68,9 @@ ffibuilder.cdef("""
     // Parser
     NetplanParser* netplan_parser_new();
     void netplan_parser_clear(NetplanParser **npp);
+    gboolean netplan_parser_set_flags(NetplanParser *npp, unsigned int flags, NetplanError** error);
+    unsigned int netplan_parser_get_flags(NetplanParser *npp);
+    unsigned int netplan_parser_get_error_count(NetplanParser *npp);
     gboolean netplan_parser_load_yaml(NetplanParser* npp, const char* filename, NetplanError** error);
     gboolean netplan_parser_load_yaml_from_fd(NetplanParser* npp, int input_fd, NetplanError** error);
     gboolean netplan_parser_load_yaml_hierarchy(NetplanParser* npp, const char* rootdir, NetplanError** error);
--- a/python-cffi/netplan/parser.py
+++ b/python-cffi/netplan/parser.py
@@ -13,10 +13,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from enum import IntEnum
 from typing import Union, IO
 
 from ._netplan_cffi import ffi, lib
 from ._utils import _checked_lib_call
+
+
+class Flags(IntEnum):
+    IGNORE_ERRORS = 1 << 0
 
 
 class Parser():
@@ -46,3 +51,15 @@ class Parser():
     def _load_nullable_overrides(self, input_file: IO, constraint: str):
         return _checked_lib_call(lib.netplan_parser_load_nullable_overrides,
                                  self._ptr, input_file.fileno(), constraint.encode('utf-8'))
+
+    @property
+    def flags(self) -> int:
+        return lib.netplan_parser_get_flags(self._ptr)
+
+    @flags.setter
+    def flags(self, flags: int):
+        _ = _checked_lib_call(lib.netplan_parser_set_flags, self._ptr, flags)
+
+    @property
+    def error_count(self) -> int:
+        return lib.netplan_parser_get_error_count(self._ptr)
--- a/src/generate.c
+++ b/src/generate.c
@@ -44,10 +44,12 @@ static gchar** files;
 static gboolean any_networkd = FALSE;
 static gboolean any_nm = FALSE;
 static gchar* mapping_iface;
+static gboolean ignore_errors = FALSE;
 
 static GOptionEntry options[] = {
     {"root-dir", 'r', 0, G_OPTION_ARG_FILENAME, &rootdir, "Search for and generate configuration files in this root directory instead of /", NULL},
     {G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &files, "Read configuration from this/these file(s) instead of /etc/netplan/*.yaml", "[config file ..]"},
+    {"ignore-errors", 'i', 0, G_OPTION_ARG_NONE, &ignore_errors, "Ignores files and/or network definitions that fail parsing.", NULL},
     {"mapping", 0, 0, G_OPTION_ARG_STRING, &mapping_iface, "Only show the device to backend mapping for the specified interface.", NULL},
     {NULL}
 };
@@ -195,11 +197,14 @@ exit_find:
     return ret;
 }
 
-#define CHECK_CALL(call) {\
-    if (!call) {\
+#define CHECK_CALL(call, ignore_errors) {\
+    if (!call && !ignore_errors) {\
         error_code = 1; \
         fprintf(stderr, "%s\n", error->message); \
         goto cleanup;\
+    } else if (error && ignore_errors) {\
+        fprintf(stderr, "Ignored: %s\n", error->message); \
+        g_clear_error(&error); \
     }\
 }
 
@@ -212,6 +217,7 @@ int main(int argc, char** argv)
     g_autofree char* generator_run_stamp = NULL;
     glob_t gl;
     int error_code = 0;
+    char* ignore_errors_env = NULL;
     NetplanParser* npp = NULL;
     NetplanState* np_state = NULL;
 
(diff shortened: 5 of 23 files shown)


python-cffi/netplan/_build_cffi.py
```diff
@@ -68,6 +68,7 @@
     // Parser
     NetplanParser* netplan_parser_new();
     void netplan_parser_clear(NetplanParser **npp);
+    void netplan_parser_set_flags(NetplanParser *npp, int flags);
```
schopin-pro, line 71: suggestion: we probably want to expose a `netplan_parser_get_flags(npp)` already.

This would give us the liberty to modify the flags as side effects of other operations, like when we're pre-loading data, etc... as the caller wouldn't need to keep an external copy of the flags and could just do `netplan_parser_set_flags(npp, netplan_parser_get_flags(npp) | NETPLAN_PARSER_IGNORE_ERRORS);`

include/parse.h
```diff
@@ -53,6 +53,15 @@ netplan_parser_reset(NetplanParser *npp);
...
 netplan_parser_clear(NetplanParser **npp);
 
+/**
+ * @brief   Set @ref NetplanParser flags.
+ * @details Parser flags are used to change the default behavior of the parser.
+ * @param[in] npp The @ref NetplanParser to set the flags.
+ * @param[in] flags The value of the flags. The possible values are defined in @ref NETPLAN_PARSER_FLAGS
+ */
+NETPLAN_PUBLIC void
+netplan_parser_set_flags(NetplanParser* npp, int flags);
```
schopin-pro, line 63: suggestion: fail on unknown flags

(making the suggestion at the header level as this is API design)

I think it'd make sense to fail loudly when seeing an unknown flag rather than silently ignore it.

python-cffi/netplan/parser.py
```diff
@@ -13,12 +13,17 @@
...
 
+from enum import IntEnum
 from typing import Union, IO
 
 from ._netplan_cffi import ffi, lib
 from ._utils import _checked_lib_call
 
 
+class NETPLAN_PARSER_FLAGS(IntEnum):
+    IGNORE_ERRORS = 1
```
schopin-pro, line 24: suggestion: Just call this class `Flags`

We're already in the `netplan.parser` namespace!

python-cffi/netplan/parser.py
```diff
@@ -46,3 +51,6 @@ def load_nullable_fields(self, input_file: IO):
     def _load_nullable_overrides(self, input_file: IO, constraint: str):
         return _checked_lib_call(lib.netplan_parser_load_nullable_overrides,
                                  self._ptr, input_file.fileno(), constraint.encode('utf-8'))
+
+    def set_flags(self, flags: int):
+        lib.netplan_parser_set_flags(self._ptr, flags)
```
schopin-pro, line 56: suggestion: use a property instead of an actual call

Assuming previous suggestions are implemented, that'd make the following possible:

```py
npp.flags |= Flags.IGNORE_ERRORS
```
rather than
```py
npp.set_flags(saved_flags | NETPLAN_PARSER_FLAGS.IGNORE_ERRORS)
```

src/parse.c
```diff
@@ -3134,6 +3134,23 @@ node_is_nulled_out(yaml_document_t* doc, yaml_node_t* node, const char* key_pref
...
+{
+    if(npp->parsed_defs) {
+        g_hash_table_steal(npp->parsed_defs, netdef->id);
+    }
+    if(npp->ordered) {
+        npp->ordered = g_list_remove(npp->ordered, netdef);
+    }
+
+    reset_netdef(netdef, NETPLAN_DEF_TYPE_NONE, NETPLAN_BACKEND_NONE);
+    g_free(netdef);
```
schopin-pro, line 3151: question: wouldn't that create dangling pointers?

IIRC we have cross-references on netdefs in cases of bridges and the likes, and we try to resolve them on the spot. If that is still true, we should at least add big fat warnings to the docs of this function so that our future selves don't start using it wrongly. Conversely, since we're only dropping the currently-parsed netdef for now we could actually make this function about that one special case?

Related question: don't we write backreferences *in* previously defined netdefs in some cases? It's been a long while since I've looked at the parser code, but I dimly remember that in some cases we go back and modify existing netdefs while parsing other ones, possibly to add a pointer to the new one.

schopin-pro (commented):
Well, that was a fun lunch break :)

include/parse.h
slyon: +1 this should return `gboolean` and take a ` NetplanError** error` argument to be filled with details in case of failure.

src/generate.c
```diff
@@ -44,10 +44,12 @@ static gchar** files;
 static gboolean any_networkd = FALSE;
 static gboolean any_nm = FALSE;
 static gchar* mapping_iface;
+static gboolean ignore_errors = FALSE;
 
 static GOptionEntry options[] = {
     {"root-dir", 'r', 0, G_OPTION_ARG_FILENAME, &rootdir, "Search for and generate configuration files in this root directory instead of /", NULL},
     {G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &files, "Read configuration from this/these file(s) instead of /etc/netplan/*.yaml", "[config file ..]"},
+    {"ignore-errors", 'i', 0, G_OPTION_ARG_NONE, &ignore_errors, "Ignores files and/or interfaces that fail parsing.", NULL},
```
slyon, line 52: nitpick: We shouldn't call it "interfaces". Maybe use "network definitions" or "network definition blocks" instead?

src/parse.c
```diff
@@ -3245,8 +3258,29 @@ handle_network_type(NetplanParser* npp, yaml_node_t* node, const char* key_prefi
...
         }
 
-        if (!process_mapping(npp, value, full_key, handlers, NULL, error))
-            return FALSE;
+        if (!process_mapping(npp, value, full_key, handlers, NULL, error)) {
+            if (npp->flags & NETPLAN_PARSER_IGNORE_ERRORS) {
+                gchar* warning_msg = NULL;
+                if (error && *error)
+                    warning_msg = (*error)->message;
+                g_warning("Skipping interface due to parsing errors. %s: %s", scalar(key), warning_msg);
```
slyon, line 3266: nitpick: "Skipping network definition ..."

include/types.h
```diff
@@ -164,3 +164,10 @@ enum NETPLAN_EMITTER_ERRORS {
 enum NETPLAN_FORMAT_ERRORS {
     NETPLAN_ERROR_FORMAT_INVALID_YAML,
 };
+
+/**
+ * @brief   Flags used to change the parser behavior.
+ */
+enum NETPLAN_PARSER_FLAGS {
+    NETPLAN_PARSER_IGNORE_ERRORS = 1, ///< Ignore parsing errors such as bad YAML files and definitions.
```
slyon, line 172: nitpick: Maybe we should define this as `0x1` (or `0b1` even) to make it more obvious that this is a flag/bit.

src/util.c
```diff
@@ -656,11 +656,21 @@ netplan_parser_load_yaml_hierarchy(NetplanParser* npp, const char* rootdir, GErr
...
 
-    for (GList* i = config_keys; i != NULL; i = i->next)
+    for (GList* i = config_keys; i != NULL; i = i->next) {
         if (!netplan_parser_load_yaml(npp, g_hash_table_lookup(configs, i->data), error)) {
+            if (npp->flags & NETPLAN_PARSER_IGNORE_ERRORS) {
+                gchar* warning_msg = NULL;
+                if (error && *error)
+                    warning_msg = (*error)->message;
+
+                g_warning("Skipping YAML file due to parsing errors. %s", warning_msg);
```
slyon, line 666: todo: We should mention the specific filename + path here, too.

tests/generator/test_errors.py
```diff
@@ -999,3 +999,76 @@ def test_nm_only_supports_unicast_routes(self):
...
+  ethernets:
+    eth123:
+      dhcp4: true
+    eth321:
+      dhcp4: 1''', confs={'b': '''network:
+  bridges:
+    br0:
+      interfaces: [ eth123 ]
+    br1:
+      interfaces: [ eth321 ]'''}, expect_fail=False, skip_generated_yaml_validation=True, ignore_errors=True)
```
slyon, line 1068: question: This is the case Simon described above IIUC. So we would still generate configuration for `br1`, even though `eth321` is broken. That would cause `br1` being unable to be configured at runtime, whith correspondig errors in journal, right?

slyon (commented):
I like the overall approach of this. It is a slight change in behavior, but only for the better, as a non-configured network (due to parsing errors) is always a bad thing. It could lead to unexpected network configuration if some files are ignored but other aren't, but IMO that is still better than no network configuration at all.

include/types.h
schopin-pro: I've seen `1 << 0` (and then `1 << 1`, etc), which means you don't even need to know your powers of 2 nor count the number of zeros ;).

src/parse.c
daniloegea: Yes, definitely. In the end I didn't do it. I decided to leave the netdef "half parsed" in the parser state in case of failure. It makes things simpler and increase the chances you'll have a working configuration in the end (I think).

If we end up generating invalid back end configuration due to this, both NM and networkd will reject the file so it should be fine.

src/util.c
daniloegea: I trust it will be present in the error->message in case of failure...

tests/generator/test_errors.py
daniloegea: Yes, in the end I didn't drop netdefs from the state, so it will generate configuration for all the interfaces. If they are broken because they are half parsed or not validated, I trust the back end will ignore the config file we generate.

daniloegea:
Thanks for your comments, lads. I basically adopted all of them.

I decided to not drop bad netdefs anymore. I think it's actually better to keep them and generate configuration for what was parsed until the point it failed. There is a chance the interface will still be brought up and work if the error happened after settings IPs or DHCP for example.

To make sure (or at least increase the confidence) that these changes will not cause a change in behavior when the flag is enabled and all the configuration is good, I changed our `integration/run.py` script to run each test twice, with and without the flag.

The new "Configuration fuzzing" CI action is catching a memory leak! As far as I can tell, it's happening inside glib and there is  a comment in the code that handles datalists saying it could happen, see FR-5666 for more details. So, this leak is not introduces by these changes.

I prepared a [PPA](https://launchpad.net/~danilogondolfo/+archive/ubuntu/netplan-parser-flags) and added to suggestions of tests in the PR comments.

src/types-internal.h
```diff
@@ -263,6 +263,9 @@ struct netplan_parser {
     GHashTable* null_fields;
     GHashTable* null_overrides;
     GHashTable* global_renderer;
+
+    /* Flags used to change the parser's behavior */
+    unsigned int flags;
```
slyon, line 268: suggestion (non-blocking): I'd love to see an additional boolean field that gets set whenever any parsing actually failed. It should get carrier over from NetplanParser into NetplanState, so we can check the final state, e.g. `State.parsing_errors` (not sure about the naming, yet).

tests/generator/test_errors.py
```diff
@@ -999,3 +999,268 @@ def test_nm_only_supports_unicast_routes(self):
...
+    eth0:
+      dhcp4: true
+    eth1:
+      dhcp4: yesplease
+    eth2:
+      renderer: NetworkManager
+      dhcp4: false''', expect_fail=False, skip_generated_yaml_validation=True, ignore_errors=True)
+
+        self.assertTrue(self.file_exists('10-netplan-eth0.network'))
+        self.assertTrue(self.file_exists('10-netplan-eth1.network'))
```
slyon, line 1017: nitpick (non-blocking): Should we check that the `eth1.network` config does not contain any DHCP config here?

tests/generator/test_errors.py
```diff
@@ -999,3 +999,268 @@ def test_nm_only_supports_unicast_routes(self):
...
+
+    def test_ignore_errors_bad_bond(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    eth0: {}
+  bonds:
+    bond0:
+      interfaces: [eth0]
+      badkey: badvalue''', expect_fail=False, skip_generated_yaml_validation=True, ignore_errors=True)
```
slyon, line 1084: question: I assume this would still be fine if we define `badkey: badvalue` first, before `interfaces: [...]`?

src/generate.c
```diff
@@ -195,11 +197,13 @@ find_interface(gchar* interface, GHashTable* netdefs)
...
 
-#define CHECK_CALL(call) {\
-    if (!call) {\
+#define CHECK_CALL(call, ignore_errors) {\
+    if (!call && !ignore_errors) {\
         error_code = 1; \
         fprintf(stderr, "%s\n", error->message); \
         goto cleanup;\
+    } else if (error && ignore_errors) {\
+        g_clear_error(&error); \
```
slyon, line 206: suggestion: We should consider still logging the error here, as least to debug log.. Maybe using an `IGNORED: ...` prefix?

src/networkd.c
```diff
@@ -620,12 +620,14 @@ write_netdev_file(const NetplanNetDefinition* def, const char* rootdir, const ch
...
-            if (first != def->id) {
-                g_string_free(s, TRUE);
-                return;
+            if (def->veth_peer_link) {
+                gchar* first = g_strcmp0(def->id, def->veth_peer_link->id) < 0 ? def->id : def->veth_peer_link->id;
+                if (first != def->id) {
+                    g_string_free(s, TRUE);
+                    return;
+                }
+                g_string_append_printf(s, "Kind=veth\n\n[Peer]\nName=%s\n", def->veth_peer_link->id);
```
slyon, line 629: nitpick: I wonder if we should still write the `Kind=veth\n` line, even if we don't have `[Peer].Name` defined?

slyon (commented):
Thanks a lot for working through all the remarks. This should be relatively safe now.

The doubling of testing-time is a though sell... OTOH more coverage is better. But at least it should be easy to disable (e.g. through ENV variable), so we can skip the double-testing for local tests.

I left a bunch of additional inline comments that need additional consideration.

src/parse.c
```diff
@@ -3729,6 +3754,21 @@ netplan_parser_clear(NetplanParser** npp_p)
     g_free(npp);
 }
 
+gboolean
+netplan_parser_set_flags(NetplanParser* npp, const unsigned int flags)
```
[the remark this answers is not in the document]
daniloegea: Right, I added it. I wasn't convinced it was necessary as it's a simple operation. But in the future we might have more flags and want to report more details about the error.

tests/generator/test_errors.py
daniloegea: If the error happened before `interfaces: [eth0]` this fields wouldn't be parsed. The `bond0` would still be there, but with an empty list of members.

tests/generator/test_errors.py
daniloegea: I think it's not necessary. There is a test in `test_netplan_parser.c:test_parser_flags_ignore_errors` that checks if `dhcp4` is still false after a failure. It should be enough to check if the netdef field has the expected value I guess, instead of checking if the configuration for it was generated or not...

src/types-internal.h
daniloegea: I like the idea. I just used a different approach, I added a counter to keep track of how many errors were ignored. I added it in two separate commits to make it easier to review.

src/generate.c
```diff
@@ -243,6 +245,9 @@ int main(int argc, char** argv)
     }
 
     npp = netplan_parser_new();
+    if (ignore_errors || called_as_generator)
```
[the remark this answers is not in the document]
daniloegea: As `netplan generate` will not include the `--ignore-errors` (only the generator), I just added a short comment about it there. I added a new explanation section about the generator.

daniloegea:
Thanks Lukas, I tried to address all of your comments.

I added an error counter to the parser state instead of a boolean, that might be more useful I think.
I also added a new parser_error to the enum and exposed it to the python code.
I will squash the two extra commits that contain these changes later.

The first commit was split into the definition and implementation of flags as suggested by @schopin-pro in a conversation we had.

The config_fuzzer was changed to run the generator with the `--ignore-errors` flag for one YAML at a time to avoid that annoying glib memory leak. It's not the ideal test though... I need to investigate how the leak can be identified and ignored.

tests/generator/test_errors.py
slyon: Thanks for the explanation, this should be fine!

slyon (approved):
Thanks for addressing the remarks!

I like the error handling to be integrated with existing error handling logic and the `.error_count` field.

PR #413, opened 2023-10-06
parse: improve the parsing of access-points (LP: #1809994)

daniloegea:
When the parser requires a second pass, it will report access-points it already parsed as duplicates and error out.

With this change, access-points will be stored in a local hash table, which will be used to find duplicates in the same netdef, and moved to the netdef->access_points hash table in the end. Access-points already present in the netdef hash table will be considered already parsed and skipped.

Add a test to check the problem is fixed and another one to test if merging the access-points when the same interface is present in multiple files is working properly.

## Checklist


src/parse.c
```diff
@@ -1437,31 +1442,43 @@ handle_wifi_access_points(NetplanParser* npp, yaml_node_t* node, const char* key
                 continue;
         }
 
+        ssid = scalar(key);
+
+        /* Skip access-points that already exist in the netdef */
+        if (npp->current.netdef->access_points && g_hash_table_contains(npp->current.netdef->access_points, ssid))
+            continue;
```
slyon, line 1449: thought: This ignores the definition of an access-point/SSID if it was already defined in a previous YAML file, AFAIU.

Usually we allow merging of any settings in our schema, where settings from "higher" files (sorting later lexicographically) will win over settings of "lower" files. Here we seem to ignore the "higher" SSID, keeping the "lower" one around. Would it make sense to clear the lower SSID instead and put the new one in place?

src/parse.c
```diff
@@ -1437,31 +1442,43 @@ handle_wifi_access_points(NetplanParser* npp, yaml_node_t* node, const char* key
...
+
+        /* Check if there's already an SSID with that name in the list of APs we are parsing */
+        if (g_hash_table_contains(access_points, ssid)) {
```
slyon, line 1452: nitpick: This is about the SSIDs defined inside the YAML file being parsed currently, IIUC. Maybe the comment could be clarified in that regard.

slyon (approved):
Thank you, this looks mostly good to me! There's one thought mentioned inline, which should be fixed, though.

Also, I think we have a more generic issue here, too: Usually we allow merging of NetDefs on a per-setting level, while we're now skipping/replacing full access-points/SSID blocks, just merging full access-points/SSID blocks from different files (ignoring their individual settings). This is still better than what we had before, where we didn't allow skipping/replacing any duplicated SSID blocks. But we should think about how we could allow merging of individual AP settings across YAML files.

src/parse.c
daniloegea: yeah that might be a bit better. One would expect a behavior like that from Netplan I guess. I'll give it a try...

daniloegea: I implemented that in a new commit. Let me know if that's what you had in mind.

slyon (approved):
lgtm.

isness:
When will the update be released? I have this issue on multiple Ubuntu 23.10 machines that results in them becoming unable to connect until network definitions are manually deleted from /etc/netplan. Manually deleting them is not that user-friendly for ordinary users.

slyon:
It has been released as part of the [0.107.1](https://github.com/canonical/netplan/releases/tag/0.107.1) bugfix release, currently available in Ubuntu 24.04 "Noble Numbat" development release and Debian testing.
We plan to backport this into Ubuntu 23.10 (and 22.04) in the March/April timeframe.

slyon:
Actually, this seems to be fixed in Ubuntu 23.10 already (as of Nov 15), see https://launchpad.net/ubuntu/+source/netplan.io/0.107-5ubuntu0.2

@isness Can you confirm this?

PR #414, opened 2023-10-10
wireguard: ignore empty endpoints (LP: #2038811)

daniloegea:
The Network Manager GUI will emit 'Endpoint=' if it's not specified when creating the tunnel. libnetplan is generating 'endpoint: ""' in this case and the parser is returning an error.

In nm-parse, if Endpoint is either empty or NULL we'll just not read it. In the parser, if we find 'endpoint: ""', we'll just ignore it as if it were not defined.

See LP: #2038811

## Checklist

### DIFF
--- a/src/parse-nm.c
+++ b/src/parse-nm.c
@@ -552,7 +552,11 @@ parse_tunnels(GKeyFile* kf, NetplanNetDefinition* nd)
                 }
 
                 /* Handle endpoint */
-                wireguard_peer->endpoint = g_key_file_get_string(kf, group, "endpoint", NULL);
+                gchar* endpoint = g_key_file_get_string(kf, group, "endpoint", NULL);
+                if (endpoint && g_strcmp0(endpoint, "")) {
+                    /* Only set the endpoint if it's not NULL nor an empty string */
+                    wireguard_peer->endpoint = endpoint;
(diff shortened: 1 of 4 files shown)


src/parse-nm.c
```diff
@@ -552,7 +552,11 @@ parse_tunnels(GKeyFile* kf, NetplanNetDefinition* nd)
                 }
 
                 /* Handle endpoint */
-                wireguard_peer->endpoint = g_key_file_get_string(kf, group, "endpoint", NULL);
+                gchar* endpoint = g_key_file_get_string(kf, group, "endpoint", NULL);
+                if (g_strcmp0(endpoint, "")) {
```
slyon, line 556: This does not catch the `endpoint = NULL` case IMO. g_strcmp0 will return -1 when comparing NULL and the empty string, which resolves to `TRUE`.
```suggestion
                if (endpoint && g_strcmp0(endpoint, "")) {
```

slyon (approved):
LGTM, except for a little inline comment, which should probably be fixed.

src/parse-nm.c
daniloegea: Ops, I guess I misinterpreted the man page :P

PR #415, opened 2023-10-10
auth: add support for LEAP and EAP-PWD (LP: #2038811)

daniloegea:
This changes add support for the LEAP and PWD EAP authentication methods for NetworkManager, wpa_supplicant (networkd) and the keyfile parser.

Using these method through the NM GUI is causing problems because libnetplan will generate a broken keyfile. See LP: #2038811

Note that the NM GUI supports more methods not recognized by netplan and we probably should also implement them later.

## Checklist


src/networkd.c
```diff
@@ -1091,6 +1091,14 @@ append_wpa_auth_conf(GString* s, const NetplanAuthenticationSettings* auth, cons
             g_string_append(s, "  eap=TTLS\n");
             break;
 
+        case NETPLAN_AUTH_EAP_LEAP:
+            g_string_append(s, "  eap=LEAP\n");
+            break;
+
+        case NETPLAN_AUTH_EAP_PWD:
+            g_string_append(s, "  eap=PWD\n");
```
slyon, line 1099: question: could you please link some official reference for those wpa_supplicant settings? I couldn't find anything when looking it up quickly.

src/parse-nm.c
```diff
@@ -386,6 +386,10 @@ parse_dot1x_auth(GKeyFile* kf, NetplanAuthenticationSettings* auth)
             auth->eap_method = NETPLAN_AUTH_EAP_PEAP;
         } else if (g_strcmp0(first_method, "ttls") == 0) {
             auth->eap_method = NETPLAN_AUTH_EAP_TTLS;
+        } else if (g_strcmp0(first_method, "leap") == 0) {
+            auth->eap_method = NETPLAN_AUTH_EAP_LEAP;
+        } else if (g_strcmp0(first_method, "pwd") == 0) {
```
slyon, line 391: thought: We probably want some `else` case here, for any other (future) method, unknown by Netplan. Those should be handled via the `networkmanager.passthrough` setting.

As you mentioned there are more, e.g. "md5", "fast", ...

src/nm.c
```diff
@@ -417,6 +417,12 @@ write_dot1x_auth_parameters(const NetplanAuthenticationSettings* auth, GKeyFile
         case NETPLAN_AUTH_EAP_TTLS:
             g_key_file_set_string(kf, "802-1x", "eap", "ttls");
             break;
+        case NETPLAN_AUTH_EAP_LEAP:
+            g_key_file_set_string(kf, "802-1x", "eap", "leap");
+            break;
+        case NETPLAN_AUTH_EAP_PWD:
+            g_key_file_set_string(kf, "802-1x", "eap", "pwd");
```
slyon, line 424: note: ACK. As descriibed in https://developer-old.gnome.org/NetworkManager/stable/settings-802-1x.html

slyon (approved):
Thank you very much! I think this should fix the case for the "leap" and "pwd" values of NetworkManager's `801-1x.eap` setting. But there might be more, similar cases and we should think about implementing some generic (future proof) fallback, using `networkmanager.passthrough`.

src/networkd.c
daniloegea: Yeah this thing is not documented! I had to look it up in the hostapd/wpa_supplicant code to be sure it existed.

Example in test scripts:
https://w1.fi/cgit/hostap/tree/tests/hwsim/test_eap_proto.py#n7699

There is also a brief note about it here https://w1.fi/cgit/hostap/plain/hostapd/hostapd.conf

daniloegea:
Something is still missing, the command below is still failing:

```
nmcli con add type wifi ifname wlan0 ssid asdasd wifi-sec.key-mgmt wpa-eap 802-1x.eap md5 802-1x.identity username 802-1x.password aaaaaaaa
```

src/parse-nm.c
slyon: Okay, it seems like the generic/fallback case is a bit more involved. Running into specific edge-cases. We will handle that in a separate PR.

slyon (approved):
LGTM with the reduced scope. Let's fix the bug at hand first and work on the generic solution afterwards.

nitpick: May I just suggest to add simple testcases for the new "leap"/"pwd" EAP methods in `test_keyfile.py`?

```diff
diff --git a/tests/parser/test_keyfile.py b/tests/parser/test_keyfile.py
index dbb992ed..7a6bd2ca 100644
--- a/tests/parser/test_keyfile.py
+++ b/tests/parser/test_keyfile.py
@@ -526,6 +526,12 @@ dns-search='''.format(UUID, method))
     def test_keyfile_type_wifi_eap_ttls(self):
         self._template_keyfile_type_wifi_eap('ttls')
 
+    def test_keyfile_type_wifi_eap_leap(self):
+        self._template_keyfile_type_wifi_eap('leap')
+
+    def test_keyfile_type_wifi_eap_pwd(self):
+        self._template_keyfile_type_wifi_eap('pwd')
+
     def test_keyfile_type_wifi_eap_leap(self):
         self.generate_from_keyfile('''[connection]
 type=wifi
```

PR #421, opened 2023-11-06
check if ovsdb-server.service is active before displaying warning

viciousdrop:
## Description
Made changes outlined in this bug
Credit:  @brianfinley https://launchpad.net/~finley
https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2041727 
Tested on Ubuntu 22.04.3 LTS and the error message no longer displays

## Checklist
```
@@ -414,4 +414,5 @@
             if exit_on_error:
                 sys.exit(1)
         except OvsDbServerNotRunning as e:
- logging.warning('Cannot call Open vSwitch: {}.'.format(e))
+ if utils.systemctl_is_active('ovsdb-server.service'):
+ logging.warning('Cannot call Open vSwitch: {}.'.format(e))
```
- [X] Closes an open bug in Launchpad.
https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2041727

### DIFF
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -31,7 +31,7 @@ import time
 from .. import utils
 from ...configmanager import ConfigManager, ConfigurationError
 from ..sriov import apply_sriov_config
-from ..ovs import OvsDbServerNotRunning, apply_ovs_cleanup
+from ..ovs import OvsDbServerNotRunning, OvsDbServerNotInstalled, apply_ovs_cleanup
 
 
 OVS_CLEANUP_SERVICE = 'netplan-ovs-cleanup.service'
@@ -433,3 +433,5 @@ class NetplanApply(utils.NetplanCommand):
                 sys.exit(1)
         except OvsDbServerNotRunning as e:
             logging.warning('Cannot call Open vSwitch: {}.'.format(e))
+        except OvsDbServerNotInstalled as e:
+            logging.debug('Cannot call Open vSwitch: %s.', e)
--- a/netplan_cli/cli/ovs.py
+++ b/netplan_cli/cli/ovs.py
@@ -20,7 +20,7 @@ import os
 import subprocess
 import re
 
-from .utils import systemctl_is_active
+from .utils import systemctl_is_active, systemctl_is_installed
 
 OPENVSWITCH_OVS_VSCTL = '/usr/bin/ovs-vsctl'
 OPENVSWITCH_OVSDB_SERVER_UNIT = 'ovsdb-server.service'
@@ -40,6 +40,10 @@ GLOBALS = {
 
 
 class OvsDbServerNotRunning(Exception):
+    pass
+
+
+class OvsDbServerNotInstalled(Exception):
     pass
 
 
@@ -125,6 +129,9 @@ def apply_ovs_cleanup(config_manager, ovs_old, ovs_current):  # pragma: nocover 
     Also filter for individual settings tagged netplan/<column>[/<key]=value
     in external-ids and clear them if they have been set by netplan.
     """
+    if not systemctl_is_installed(OPENVSWITCH_OVSDB_SERVER_UNIT):
+        raise OvsDbServerNotInstalled("Cannot apply OVS cleanup: %s is 'not-found'.",
+                                      OPENVSWITCH_OVSDB_SERVER_UNIT)
     if not systemctl_is_active(OPENVSWITCH_OVSDB_SERVER_UNIT):
         raise OvsDbServerNotRunning('{} is not running'.format(OPENVSWITCH_OVSDB_SERVER_UNIT))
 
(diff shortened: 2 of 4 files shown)


netplan_cli/cli/commands/apply.py
```diff
@@ -414,4 +414,5 @@ def process_ovs_cleanup(config_manager, ovs_old, ovs_current, exit_on_error=True
             if exit_on_error:
                 sys.exit(1)
         except OvsDbServerNotRunning as e:
-            logging.warning('Cannot call Open vSwitch: {}.'.format(e))
+            if utils.systemctl_is_active('ovsdb-server.service'):
```
slyon, line 417: Hmm, `utils.systemctl_is_active()` is the method that is actually used to produce the `OvsDbServerNotRunning` error in the first place. So muting the warning message on that condition is doing unnecessary work, just to ignore it afterward.

I think we should rather consider implementing something like `utils.systemctl_list_units()` to check if that service unit is actually installed, and then skip all of `apply_ovs_cleanup()` (in `ovs.py`) in that case.

Additionally, we should make use of the `OPENVSWITCH_OVSDB_SERVER_UNIT` constant, instead of putting the `ovsdb-server.service` name directly.

slyon (changes requested):
Thank you very much @viciousdrop for your contribution and picking up the information from the bug report. Unfortunately, I think it's a little bit more complicated than that and we should actually check if OVS is installed on the system, instead of just hiding the warning message unconditionally.

See my inline comment.

slyon:
I rebased your branch and adopted it according to my comments above.

What do you think about this changeset @viciousdrop and @daniloegea ?

netplan_cli/cli/ovs.py
```diff
@@ -125,6 +125,9 @@ def apply_ovs_cleanup(config_manager, ovs_old, ovs_current):  # pragma: nocover
     Also filter for individual settings tagged netplan/<column>[/<key]=value
     in external-ids and clear them if they have been set by netplan.
     """
+    if not systemctl_is_installed(OPENVSWITCH_OVSDB_SERVER_UNIT):
+        logging.debug("Cannot apply OVS cleanup: %s is 'not-found'.", OPENVSWITCH_OVSDB_SERVER_UNIT)
+        return False
```
daniloegea, line 130: Hmm this method doesn't really return a value. This is the only explicit return point in the code. We probably should throw an exception here and handle it as we handle `OvsDbServerNotRunning`.

slyon: Fair point! I've updated it accordingly, catching the error in `apply.py`, then logging a debug message that is usually muted.

@daniloegea What do you think now?

daniloegea (approved):
lgtm

viciousdrop:
No issues. Lgtm. Thanks for taking the time to correct this.

slyon (approved):
OK, let's get this merged.

dauji:
I'm still encountering the same error, when I'm modifying and applying the 00-installer-config.yaml

haywoodspartan:
I have this issue when deploying Kolla-Ansible and the function fails on checking bonded ports provided by netplan
This was a non issue prior to OSP ver 2023.1 currently if I try to apply a netplan apply while even the containerized ovsdb-server and openvswitch-vswitchd is running the netplan does not pick up the port for ovsdb server even if it is current up.
This causes the dataplane to break for all associated VLAN's on the physical bond

slyon:
> I'm still encountering the same error, when I'm modifying and applying the 00-installer-config.yaml

@dauji Did you use the latest Netplan version from git for your test? I cannot seem to reproduce this with the latest version.

slyon:
> I have this issue when deploying Kolla-Ansible and the function fails on checking bonded ports provided by netplan
> This was a non issue prior to OSP ver 2023.1 currently if I try to apply a netplan apply while even the containerized ovsdb-server and openvswitch-vswitchd is running the netplan does not pick up the port for ovsdb server even if it is currently up and running.
> This causes the dataplane to break for all associated VLAN's on the physical bond.

This seems to happen in a very specific context. Can you please open a [bug report](https://bugs.launchpad.net/netplan/+filebug) about it?

Kunzol:
Any thought about the situation if _netplan_ config contains _openvswitch_ configuration, but _openvswitch_ is not installed or not active?

slyon:
> Any thought about the situation if _netplan_ config contains _openvswitch_ configuration, but _openvswitch_ is not installed or not active?

In that case you should probably get a `OvsDbServerNotInstalled` exception in your debug log. E.g. something like this (inside the `netplan --debug apply` output):

> Cannot apply OVS cleanup: ovsdb-server.service is 'not-found'

vitalii-sharapov:
today is 21st Aug 2024, just wanted to add that on freshly installed Ubuntu 22.04 this issue still exists.

slyon:
The SRU in https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2058031 is current blocked due to a regression in juju. But you can already install `netplan.io` from `jammy-proposed` to get the fix on Ubuntu 22.04.

PR #422, opened 2023-11-15
Fix permissions on folder '/run/NetworkManager/'

matdug:
When the Netplan systemd generator runs before the NetworkManager service, Netplan will preempt NetworkManager from creating folder '/run/NetworkManager' and will assign it permissions 700.
This is a side effect of the call to umask in function write_nm_conf_access_point, made to restrict the permissions on subfolder '/run/NetworkManger/system-connections'. A regular user needs to read from /run/NetworkManager where, in some systems, files such as resolv.conf ultimately reside.

## Description

In my Yocto image /etc/resolv.conf  is a symlink to /etc/resolv-conf.NetworkManager itself to /run/Netw
orkManager/resolv.conf.  Because the /run/NetworkManager folder created by Netplan has 700 permissions a regular user has no access to resolv.conf which breaks domain name resolution.

This fix assigns the proper access to rights to /run/NetworkManager, 755,  if Netplan happens to be the first to create it, without elevating privileges to /run/NetworkManager/system-connections, which is what was intended by the original code.

## Checklist

### DIFF
--- a/src/nm.c
+++ b/src/nm.c
@@ -618,6 +618,7 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
     g_autoptr(GKeyFile) kf = NULL;
     g_autofree gchar* conf_path = NULL;
     g_autofree gchar* full_path = NULL;
+    g_autofree gchar* nm_run_path = NULL;
     g_autofree gchar* nd_nm_id = NULL;
     const gchar* nm_type = NULL;
     gchar* tmp_key = NULL;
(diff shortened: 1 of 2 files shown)


src/nm.c
```diff
@@ -966,8 +967,16 @@ write_nm_conf_access_point(const NetplanNetDefinition* def, const char* rootdir,
         }
     }
 
-    /* NM connection files might contain secrets, and NM insists on tight permissions */
+    /* Create /run/NetworkManager/ with 755 permissions if the folder is missing.
+     * Letting the next invokation of safe_mkdir_p_dir do it would result in 
+     * more restrictive access because of the call to umask. */
+    nm_run_path = g_strjoin(G_DIR_SEPARATOR_S, rootdir ?: "", "run/NetworkManager/", NULL);
+    if (!g_file_test(nm_run_path, G_FILE_TEST_EXISTS))
```
slyon, line 974: nitpick: you had some leading whitespace here, which doesn't match our code style, so I quickly fixed this up.

slyon (approved):

This is a good catch and you're very right! `/run/NetworkManager/` should be `0o755`, while only `/run/NetworkManager/system-connections/` should be using `0o700` permissions (and we have tests in `tests/generator/base.py` to actually check for the latter).

matdug:
Thanks for the merge, and the edits

PR #423, opened 2023-11-27
doc: libnetplan API reference, using Doxygen and Sphinx.breathe

slyon:
## Description
Setting up a structure to make use of [Doxygen](https://www.doxygen.nl/) and Sphinx' [Breathe](https://breathe.readthedocs.io/) the generate a API documentation reference for `libnetplan` directly from the source and header files.

The documentation is generated at doc-build time (`make run` inside `doc/`) and also run on readthedocs, so we show the most current documentation from the sources at all times.

We make use of Breathe's `autodoxygen` features, so that we don't need to generate and parse Doxygen's XML output manually.

FR-3781

## Checklist


slyon:
We might even consider dropping the top-level `Doxyfile` configuration, I think it's not necessarily needed.

include/util.h
```diff
@@ -15,12 +15,27 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+/*! \file util.h
+ *  \brief A set of high level helper functions, that can be used when working
```
rkratky, line 19:
```suggestion
 *  \brief A set of high-level helper functions that can be used when working
```

include/util.h
```diff
@@ -15,12 +15,27 @@
...
+ *  with datastructures of the library.
+ *
+ *  For example it can be used to initialize iterators or clear error structures.
```
rkratky, line 22:
```suggestion
 *  For example, it can be used to initialize iterators or clear error structures.
```

include/util.h
```diff
@@ -15,12 +15,27 @@
...
+ */
+
 #pragma once
 
 #include <glib.h>
 #include <stdint.h>
 #include "types.h"
 
+/**
+ * @brief Parses YAML hierarcy from @rootdir, drops the configuration for @id
```
rkratky, line 32:
```suggestion
 * @brief Parses YAML hierarchy from @rootdir, drops the configuration for @id
```

include/util.h
```diff
@@ -62,8 +77,14 @@ netplan_util_dump_yaml_subtree(const char* prefix, int input_fd, int output_fd,
 
 /********** Old API below this ***********/
 
+/**
+ * \deprecated Make use of `netplan_netdef_get_filepath()` instead.
```
rkratky, line 81:
```suggestion
 * \deprecated Use of `netplan_netdef_get_filepath()` instead.
```

include/util.h
```diff
@@ -62,8 +77,14 @@ netplan_util_dump_yaml_subtree(const char* prefix, int input_fd, int output_fd,
...
+ */
 NETPLAN_DEPRECATED NETPLAN_PUBLIC gchar*
 netplan_get_filename_by_id(const char* netdef_id, const char* rootdir);
 
+/**
+ * \deprecated Make use of `netplan_get_id_from_nm_filepath()` instead.
```
rkratky, line 87:
```suggestion
 * \deprecated Use of `netplan_get_id_from_nm_filepath()` instead.
```

rkratky (approved):
@slyon A few nitpicks.

include/util.h
slyon: This should probably be just "Use `netplan..."

slyon:
Thanks a lot for the prompt review! I addressed all the nitpicks. Will merge once the CI is (mostly) green.

PR #425, opened 2023-12-02
RFC: Support consuming Open vSwitch from snap

fnordahl:
## Description
There are multiple projects in development that provide Open vSwitch in a snap (MicroOVN/MicroCloud, Sunbeam and possibly others).  These projects have use for the Netplan Open vSwitch integration.

I had use for this functionality in my homelab, so as an itch to scratch I'm proposing this early to get feedback on the approach.

There are a couple of TODO's that I'm still working on:
* Figure out how to hanlde the ovsdb-server.service unit check
* Add test cases that checks the snap workflow

Please review commit by commit.

## Checklist


src/openvswitch.c
```diff
@@ -29,6 +30,48 @@
...
+                 sizeof netplan_openvswitch_ovs_vsctl_path,
+                 BINDIR"/"OPENVSWITCH_OVS_VSCTL);
+    }
+
+}
+
+const char *
+netplan_openvswitch_ovs_vsctl(void)
+{
+    if (!*netplan_openvswitch_ovs_vsctl_path) {
```
slyon, line 69: Can we be sure this is always null-initialized? Some architectures/compilers might not do that for us. We could probably do that above, or use a higher level concept, like glib's [GString](https://docs.gtk.org/glib/struct.String.html), that takes care of that for us.

Another thought/nitpick: What happens to a longer-running Netplan process (e.g. a daemon linking to `libnetplan`), where the OVS snap gets installed midway through the Netplan process' lifetime? I think it would continue calling into the system-OVS, which might or might not be OK. Granted, that is more of a future topic, as the OVS generator is mostly executed in isolation currently.

src/openvswitch.h
```diff
@@ -29,3 +29,6 @@ netplan_netdef_write_ovs(
 
 NETPLAN_INTERNAL gboolean
 netplan_ovs_cleanup(const char* rootdir);
+
+NETPLAN_INTERNAL const char *
```
slyon, line 33: We might want to drop the `NETPLAN_INTERNAL` attribute here. It will make the `netplan_openvswitch_ovs_vsctl` symbol visible as part of the library, which we don't need here, it should be part of all the relevant object files. (Yes, the NETPLAN_INTERNAL naming is confusing here, sorry for that.)

Should it be needed, the new symbol should be prefixed with "_" (`_netplan_openvswitch_ovs_vsctl`) to clearly mark it as private API.

tests/ctests/meson.build
```diff
@@ -1,3 +1,10 @@
+project('ctests', 'c')
```
slyon, line 1: I'd like to get @daniloegea's opinion on this new project structure. Overall, it looks fine to me and I'm happy to change it that way, as long as we can run our tests as usual and don't need to adopt the packaging.

netplan_cli/cli/ovs.py
```diff
@@ -20,10 +20,9 @@
 import subprocess
 import re
 
-from .utils import systemctl_is_active
-
-OPENVSWITCH_OVS_VSCTL = '/usr/bin/ovs-vsctl'
-OPENVSWITCH_OVSDB_SERVER_UNIT = 'ovsdb-server.service'
+OPENVSWITCH_OVS_VSCTL = (
```
slyon, line 23: thought: We also have some NetworkManager.snap handling in `cli/utils.py:is_nm_snap_enabled()`, which handles a similar situation a bit differently (checking for the snap related systemd service). Both have in common that the snap path gets a higher priority.

My initial idea was to bring the OVS handling in line with the NM handling, using a similar logic. But OTOH, we also have the changes in `src/openvswitch.c` (that we don't have for NM.snap), mimicking the path-fallback logic implemented here. So it might be better to keep the OVS-VSCL handling in-line between .c and .py and rather adopt/refactor the NM.snap handling at some point in the future.

netplan_cli/cli/ovs.py
```diff
@@ -43,6 +42,19 @@ class OvsDbServerNotRunning(Exception):
...
 
+def _assert_ovsdb_server_connection():
+    """Invoke OPENVSWITCH_OVS_VSCTL, raise OvsDbServerNotRunning on error."""
+    try:
+        # The `get-manager` command is a light weight database operation which
+        # we use as indication of the client being able to connect.
+        subprocess.check_call(
+            [OPENVSWITCH_OVS_VSCTL, '--timeout', '5', 'get-manager'])
+    except (subprocess.CalledProcessError, OSError):
+        raise OvsDbServerNotRunning('{} is unable to connect to database, '
```
slyon, line 53: note: This might have some implications for https://github.com/canonical/netplan/pull/421

slyon (commented):
Thank you very much for bringing this up early, Frode!

Besides the obvious CI failures that need fixing, the approach LGTM!
```
# COVERAGE
11/12 coverage-c          FAIL             0.00s   exit status 1
12/12 coverage-py         FAIL             0.61s   exit status 2

# AUTOPKGTEST-CI
FAIL: test_ovsdb_server_is_not_running (__main__.TestOVS)
```

I left some inline-comments for more specific issues that need some extra thoughs or discussions.

tests/ctests/meson.build
daniloegea: The ctests were intentionally not linked against libnetplan. The idea was to build only the necessary files with the test files. The problem is that each file has dependencies spread across almost all the others so to get it working properly we need to do a good refactoring in the C code...

I think it should be fine to separate and link them against libnetplan to make things easier for the time being as long as we still can run them as part of `meson test` and the coverage checks still work.

daniloegea: thought: would adding `link_with: libnetplan` to `tests/ctests/meson.build` work?

slyon:
Closing in favour of https://github.com/canonical/netplan/pull/549

PR #426, opened 2023-12-06
state_diff: fix filtering of host scoped routes

daniloegea:
When an interface has multiple IPs in the same subnet, there will be host scoped routes for each of them and the "src" address will be one of the IPs for all of them.

The filtering of this type of route was not working properly and including the extra routes in the diff.

This change generalizes the verification to see if the destination address is one of the system own addresses instead of comparing it with the src address.

While here, fix the data structure passed to the filtering method. It should be a list but a dictionary was being passed to it.

## Checklist

### DIFF
--- a/netplan_cli/cli/state_diff.py
+++ b/netplan_cli/cli/state_diff.py
@@ -298,7 +298,7 @@ class NetplanDiffState():
         netplan_routes = self._normalize_routes(netplan_routes)
 
         # Filter out some routes that are expected to be added automatically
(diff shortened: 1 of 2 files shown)


daniloegea:
This patch was integrated in this PPA https://launchpad.net/~danilogondolfo/+archive/ubuntu/netplan-diff/

slyon (approved):
LGTM, in order to fix the current logic, passing a dict instead of a list.

But beware that in spec `FO113` there's some discussion about the filtering of routes. So this logic might change in the future.

PR #427, opened 2023-12-07
parser: accept the special MAC address options (LP: #2045096)

daniloegea:
Besides the actual MAC address value, Network Manager also accepts the following values: preserve, permanent, random and stable.

When set via NetworkManager, these special values would end up in the passthrough section. But if users try to set it manually, the parsing would fail.

This changes add support for these special values for the NetworkManager renderer and return an error if the user tries to use them with networkd.

Related to https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/2045096

## Checklist

### DIFF
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -491,7 +491,10 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
 - **`macaddress`** (scalar)
 
   > Set the device's MAC address. The MAC address must be in the form
-  > `XX:XX:XX:XX:XX:XX`.
+  > "XX:XX:XX:XX:XX:XX". The following special options are also accepted:
+  > `permanent` and `random`.
+  > In addition to these options, the NetworkManager renderer also accepts
+  > `stable` and `preserve`.
   >
   > **Note:** This will not work reliably for devices matched by name
   > only and rendered by networkd, due to interactions with device
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -233,6 +233,7 @@ write_link_file(const NetplanNetDefinition* def, const char* rootdir, const char
     if (!def->set_name &&
         !def->wake_on_lan &&
         !def->mtubytes &&
+        !(_is_macaddress_special_nd_option(def->set_mac) && def->backend == NETPLAN_BACKEND_NETWORKD) &&
         (def->receive_checksum_offload == NETPLAN_TRISTATE_UNSET) &&
         (def->transmit_checksum_offload == NETPLAN_TRISTATE_UNSET) &&
         (def->tcp_segmentation_offload == NETPLAN_TRISTATE_UNSET) &&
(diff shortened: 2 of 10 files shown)


slyon (commented):
This is OKish, I guess.
Although, I don't like the fact of diverging possible values, depending on the chosen backend too much.

I wonder if similar behavior could be implemented on the "networkd" backend, using udev's `[Link].MACAddressPolicy=` setting? https://www.freedesktop.org/software/systemd/man/latest/systemd.link.html#MACAddressPolicy=

As I see from `parse-nm.c` those cases have already been handled for NetworkManager integration (using "passthrough"). I wonder how this bug was hit in the first place? Was it a manual modification of the NetworkManager genrated YAML by the user?

daniloegea:
I assumed the used modified it manually, yes. Although I can't confirm that (I saw you just asked them).

I'll look at the networkd options, supporting them sounds like a good idea.

Thanks!

daniloegea:
So, I added support for the options supported by networkd as well, "persistent" and "random". It was a bit harder because it involves multiple files and the option is different (MACAddress vs MACAddressPolicy).

doc/netplan-yaml.md
```diff
@@ -491,7 +491,10 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
 - **`macaddress`** (scalar)
 
   > Set the device's MAC address. The MAC address must be in the form
-  > `XX:XX:XX:XX:XX:XX`.
+  > "XX:XX:XX:XX:XX:XX". When using `NetworkManager` as renderer, the following
+  > values are also accepted: `preserve`, `permanent`, `random` and `stable`.
+  > When using `networkd`, the following values are also accepted: `persistent`,
+  > `random`.
```
slyon, line 497: IIUC networkd's `persistent` is the same as NetworkManager's `permanent` [1,2], where "permanent MAC" is the more common terminology IMO.

We should probably combine persistent+permanent into a single "permanent" Netplan value. Also "random" is the same for both backend anyway. So we'd only need to specify the `preserve` and `stable` special values for NM, everything else could be shared. Alternatively, "persistent" could be a (secondary) alias to "permanent".

We want to reduce such differences and make Netplan do the right thing, depending on the chosen renderer.

[1] https://developer-old.gnome.org/NetworkManager/stable/settings-802-3-ethernet.html
[2] https://www.freedesktop.org/software/systemd/man/latest/systemd.link.html#MACAddressPolicy=

src/networkd.c
```diff
@@ -234,6 +234,7 @@ write_link_file(const NetplanNetDefinition* def, const char* rootdir, const char
     if (!def->set_name &&
         !def->wake_on_lan &&
         !def->mtubytes &&
+        !(_is_macaddress_special_nd_option(def->set_mac) && def->backend != NETPLAN_BACKEND_NM) &&
```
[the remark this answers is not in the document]
daniloegea: I changed this to `== NETPLAN_BACKEND_NETWORKD` as it's only required by the networkd renderer.

doc/netplan-yaml.md
daniloegea: I'm using `permanent` for both renderers now.

daniloegea:
Thanks Lukas, I tried to address all of your comments. As for the integration test, I couldn't get the option `random` working in our autopkgtest environment. I believe it has to do with the virtual ethernet we use to simulate an ethernet interface. It works fine with a real ethernet interface though. It also works when NetworkManager is used so I added a test for NetworkManager only.

slyon (approved):
Thanks, LGTM!

> Is the documentation checks failing because of this?
> 
> ( index: line 70) broken https://askubuntu.com/questions/tagged/netplan - 403 Client Error: Forbidden for url: https://askubuntu.com/questions/tagged/netplan

Yes, we can ignore that for now.

PR #428, opened 2023-12-13
parse-nm/wg: append the correct prefix to IPv6 addresses (LP: #2046158)

daniloegea:
When the prefix is omitted for IPs in the allowed-ips list, we were appending a /32 to them without checking the address family.

IPv6 addresses will have a /128 appended to them if it's not present.

See LP#2046158

## Checklist


slyon (approved):
Thank you, LGTM. The modified unit-test is covering the case.

PR #429, opened 2023-12-14
Adopt Docs Starter Pack

rkratky:
## Description

* Exclude changes in 'doc/' dir from code CI checks.
* Adopt Docs Starter Pack:
    - Add CI linkcheck
    - Substitute spellcheck & incl. lang. check w/ solution from Starter Pack
    - Fix ton of spelling mistakes
    - Update & fix markup for consistency
    - Use redirects instead of symlinks
* Fix Doxygen config. & remove Doxyfile.

## Checklist

- [x] Runs `make check` successfully.

- [x] Retains 100% code coverage (`make check-coverage`).

- [ ] N/A: New/changed keys in YAML format are documented.
- [ ] N/A: \(Optional\) Adds example YAML for new feature.
- [x] \(Optional\) Closes an open bug in Launchpad. -- Fixes [FR-6115](https://warthogs.atlassian.net/browse/FR-6115), [FR-6157](https://warthogs.atlassian.net/browse/FR-6157)

[FR-6115]: https://warthogs.atlassian.net/browse/FR-6115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FR-6157]: https://warthogs.atlassian.net/browse/FR-6157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #430, opened 2023-12-15
Fix howto docs

rkratky:
## Description

This is a follow up to #429 -- don't merge before it.

Improve how-to docs:
    
- Move intro to how-to landing page
- Make headings consistently use "How to"
- Remove (post-build) "How to" from TOC

## Checklist

- [ ] N/A: New/changed keys in YAML format are documented.
- [ ] N/A: \(Optional\) Adds example YAML for new feature.
- [x] \(Optional\) Closes an open bug in Launchpad. Fix [FR-6139](https://warthogs.atlassian.net/browse/FR-6139)

[FR-6139]: https://warthogs.atlassian.net/browse/FR-6139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/howto.md
+++ b/doc/howto.md
@@ -1,6 +1,26 @@
 # How-to guides
 
+Below is a collection of how-to guides for common scenarios.
+If you see a scenario missing or have one to contribute, please,
+[file a bug](https://bugs.launchpad.net/netplan/+filebug) against this
+documentation with the example.
+
+To configure Netplan, save configuration files in the `/etc/netplan/` directory
+with a `.yaml` extension (e.g. `/etc/netplan/config.yaml`), then run
+`sudo netplan apply`. This command parses and applies the configuration to the
+system. Configuration written to disk under `/etc/netplan/` persists between
+reboots.
+
+For each of the example below, use the `renderer` that applies to your scenario.
+For example, for Ubuntu Desktop the `renderer` is usually `NetworkManager`,
+and `networkd` for Ubuntu Server.
+
+Also, see [/examples](https://github.com/canonical/netplan/tree/main/examples)
+on GitHub.
+
 ```{toctree}
+:maxdepth: 1
+
 examples
 dbus-config
 netplan-everywhere
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,6 +1,9 @@
 import sys
-
 import datetime
+import atexit
+import fileinput
+import os
+import re
 
 # Custom configuration for the Sphinx documentation builder.
 # All configuration specific to your project should be done in this file.
@@ -373,3 +376,31 @@ html_js_files = ['header-nav.js']
 if 'github_issues' in html_context and html_context['github_issues'] and not disable_feedback_button:
     html_js_files.append('github_issue_links.js')
 html_js_files.extend(custom_html_js_files)
+
+
+# Clip the "How to" string from how-to titles in the left navigation sidebar
+#     and capitalizes the first letter of the remaining title.
+def clip_howto():
+    pattern = re.compile(r'(toctree-l2.*>)How to ([a-zA-Z])')
+    # Check if the build is running on ReadTheDocs
+    if os.environ.get('READTHEDOCS') == 'True':
+        output_dir = os.environ.get('READTHEDOCS_OUTPUT') + '/html'
+    else:
+        output_dir = '_build'
+
+    print("Clipping 'How to' from the Table of Contents...")
+
+    for root, dirs, files in os.walk(output_dir):
+        for file in files:
+            if file.endswith('.html'):
+                file_path = os.path.join(root, file)
+
+                with fileinput.FileInput(file_path, inplace=True) as f:
+                    for line in f:
+                        line = pattern.sub(lambda match: match.group(1) + match.group(2).upper(), line)
+                        print(line, end='')
+
+
+# Register the clip_howto function to be called on exit
+atexit.register(clip_howto)
+# End how-to clipping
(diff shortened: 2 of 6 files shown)


doc/Makefile
```diff
@@ -58,6 +58,8 @@ run: install
 # Doesn't depend on $(BUILDDIR) to rebuild properly at every run.
 html: install
 	. $(VENV); $(SPHINXBUILD) -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" -w .sphinx/warnings.txt $(SPHINXOPTS)
+	@echo "\nClipping 'How to' from Table of Contents:"
+	sed -i '/toctree-l2.*>How to / s/How to \(.\)/\U\1/' _build/**/*.html
```
daniloegea, line 62: thought: are the repetitive "How to"s really required in the titles? I mean, the section itself is called "How-to guides"... if we could remove them this hack wouldn't be necessary. Anyway, apparently there is no easy way to adjust the title in the sidebar and this string replacement seems to work just fine.

daniloegea (commented):
Thanks, it looks good to me (only check the last commit).

doc/howto.md
```diff
@@ -1,6 +1,27 @@
...
+For each of the example below, use the `renderer` that applies to your scenario.
+For example, for Ubuntu Desktop the `renderer` is usually `NetworkManager`,
+and `networkd` for Ubuntu Server.
+
+Also, see [/examples](https://github.com/canonical/netplan/tree/main/examples)
+on GitHub.
+
 ```{toctree}
+:maxdepth: 1
+:name: cipisek
```
slyon, line 23: question: Is this "name" value referenced somewhere? Do we need it? Depending on source language I translate it from, it could be insulting :D So if anything, I'd prefer an english term.

slyon (changes requested):
Thanks, the changes are looking mostly good to me. But we hit some build failures. Locally, `make run` hits an endless loop and I see a failure on the readthedocs CI, which might be related to `doc/howtos.sh` being incomplete. (Some of which might already be fixed in #429)

I also left some inline comments.

doc/howto.md
rkratky: OMG, I'm an idiot! :) This was for testing purposes only. I wouldn't dream of leaving it in the content intentionally. Thank you for catching it!

doc/Makefile
rkratky: > I mean, the section itself is called "How-to guides"... if we could remove them this hack wouldn't be necessary.

True, but to keep it copnsistent with other Diataxis docs, this is how we're advised to do it: have "How to" in the heading but not in navigation.

> apparently there is no easy way to adjust the title in the sidebar and this string replacement seems to work just fine

I'm not happy with this hack. It's brittle and asks for trouble down the road. But modifying the navig. titles only is not possible with Markdown sources :(

rkratky:
@slyon This is finally building successfully on RTD, so I believe it's good to merge (after #429).

> Thanks, the changes are looking mostly good to me. But we hit some build failures. Locally, `make run` hits an endless loop and I see a failure on the readthedocs CI, which might be related to `doc/howtos.sh` being incomplete. (Some of which might already be fixed in #429)
> 
> I also left some inline comments.

Thanks. I was aware of those problems. Should be fixed now.

slyon:
@rkratky Could you please rebase this branch on top of `main`, in order to resolve the merge conflicts?

rkratky:
> @rkratky Could you please rebase this branch on top of `main`, in order to resolve the merge conflicts?

Done. Sorry for the mess :)

doc/Makefile
```diff
@@ -59,6 +59,8 @@ run: install
 # Doesn't depend on $(BUILDDIR) to rebuild properly at every run.
 html: install
 	. $(VENV); $(SPHINXBUILD) -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" -w .sphinx/warnings.txt $(SPHINXOPTS)
+	@echo -e "\nClipping 'How to' from Table of Contents:"
+	@sed -i '/toctree-l2.*>How to / s/How to \(.\)/\U\1/' _build/**/*.html _build/*.html
```
slyon, line 63: This seems duplicated with `howto.sh`, can we re-use the script here, so we don't define the same logic in two places?

slyon (commented):
I guess we could squash both commits into "[Improve how-to docs](https://github.com/canonical/netplan/pull/430/commits/4a1d1de2886242b7a3cb5d43160d87d8873a3f13)" (or I can do a squash merge). The first commit seems to be a left-over from the rebase.

Also see my inline comments for a few final nitpicks before we can merge this.

doc/Makefile
rkratky: > The first commit seems to be a left-over from the rebase.

Right. I'll squash them.

> It's still here :)

Aargh. It came back wirth the rebase.

> This seems duplicated with howto.sh, can we re-use the script here, so we don't define the same logic in two places?

OK

rkratky: > This seems duplicated with `howto.sh`, can we re-use the script here, so we don't define the same logic in two places?

I replaced the `sed` command (and the `howtos.sh` script) with a bit of Python code added to `conf.py`. That way, it's called as a part of the Sphinx build both on local and on RTD, and it also works when `make run` (.e. `sphinx-autobuild`) is used.

doc/howto.md
rkratky: Fixed :)

slyon (approved):
Thanks a lot for cleaning it all up! :heart:
LGTM.

PR #431, opened 2024-01-05
ovs: disable StartLimitBurst in the ovs-cleanup service (LP: #2047827)

daniloegea:
With systemd 255, the ovs integration test is hitting this limit and the test fails due to that. As it's a oneshot service, it should be OK to disable this limit.

## Description

The alternative solution is to add a delay between each OVS integration test. In my tests, the required delay to get all the tests passing was 3 seconds. This change is to avoid increasing the time spent during tests.

LP bug: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2047827

What do you think, Lukas?

## Checklist


slyon (approved):
LGTM

PR #432, opened 2024-01-05
ctests: stop including C files in the test files

daniloegea:
With these changes, a copy of libnetplan without static functions will be build. All the test files will be linked against this separate library. With that, there is no need to include the C files in the test files anymore.

To make this possible, all the static keywords in function declarations were replaced with a macro that will be defined at compilation time.

## Description

My goal is to make #425 easier to be implemented. Also, refactoring the C code is less likely to break the build of test files...

## Checklist

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -31,7 +31,7 @@
  * Loading and error handling
  ****************************************************/
 
-static void
+STATIC void
 write_error_marker(GString *message, int column)
 {
     int i;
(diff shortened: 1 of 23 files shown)


slyon:
This approach sounds reasonable to me. Especially with the default `STATIC -> static` fallback in place, as long as UNITTESTS is not defined.

CC @fnordahl WDYT? Would that work for you?

daniloegea:
Rebased on top of the last batch of commits.

slyon (approved):
This is a nice refactoring that makes testing easier. It doesn't affect the production code.

LGTM, let's get this merged!

PR #434, opened 2024-01-18
ABI: Refactoring for libnetplan.so.1

slyon:
## Description
Please review commit-by-commit.

Bumping libnetplan's SOVERSION to 1, indicating breaking changes. Also, bumping Netplan's version to "1.0" already, in preparation for the next release.

I've dropped some more legacy code and cleaned up the existing public & internal APIs, making sure we do not export anything that is either prefixed with `_netplan_` (internal API) or `netplan_` (public API).

### Note
Feel free to ignore the last (big) commit in "abi-compat", as that is auto-generated (https://github.com/canonical/netplan/pull/434/commits/e6b3c7d143cd91f7e88d69e571ed63211f58f694)

### Known issues
One thing I noticed already is, that it is building against `noble-release` using systemd v253 (instead of v255 from -proposed), therefore getting the wrong systemd-generator path from `pkgconf`.

FR-3760

## Checklist


src/meson.build
```diff
@@ -18,7 +18,7 @@ libnetplan = library(
     gnu_symbol_visibility: 'hidden',
     dependencies: [glib, gio, yaml, uuid],
     include_directories: inc,
-    soversion: '0.0',
+    soversion: 1,
     install: true)
 
 libnetplan_testing = library(
```
daniloegea, line 24: nitpick (non-blocking): `soversion` should be changed in `libnetplan_testing` as well for consistency.

.github/workflows/build-abi.yml
```diff
@@ -44,4 +44,4 @@ jobs:
       # Abigail ABI checker
       - name: Check ABI compatibility
         run: |
-          abidiff abi-compat/jammy_0.107.xml _build/src/libnetplan.so.0.0 --headers-dir2 include/ --header-file2 src/abi.h --suppressions abi-compat/suppressions.abignore --no-added-syms
+          abidiff abi-compat/jammy_1.0.xml _build/src/libnetplan.so.1 --headers-dir2 include/ --header-file2 src/abi.h --suppressions abi-compat/suppressions.abignore --no-added-syms
```
daniloegea, line 47: question: do we have plans to update jammy to netplan 1.0? I'm just confused with the name of the file...

daniloegea (approved):
lgtm! I just left 2 non-blocking comments.

daniloegea:
Oh, about the Network Manager tests. I'm not sure why they are failing :(  I tried it locally with netplan.io from -proposed and systemd from release (253) they just work... not sure if it's related to the usrmerge stuff...

daniloegea:
wait... we got rid of the old .0 shared library, Network Manager needs to be rebuilt on Noble to work properly.

slyon:
> wait... we got rid of the old .0 shared library, Network Manager needs to be rebuilt on Noble to work properly.

Ahh, yes! That might be it. On the distro side this will be enforced by changing the `libnetplan1` package name. Inside the CI I need some way to work around it... maybe rebuilding NM.

PR #435, opened 2024-01-18
cli/sriov: remove unused code

daniloegea:
We are parsing all the configuration unnecessarily here. np_state is being overridden by the State instance from the config manager.

Fix a related unit test. The configuration is expected to fail but now it's failing in a different place and raising a different exception.

## Checklist


slyon (approved):
LGTM!

This was apparently a left-over from times where `configmanager.py` was not using `netplan.Parser()` and `netplan.State()` logic yet, but had its own Netplan parser, implemented in Python.

See https://github.com/canonical/netplan/pull/253#discussion_r802447389 for reference.

PR #439, opened 2024-01-25
SR-IOV improvements

daniloegea:
WIP

## Checklist


netplan_cli/cli/commands/sriov_rebind.py
```diff
@@ -46,5 +61,85 @@ def command_rebind(self):
...
+            # Mellanox specific quirks
+
+            parser = netplan.Parser()
+            parser.load_yaml_hierarchy(rootdir='/')
+            np_state = netplan.State()
+            np_state.import_parser_results(parser)
+
+            for netdef in np_state.ethernets.values():
+                if netdef.id == iface or netdef.set_name == iface:
+                    if netdef.bond_link:
```
mkalcok, line 89: I believe there can be some pitfalls in using netplan's state to determine interface's bond state. In the following situation:

* System has two non-bonded interfaces
* User edits netplan config and sets up a bonding
* User calls `netplan apply --sriov-only`

In this scenario, Netplan's state would tell you that the two interfaces are bonded because that's what defined in the config, but since the `--sriov-only` flag [tells netplan to process only sriov stuff and exit](https://github.com/canonical/netplan/blob/d55ae65657d7d0a9611c1be0f6417bff19290fe4/netplan_cli/cli/commands/apply.py#L74), the bond would not be actually created yet.

mkalcok (changes requested):
Hi @daniloegea, thanks for figuring out resolution to the multiple calls of `netplan apply --sriov-only`. I only have two suggestions here.

One is that we should check if the bond type is one of the supported types for VF LAG. According to [the NVIDIA docs](https://enterprise-support.nvidia.com/s/article/Configuring-VF-LAG-using-TC) it must be either `Active-Backup`, `Balance-XOR` or `LACP`.

Second suggestion is in the in-lined comment.

netplan_cli/cli/commands/sriov_rebind.py
daniloegea: Well, that's true. But one shouldn't run `netplan apply --sriov-only` if they also set up a new bond device in the configuration. In this case you'd need to run a plain `netplan apply`.

But the main reason why I used the netplan state instead of checking the system is that, by the time we do the check, the bond might not be configured yet by networkd. In this case the `master` directory will not be present in sysfs yet. Note that I'm not really sure if this race condition could really happen, but I feel like it safer to use the netplan state.

mkalcok: I see, I didn't really consider the possibility of that race condition. I guess both approaches have some drawbacks and it's very likely that user would just call plain `netplan apply` after the change to the config. I just wanted to point out this possibility.

daniloegea: Let's keep it open and see if @slyon has a different approach in mind :)

src/sriov.c
```diff
@@ -67,6 +69,40 @@ write_sriov_rebind_systemd_unit(GHashTable* pfs, const char* rootdir, GError** e
...
+    g_string_append(s, "\n[Service]\nType=oneshot\n");
+    g_string_append_printf(s, "ExecStart=" SBINDIR "/netplan apply --sriov-only\n");
+
+    _netplan_g_string_free_to_file(s, rootdir, path, NULL);
+
+    _netplan_safe_mkdir_p_dir(link);
+    if (symlink(path, link) < 0 && errno != EEXIST) {
+        // LCOV_EXCL_START
+        g_set_error(error, G_MARKUP_ERROR, G_MARKUP_ERROR_INVALID_CONTENT,
+                    "failed to create enablement symlink: %m\n");
```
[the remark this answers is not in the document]
daniloegea: Probably not, I copied it from the other function and didn't notice it. I will remove from the other one too.

src/sriov.c
```diff
@@ -40,6 +40,8 @@ write_sriov_rebind_systemd_unit(GHashTable* pfs, const char* rootdir, GError** e
     GString* s = g_string_new("[Unit]\n");
     g_string_append(s, "Description=(Re-)bind SR-IOV Virtual Functions to their driver\n");
     g_string_append_printf(s, "After=network.target\n");
+    g_string_append_printf(s, "Requires=netplan-sriov-apply.service\n");
+    g_string_append_printf(s, "After=netplan-sriov-apply.service\n");
```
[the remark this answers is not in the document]
daniloegea: As we discussed, After= should be enough.

netplan_cli/cli/commands/sriov_rebind.py
```diff
@@ -46,5 +61,84 @@ def command_rebind(self):
...
+
+        while counter > 0:
+            try:
+                with open(path, 'r') as f:
+                    state = f.read().strip()
+                    if state != 'active':
+                        print(f'{pci_addr} VF LAG state is not active yet, retrying in 1 second...')
+                        # Based on tests with a ConnectX-5 NIC, a single 1-second cycle was enough time to
+                        # allow the interfaces to change state.
+                        sleep(1)
```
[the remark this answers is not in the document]
daniloegea: well, 1 second was enough with the hardware I had to test. But decreasing the polling rate keeping the total time is a good idea I think.

netplan_cli/cli/commands/sriov_rebind.py
```diff
@@ -46,5 +61,84 @@ def command_rebind(self):
...
+            for netdef in np_state.ethernets.values():
+                if netdef.id == iface or netdef.set_name == iface:
+                    if netdef.bond_link:
+                        # VF LAG support. See LP: #1988018
+                        # If the PF is a member of a bond, the user might be trying to enable the
+                        # VF LAG feature.
+                        # Mellanox VF LAG requires that the LAG state reports as 'active'
+                        # *before* VFs can be bound to the driver. Performing the bind operation
+                        # before the device is ready will cause the VF LAG feature to never be enabled.
+                        print(f'{iface} - waiting for the LAG state to be \'active\'')
```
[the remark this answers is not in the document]
daniloegea: so, after failing to capture the output of `logging.debug` in the unittests, I decided to implement the logic for `netplan rebind --debug` (not `netplan --debug rebind`!). By creating a new logger I could avoid having the glib debug noise and capture it in the unit tests. I also used `--debug` by default in the systemd service so we can see in the journal what happened during the configuration.

daniloegea:
Hi Lukas, thanks for looking at it. I think it's ready for another review. Please look at all the commits again as I did some significant changes and added new commits.

netplan_cli/cli/commands/sriov_rebind.py
```diff
@@ -44,7 +66,105 @@ def command_rebind(self):
...
+            # driver to change state.
+            # Based on tests with a ConnectX-5 NIC, 1 second is enough time, so let's wait a bit more
+            # just in case. This delay will only be introduced if the PF is part of a bond.
+            sleep(FALLBACK_WAIT_TIME_SEC)
+            raise MLX5VFLAGStateNotFound
+
+        while retries > 0:
+            try:
+                if self._get_mlx5_vf_lag_state(pci_addr) != 'active':
+                    logging.debug(f'{pci_addr} VF LAG state is not active yet, retrying in 1 second...')
```
slyon: nitpick (non-blocking): We're retrying in `INTERVAL_SEC` seconds. Just drop the specific time from the log message... "VF LAG state is not active yet, retrying..."

src/sriov.c
```diff
@@ -90,6 +127,13 @@ netplan_state_finish_sriov_write(const NetplanState* np_state, const char* rootd
                     pf = def;
                 else if (def->sriov_link)
                     pf = def->sriov_link;
+
+                if (pf) {
+                    if (pf->set_name)
+                        g_hash_table_add(apply_pfs, pf->set_name);
+                    else if (!pf->has_match) /* netdef_id == interface name */
+                        g_hash_table_add(apply_pfs, pf->id);
```
slyon, line 135: nitpick: We might want to log some warning here in the "else" case where we don't have `set-name`, but still have a `match` stanza (i.e. when matching multiple interfaces). In which case it won't work.

slyon (approved):
Thanks, LGTM!

I left a few nitpicks inline, nothing urgent. I assume you tested in on the ConnectX-5 hardware we used for development, so should be fine.

Still, it would be great to get confirmation from @mkalcok that this is working as expected, using the PPA provided by Danilo.

mkalcok:
Today I tested this change with the provided PPA and it seems to work as expected. The only concern that I have, and it may be caused by something outside of Netplan's control, is the delays that I see between individual steps. Following is a snippet from `dmesg`:
```
[  126.053468] pci 0000:03:0f.5: [15b3:1018] type 00 class 0x020000
[  126.053554] pci 0000:03:0f.5: enabling Extended Tags
[  126.054920] pci 0000:03:0f.5: Adding to iommu group 218
[  126.055293] mlx5_core 0000:03:0f.5: enabling device (0000 -> 0002)
[  126.055413] mlx5_core 0000:03:0f.5: firmware version: 16.35.2000
[  126.366847] mlx5_core 0000:03:0f.5: Rate limit: 127 rates are supported, range: 0Mbps to 24414Mbps
[  126.382849] mlx5_core 0000:03:0f.5: Assigned random MAC address 72:03:0c:2f:38:0c
[  126.567068] mlx5_core 0000:03:0f.5: MLX5E: StrdRq(1) RqSz(8) StrdSz(2048) RxCqeCmprss(0 basic)
[  126.580817] mlx5_core 0000:03:0f.5 enp3s0f1v59: renamed from eth0
# ^Last VF being un-bound

[  222.821313] mlx5_core 0000:03:00.0: E-Switch: Disable: mode(LEGACY), nvfs(60), necvfs(0), active vports(61)
[  224.755201] mlx5_core 0000:03:00.0: E-Switch: Supported tc chains and prios offload
[  226.392238] mlx5_core 0000:03:00.0 enp3s0f0np0: Link down
[  226.396783] mlx5_core 0000:03:00.0 enp3s0f0np0: Dropping C-tag vlan stripping offload due to S-tag vlan
[  226.396788] mlx5_core 0000:03:00.0 enp3s0f0np0: Disabling HW_VLAN CTAG FILTERING, not supported in switchdev mode
[  232.743648] mlx5_core 0000:03:00.0 enp3s0f0np0: Link up
[  233.703233] mlx5_core 0000:03:00.0: E-Switch: Enable: mode(OFFLOADS), nvfs(60), necvfs(0), active vports(60)
# ^ PF1 switched to e-switch mode

[  320.650100] mlx5_core 0000:03:00.1: E-Switch: Disable: mode(LEGACY), nvfs(60), necvfs(0), active vports(61)
[  322.486716] mlx5_core 0000:03:00.1: E-Switch: Supported tc chains and prios offload
[  323.173694] mlx5_core 0000:03:00.1 enp3s0f1np1: Link down
[  323.179134] mlx5_core 0000:03:00.1 enp3s0f1np1: Dropping C-tag vlan stripping offload due to S-tag vlan
[  323.179138] mlx5_core 0000:03:00.1 enp3s0f1np1: Disabling HW_VLAN CTAG FILTERING, not supported in switchdev mode
[  328.352086] mlx5_core 0000:03:00.1 enp3s0f1np1: Link up
[  330.213742] mlx5_core 0000:03:00.1: E-Switch: Enable: mode(OFFLOADS), nvfs(60), necvfs(0), active vports(60)
# ^ PF2 switched to e-switch mode
```
So there's about 100 seconds between last VF getting un-bound and first PF switching to `eswitch` mode, then another ~90 seconds before second PF swithces to `eswitch` mode.
I don't see any obvious reason from the code why it should take this long, do you have any thoughts on this?

mkalcok:
I'll double-check @daniloegea I suspect that this delay is not coming from netplan. We are double checking actual connectivity via the configured bond, I'll get back to you as soon as we confirm that everything works.

I have one more question that might be bit outside of scope of this change, but is there a way to wait until netplan is done with this setup? Particularly at boot, we'll need a way to tell `nova` (or other hypervisor) to wait until netplan is finished, otherwise it will try to start VMs and bind VFs to them. This would either fail because the VFs are not ready yet, or it will screw up with VF-LAG setup because it will bind VFs that are supposed to be unbound.

slyon:
I assume we should be able to sort a system unit after the corresponding Netplan services to make that work?
```
After=netplan-sriov-apply.service
After=netplan-sriov-rebind.service
```

slyon:
Looks like we might want some `DefaultDependencies=no` changes in the newly generated services, too.

daniloegea:
I added the DefaultDependencies=no. It fixes the dependency cycles found during tests.

slyon:
The openstack team confirmed that their VMs come up with VF-LAG enabled, using this branch. So let's get it merged.

fnordahl:
We have had requests from users on Ubuntu Jammy that are interested in these improvements to resolve issues with consistent upbringing of VFs, embedded switch mode and VF LAG on their systems.

Are there any plans to bring Netplan 1.0, or to backport these fixes, to Ubuntu Jammy?

cc @narindergupta

slyon:
> Are there any plans to bring Netplan 1.0, or to backport these fixes, to Ubuntu Jammy?

Backporting the isolated fixes might be an option. Would you mind creating a bug report on the Ubuntu package about this? https://launchpad.net/ubuntu/+source/netplan.io

Our full-version backport policy is usually to support the latest Ubuntu LTS, which would be Noble in this case. Porting 1.0 back to Jammy is especially tricky due to the breaking ABI changes and SOVER bump of libnetplan0 -> libnetplan1, introducing a transition in a stable release.

PR #420, opened 2023-11-02
state/status: add support for identifying bridge/bond/vrf members

daniloegea:
If the interface is a bridge/bond/vrf, display its members. If it is a member of a bridge/bond/vrf, display the bridge/bond/vrf name.

It will also use the "Kind" property to be more specific about the interface type.

Example

```
     Online state: online
    DNS Addresses: 127.0.0.53 (stub)
       DNS Search: lxd

●  1: lo ethernet UNKNOWN/UP (unmanaged)
      MAC Address: 00:00:00:00:00:00
        Addresses: 127.0.0.1/8
                   ::1/128
                   ::1 metric 256

●  2: eth0 ethernet UP (networkd: mynic)
      MAC Address: 76:25:f1:0d:14:29 (Red Hat, Inc.)
        Addresses: 10.86.126.8/24 (dhcp)
                   fd42:bc43:e20e:8cf7:7425:f1ff:fe0d:1429/64
                   fe80::7425:f1ff:fe0d:1429/64 (link)
    DNS Addresses: 10.86.126.1
                   fe80::216:3eff:feab:beb9
       DNS Search: lxd
           Routes: default via 10.86.126.1 from 10.86.126.8 metric 100 (dhcp)
                   10.86.126.0/24 from 10.86.126.8 metric 100 (link)
                   10.86.126.1 from 10.86.126.8 metric 100 (dhcp, link)
                   fd42:bc43:e20e:8cf7::/64 metric 100 (ra)
                   fe80::/64 metric 256
                   default via fe80::216:3eff:feab:beb9 metric 100 (ra)

●  3: bond0 bond UP (networkd: bond0)
      MAC Address: 1a:b2:f0:35:da:4f
        Addresses: fe80::18b2:f0ff:fe35:da4f/64 (link)
           Routes: fe80::/64 metric 256
          Members: dm6
                   dm7

●  4: bond1 bond UP (networkd: bond1)
      MAC Address: ba:67:d9:67:44:55
        Addresses: fe80::b867:d9ff:fe67:4455/64 (link)
           Routes: fe80::/64 metric 256
          Members: dm10
                   dm8
                   dm9

●  5: bond2 bond UP (networkd: bond2)
      MAC Address: 42:f8:57:f9:b4:f6
        Addresses: fe80::40f8:57ff:fef9:b4f6/64 (link)
           Routes: fe80::/64 metric 256
          Members: dm11
                   dm12
                   dm13

●  6: br0 bridge UP (networkd: br0)
[...]

PR #440, opened 2024-02-02
Netplan status --diff

daniloegea:
## Description

This PR introduces `netplan status --diff`. It will use the `state_diff` module to output the inconsistencies in the network configuration between the system and the netplan configuration.

The diff will be displayed the same way `netplan status` already displays information. Even though it's a nice approach, plugging the `diff` display code in the existing `status` code made things a bit more complicated. I have refactored it (only after have plugged the code, sorry for that) and split the code responsible for each block of information in their own methods.

It also adds the information about bonds, bridges and VRFs introduced by #420.

## Checklist


netplan_cli/cli/commands/status.py
```diff
@@ -88,18 +88,164 @@ def pretty_print(self, data: JSON, total: int, _console_width=None) -> None:
...
+        if self.state_diff:
+            if diff := self.state_diff['interfaces'].get(ifname):
+                return diff.get('system_state', {}).get('missing_bond_link', '')
+        return ''
+
+    def _get_missing_netplan_bond_link(self, ifname) -> str:
+        if self.state_diff:
+            if diff := self.state_diff['interfaces'].get(ifname):
+                return diff.get('netplan_state', {}).get('missing_bond_link', '')
+        return ''
```
slyon, line 194: nitpick (non-blocking): This (and surrounding) methods feel like lots of duplicated code, doing almost the same thing for `system_state` and `netplan_state`, I wonder if we could pass the relevant `diff.get('NETPLAN/SYSTEM_state')` data as an input variable, to use the same method for both cases?

netplan_cli/cli/commands/status.py
```diff
@@ -88,18 +88,164 @@ def pretty_print(self, data: JSON, total: int, _console_width=None) -> None:
...
+
+    def _get_missing_system_bridge_link(self, ifname) -> str:
+        if self.state_diff:
+            if diff := self.state_diff['interfaces'].get(ifname):
+                return diff.get('system_state', {}).get('missing_bridge_link', '')
+        return ''
```
slyon, line 200: question: In `_get_missing_system_macaddress` we're returning `None` for an undefined string, here (and in some other, similar methods) it's the empty string.. Why the difference?

netplan_cli/cli/commands/status.py
```diff
@@ -109,184 +255,613 @@ def pretty_print(self, data: JSON, total: int, _console_width=None) -> None:
...
+        if self.diff:
+            self._display_missing_interfaces()
 
         hidden = total - len(interfaces)
         if (hidden > 0):
-            pprint('{} inactive interfaces hidden. Use "--all" to show all.'.format(hidden))
+            self.pprint('{} inactive interfaces hidden. Use "--all" to show all.'.format(hidden))
+
+        if (self.diff and not self.diff_only):
+            self.pprint('Use "--diff-only" to omit the information that is consistent between the system and Netplan.')
```
slyon, line 833: nitpick: This looks a bit odd, due to the `'netplan.str': 'grey62',` definition in the diff Theme. Can we make this line to use the usual theme?

tests/cli/test_status.py
```diff
@@ -503,3 +512,119 @@ def test_call_cli_networkd_masked(self, is_masked_mock, is_active_mock):
...
+        rd_mock.return_value = (None, None)
+        resolvconf_mock.return_value = {'addresses': [], 'search': [], 'mode': None}
+        online_mock.return_value = False
+        state_init_mock.return_value = None
+        diff_state_init_mock.return_value = None
+        diff_state_get_diff_mock.return_value = {}
+        state = SystemConfigState()
+        networkd_mock.return_value = state.process_networkd(NETWORKD)
+        out = self._call(['--diff-only'])
+        self.assertEqual('', out)
```
slyon, line 572: question (non-blocking): Should we put some message in case where there isn't any diff to show?

tests/cli/test_status_diff.py
```diff
@@ -0,0 +1,1188 @@
...
+            status.state_diff = state_diff
+            status.pretty_print(input_data, 2, _console_width=130)
+            out = f.getvalue()
+            self.assertEqual(out, expected)
+
+        # Same test for --diff-only
+
+        expected = '''+ ●  1: lo ethernet UNKNOWN/UP (unmanaged)
+
+'''
```
slyon, line 72: question (non-blocking): Should we trim unused whitespace in the output?

tests/cli/test_status_diff.py
```diff
@@ -0,0 +1,1188 @@
...
+
+        expected = '''+ ●  1: lo ethernet UNKNOWN/UP (unmanaged)
+          MAC Address: 00:00:00:00:00:00
+            Addresses: 127.0.0.1/8
+                       ::1/128
+               Routes: ::1 metric 256
+
+  ●  2: enp5s0 ethernet UP (networkd: enp5s0)
++         MAC Address: 00:16:3e:71:d0:1f (Red Hat, Inc.)
+-                      aa:bb:cc:dd:ee:ff (Red Hat, Inc.)
```
slyon, line 163: parise: Thanks for adding the +/- signs, so we can test this code even without the color theming.

slyon (commented):
Thank you very much Danilo, I very much like the looks of this!
And wow! Crafting all the data for `test_status_diff.py` must have been lots of busywork, but it's worth the effort. Thanks for going the extra mile!

Should this PR become a squash merge (due to the refactoring happening only after the previous commits)?

I left some non-critical inline-comments, which I'd like to see discussed before merging it.
Also, I think we should update the manpage/docs around `status --diff[-only]`, but this could (should?) happen in a separate PR, so we can more easily get Robert's feedback, too.

A final (non-blocking) thought: We should consider using `ip addr` information to decide if an IP address got assigned through DHCP, looking at the `dynamic` flag. As DHCPv6 IP addresses are mostly not correctly detected on my system. Probably also in a separate PR, as this one is already pretty big.
```
$ ip -d -j addr show dev lan0 | jq
[...]
    "addr_info": [
      {
        "family": "inet",
        "local": "192.168.178.141",
        "prefixlen": 24,
        [...]
        "scope": "global",
        "dynamic": true,
      },
      {
        "family": "inet6",
        "local": "2001:16b8:bdba:b100:fae4:3bff:fe2d:3bb7",
        "prefixlen": 64,
        [...]
        "scope": "global",
        "dynamic": true,
      }
    ]
```

netplan_cli/cli/utils.py
```diff
@@ -213,6 +213,20 @@ def find_matching_iface(interfaces: list, netdef):
...
+def route_table_lookup() -> dict:
+    lookup_table = {}
+    try:
+        with open('/etc/iproute2/rt_tables', 'r') as rt_tables:
+            for line in rt_tables:
+                split_line = line.split()
+                if len(split_line) == 2 and split_line[0].isnumeric():
+                    lookup_table[int(split_line[0])] = split_line[1]
+    except Exception:
+        return {}
```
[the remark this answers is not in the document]
daniloegea: Ok, but maybe only in debug mode. It's not really critical and I don't want the message to appear every time you call `netplan status`.

netplan_cli/cli/commands/status.py
```diff
@@ -61,22 +65,18 @@ def run(self):
                                  help='Show extra information')
         self.parser.add_argument('-f', '--format', default='tabular',
                                  help='Output in machine readable `json` or `yaml` format')
+        self.parser.add_argument('--diff', action='store_true',
+                                 help='Show the differences between the system\'s and netplan\'s states')
+        self.parser.add_argument('--diff-only', action='store_true',
```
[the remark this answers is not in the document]
daniloegea: hmm I don't know, every time I see `-d` I automatically assume it's for `--debug`. Maybe we should avoid that.

netplan_cli/cli/commands/status.py
```diff
@@ -109,184 +255,613 @@ def pretty_print(self, data: JSON, total: int, _console_width=None) -> None:
...
+        if self.diff:
+            netplan_state = NetplanConfigState(rootdir=self.root_dir)
+            diff_state = NetplanDiffState(system_state, netplan_state)
+
+            self.state_diff = diff_state.get_diff(self.ifname)
+
+            if output_format == 'json':
+                print(json.dumps(self.state_diff, cls=DiffJSONEncoder))
+                return
+            elif output_format == 'yaml':
```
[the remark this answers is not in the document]
daniloegea: That's a good point. I fixed that for the normal output for missed for json and yaml. I'll look at that.

netplan_cli/cli/commands/status.py
daniloegea: No reason, I don't why I did that :)

netplan_cli/cli/commands/status.py
daniloegea: To use the other theme I'd need to create a new Console() so I guess I'll just add [yellow] around the --diff-only. I think it's the only part being affected by the theme...

tests/cli/test_status_diff.py
daniloegea: I suppose it would be nice. It will require checking if the interface we are displaying is or is not the last one. I'll do that

daniloegea:
Thank you, Lukas. I think it's ready for another round of code review. I believe I addressed most of your comments, I think the main one being the linebreak at the end of the output and the code duplication.

Regarding squashing all the commits, I don't know. While there is a little mess in the middle, there are multiple commits that are well separated. Maybe all the ones names as `status` and `status/diff` (and the tests I guess) could be squashed together, but not the rest.

tests/cli/test_status.py
slyon: Agreed. That also seems to be what `git diff` does.

slyon (approved):
Thank you for the fixes! This is looking really good. I like using the new resulting sub-command a lot!

Could you please rebase and maybe clean up the git history a bit, before we merge it?

Besides that I left have some tiny comments
* Unknown device type: "other": warnings on my system (probably unrelated, but easy to fix)
* `_get_missing_property_str` can return `None` or the empty string
* Should we drop the `MISSING` state, to unclutter the output?
* Should we display unknown interfaces types as `other` instead of `unknown` to stay in line with `status.py`?

I'll push some commits on top of your branch to fix the mentioned issues. Feel free to integrate them into your rebase or revert/drop them with a comment.

netplan_cli/cli/commands/status.py
```diff
@@ -121,113 +121,83 @@ def _is_interface_missing_in_netplan(self, ifname) -> bool:
...
         return []
 
-    def _get_missing_system_nameservers(self, ifname) -> list[str]:
+    def _get_missing_property_str(self, ifname: str, state: str, property: str) -> str:
         if self.state_diff:
             if diff := self.state_diff['interfaces'].get(ifname):
-                return diff.get('system_state', {}).get('missing_nameservers_addresses', [])
-        return []
+                return diff.get(state, {}).get(property)
+        return ''
```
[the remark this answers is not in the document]
daniloegea: I think it's better to fix the .get() and return an empty string by default, just for consistency with the return type hint (as None is not a str)

netplan_cli/cli/commands/status.py
```diff
@@ -109,184 +225,625 @@ def pretty_print(self, data: JSON, total: int, _console_width=None) -> None:
...
+            if self.ifname and self.ifname != interface:
+                continue
+
+            self.pprint('{sign} [{col}]● {idx:>2}  {name} {type} {state}[/{col}]'.format(
+                sign=sign,
+                col='red',
+                idx='',
+                name=interface,
+                type=properties.get('type'),
+                state='MISSING',
```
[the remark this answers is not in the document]
daniloegea: that makes sense, I'll drop it.

netplan_cli/cli/commands/status.py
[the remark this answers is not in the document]
slyon: wfm.

slyon (approved):
Thanks a lot. That works for me!

PR #444, opened 2024-02-27
Netplan status --diff refactoring

daniloegea:
A few improvements for `netplan status --diff`:

- route tables name to number (and vice-versa) mapping: take into account that the files from `/etc/iproute2/` were moved to `/usr/share` in a recent release of iproute2. Fix unit tests to not open the real file.
- when comparing MAC addresses, take into account the new options added to netplan recently. For now, we will not analyze the MAC address if an option such as `random` is used in the YAML.
- expose the `link-local` property via a new API.
- improve how link local addresses and routes are analyzed and take into account the `link-local` property. Note that currently it will not check if the system interface should have a link local address (in other words, it will not show a diff if `link-local` is enabled but the address is not present in the system). It will only check if the address is present and check the `link-local` property to see if the address is expected to be there.
- Update the `netplan-status` manpage and autocomplete files.

Example of link-local changes:
In this case, link-local has the default configuration, which enables it only for IPv6, but the IPv4 link local address is also present so it will show up as a diff.

```
  ● 16: dummy123 dummy-device UNKNOWN/UP (NetworkManager: dummy123)
          MAC Address: 06:e7:f4:6e:c9:17
+           Addresses: 169.254.65.85/16 (link)
                       fe80::4e7:f4ff:fe6e:c917/64 (link)
+              Routes: 169.254.0.0/16 from 169.254.65.85 metric 550 (link)
                       224.0.0.0/4 metric 550 (static, link)
                       fe80::/64 metric 256

```

## Checklist


daniloegea:
Hi @rkratky, may I ask you to review the last commit of this PR, please? It's updating the `netplan-status` manual page. Thanks!

doc/netplan-status.md
```diff
@@ -34,6 +34,23 @@ Currently, **`netplan status`** depends on `systemd-networkd` as a source of dat
 `-a`, `--all`
 :   Show all interface data including inactive.
 
+`--diff`
+:   Analyze and display differences between the current system configuration and interface definitions present in the YAML files.
```
slyon, line 38: nitpick (non-blocking):
```suggestion
:   Analyze and display differences between the current system configuration and network definitions present in the YAML files.
```

doc/netplan-status.md
```diff
@@ -34,6 +34,23 @@ Currently, **`netplan status`** depends on `systemd-networkd` as a source of dat
...
+    The configuration analyzed includes IP addresses, routes, MAC addresses, DNS addresses and search domains and missing network interfaces.
```
slyon, line 39: nitpick (non-blocking):
```suggestion
    The configuration analyzed includes IP addresses, routes, MAC addresses, DNS addresses, search domains and missing network interfaces.
```

python-cffi/netplan/netdef.py
```diff
@@ -50,6 +50,15 @@ def dhcp4(self) -> bool:
...
         return bool(lib.netplan_netdef_get_dhcp6(self._ptr))
 
+    @property
+    def link_local(self) -> list:
+        linklocal = []
+        if bool(lib.netplan_netdef_get_link_local_ipv4(self._ptr)):
+            linklocal.append('ipv4')
+        if bool(lib.netplan_netdef_get_link_local_ipv6(self._ptr)):
+            linklocal.append('ipv6')
+        return linklocal
```
slyon, line 60: praise: I link that design, which makes it very similar to the corresponding YAML stanza!

slyon (changes requested):
I like the changes overall. Left some minor inline comments.

~When testing the changes, it seems like IP-address diffing is broken after the link-local changes, though?~
-> Sorry, configuration error on my end.

slyon (approved):
LGTM with the configuration error resolved.

PR #449, opened 2024-03-26
apply: make sure that networkd is restarted when needed

alfonsosanchezbeato:
This PR fixes these two problems:

* ~It restarts networkd instead of reload/reconfigure
Due to systemd-networkd bugs, reload/reconfigure was not working in
some cases. Instead, restart the service if the configuration has
changed (LP#2058976).~
- It compares full configuration to decide to restart or not networkd.
Copy the current networkd configuration and then call generate so we
know more accurately if systemd-networkd needs to be restarted.

### DIFF
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -272,20 +272,15 @@ class NetplanApply(utils.NetplanCommand):
             # exclude the special 'netplan-ovs-cleanup.service' unit
             netplan_ovs = [os.path.basename(f) for f in glob.glob('/run/systemd/system/*.wants/netplan-ovs-*.service')
                            if not f.endswith('/' + OVS_CLEANUP_SERVICE)]
-            # Run 'systemctl (re)start' command synchronously, to avoid race conditions
+            # Run 'systemctl start' command synchronously, to avoid race conditions
             # with 'oneshot' systemd service units, e.g. netplan-ovs-*.service.
-            #
-            # In the past, calls to networkctl_reload/networkctl_reconfigure
-            # were tried, but due maybe to systemd-networkd bugs, they were not
-            # working as expected: the interface was getting initially the
-            # expected state, but after some minutes it ignored the state of
-            # /run/systemd/network/ and used the configuration read when the
-            # service was started. This happened in the case of removed
-            # .network files. Looking at files in
-            # /run/systemd/netif/links/<num>, the internal state seemed to keep
-            # removed files in NETWORK_FILE= (see LP#2058976).
-            logging.info('Restarting systemd-networkd.service')
-            utils.systemctl('restart', ['systemd-networkd.service'], sync=True)
+            try:
+                utils.networkctl_reload()
+                utils.networkctl_reconfigure(utils.networkd_interfaces())
+            except subprocess.CalledProcessError:
+                # (re-)start systemd-networkd if it is not running, yet
+                logging.warning('Falling back to a hard restart of systemd-networkd.service')
+                utils.systemctl('restart', ['systemd-networkd.service'], sync=True)
             # 1st: execute OVS cleanup, to avoid races while applying OVS config
             utils.systemctl('start', [OVS_CLEANUP_SERVICE], sync=True)
             # 2nd: start all other services
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -25,6 +25,8 @@ import sys
 import glob
 import subprocess
 import shutil
+import tempfile
+import filecmp
 import netifaces
 import time
 
@@ -109,7 +111,6 @@ class NetplanApply(utils.NetplanCommand):
                 return
 
         ovs_cleanup_service = '/run/systemd/system/netplan-ovs-cleanup.service'
-        old_files_networkd = bool(glob.glob('/run/systemd/network/*netplan-*'))
         old_ovs_glob = glob.glob('/run/systemd/system/netplan-ovs-*')
         # Ignore netplan-ovs-cleanup.service, as it can always be there
         if ovs_cleanup_service in old_ovs_glob:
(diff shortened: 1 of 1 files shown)


slyon:
> alfonsosanchezbeato: to be clear, the networkctl reload/reconfigure commands seem to work, but after 6-8 minutes the state goes back to the one expressed by the old .network files

daniloegea:
Do you know how safe a networkd restart is? I mean, I'm trying to determine what kind of change in behavior it would introduce from the point of view of the user. I'm concerned it would cause this kind of problem: https://discuss.linuxcontainers.org/t/systemd-update-networkd-restart-breaks-routed-containers/9979

slyon:
This is going backwards wrt. https://github.com/canonical/netplan/pull/200

alfonsosanchezbeato:
> This is going backwards wrt. #200

Right, as it looks like networkd is not really respecting the reload/reconfigure commands, at least on 20.04.

alfonsosanchezbeato:
> Do you know how safe a networkd restart is? I mean, I'm trying to determine what kind of change in behavior it would introduce from the point of view of the user. I'm concerned it would cause this kind of problem: https://discuss.linuxcontainers.org/t/systemd-update-networkd-restart-breaks-routed-containers/9979

Restarting it should be in theory fine, this is how it was being done in the past. The issue you shared seemed to happen only on 18.04 I think.

alfonsosanchezbeato:
I have tested this through different UC versions, and I see:
- 20.04 systemd-networkd has the bug
- 22.04 systemd-networkd has the bug
- 24.04  systemd-networkd has apparently fixed the bug

Therefore, it would make sense to backport b9ef830add2449ae00174113c916db6b818aea13 to netplan in 20/22, and remove it from this PR, but keep 8dabb54f0a2d0e9a46e39e81644f88705b023dd7. Are there branches where the commit could be proposed for 20/22?

An alternative could be to find out the patch fixing the issue in systemd-networkd and SRU to 20/22, but being systemd it might take longer than doing an SRU of netplan, I think.

slyon:
> An alternative could be to find out the patch fixing the issue in systemd-networkd and SRU to 20/22, but being systemd it might take longer than doing an SRU of netplan, I think.

Do we have a clean reproducer? I could bisect systemd to find the commit fixing the behaviour. I'd prefer fixing it properly in systemd that backporting a workaround.

alfonsosanchezbeato:
> Thanks a lot for your continued investigation @alfonsosanchezbeato !
> 
> > An alternative could be to find out the patch fixing the issue in systemd-networkd and SRU to 20/22, but being systemd it might take longer than doing an SRU of netplan, I think.
> 
> Do we have a clean reproducer? I could bisect systemd to find the commit fixing the behaviour. I'd prefer fixing it properly in systemd that backporting a workaround.

Create a UC22 image, then install network-manager snap, then run `sudo udevadm trigger`. If the `networkctl` shows something like:
```
IDX LINK TYPE     OPERATIONAL SETUP    
  1 lo   loopback carrier     unmanaged
  2 ens3 ether    routable    failed   

2 links listed.
```
then you have the bug (if not, the interfaces should be still `unmanaged`).

slyon:
We had/have some similar quirks for Netplan in Focal already:
* 0.103: n/+git/ubuntu/tree/debian/patches/0002-Revert-cli-apply-reload-reconfigure-networkd-instead.patch?h=ubuntu/0.103-0ubuntu5_20.04.6
* 0.104: https://git.launchpad.net/~ubuntu-core-dev/netplan/+git/ubuntu/tree/debian/patches/0006-cli-apply-give-some-extra-time-for-networkctl-reload.patch?h=ubuntu/0.104-0ubuntu2_20.04.2

So I guess I'd be fine SRUing it for Focal/UC20 in Netplan. While for Jammy/UC22 I'd still like to find the proper fix in systemd-networkd.

slyon:
> Create a UC22 image, then install network-manager snap, then run `sudo udevadm trigger`. If the `networkctl` command shows something like:

I wonder if there is any reproducer to show this issue on Ubuntu classic?

daniloegea:
An alternative (instead of changing the current default behavior of `apply`) would be to add a new parameter to `netplan apply` (something like `--force-backend-restart`) to enable consumers of Netplan to force the restart of the backend in cases where the backend doesn't behave as expected or introduces a bug.

slyon (approved):
> Therefore, it would make sense to backport https://github.com/canonical/netplan/commit/b9ef830add2449ae00174113c916db6b818aea13 to netplan in 20/22, and remove it from this PR, but keep https://github.com/canonical/netplan/commit/8dabb54f0a2d0e9a46e39e81644f88705b023dd7.

So I've reverted the first commit in this PR (https://github.com/canonical/netplan/pull/449/commits/6874d4a765a37d1babdb8e549ac4d17e43eb114a), but kept the "apply: restart networkd instead of reload/reconfigure" commit as suggested (https://github.com/canonical/netplan/pull/449/commits/ac0d6154439df766e2e24dee3b3966182b95202e).

Additionally, I've added a small fix that was failing autopkgtests when we have no networkd configuration at all in `/run/systemd/networkd/` (directory doesn't exist) before or after the generator run.

With the workaround moved to https://github.com/snapcore/core-base/pull/214 I think this PR (just the cleaner networkd-config comparison logic) should now be good for a (squash) merge!

@daniloegea WDYT? (CC @alfonsosanchezbeato)

slyon:
Rebased.

daniloegea (approved):
I think it looks fine.

So, effectively now we are going to continue to **reload** (not restart) networkd only when we detect changes in the generated configuration. It's sounds like a nice improvement.

The problem is that it doesn't fix the issue for UC and now they need to maintain a patch ([PR#214](https://github.com/snapcore/core-base/pull/214)) that will force the restart.

Wouldn't it help them if we introduce a way to force-restart the backends via a CLI parameter?

alfonsosanchezbeato:
> > Therefore, it would make sense to backport [b9ef830](https://github.com/canonical/netplan/commit/b9ef830add2449ae00174113c916db6b818aea13) to netplan in 20/22, and remove it from this PR, but keep [8dabb54](https://github.com/canonical/netplan/commit/8dabb54f0a2d0e9a46e39e81644f88705b023dd7).
> 
> So I've reverted the first commit in this PR ([6874d4a](https://github.com/canonical/netplan/commit/6874d4a765a37d1babdb8e549ac4d17e43eb114a)), but kept the "apply: restart networkd instead of reload/reconfigure" commit as suggested ([ac0d615](https://github.com/canonical/netplan/commit/ac0d6154439df766e2e24dee3b3966182b95202e)).
> 
> Additionally, I've added a small fix that was failing autopkgtests when we have no networkd configuration at all in `/run/systemd/networkd/` (directory doesn't exist) before or after the generator run.
> 
> With the workaround moved to [snapcore/core-base#214](https://github.com/snapcore/core-base/pull/214) I think this PR (just the cleaner networkd-config comparison logic) should now be good for a (squash) merge!
> 
> @daniloegea WDYT? (CC @alfonsosanchezbeato)

This is fine by me too, thank you!

PR #450, opened 2024-04-03
parse: don't remove datalist items during iteration

daniloegea:
The glib documentation says that it should be safe to remove items from inside the function called by g_datalist_foreach but I've found some crashes related to it where the last element was returning NULL.

This is one of the instances that causes the crash:

passthrough:
  Utf14: sit cupidatat aliquip proident ut
  reprehenderit6e2: 16408719.105799645
  connection.type: ullamco officia

Message from glib:
GLib-CRITICAL **: 10:16:09.406: g_strsplit: assertion 'string != NULL' failed

With this change the bad keys will be stored in an GArray and removed from the datalist outside the foreach.

## Description
I was testing my https://github.com/daniloegea/netplan-fuzz and stumbled upon this issue.

Error from glib:
```
(generate:272642): GLib-CRITICAL **: 10:16:09.406: g_strsplit: assertion 'string != NULL' failed
```

Note this case, the last warning mentions a key called `GParamGType` but it is not part of the keys. That means the pointer it's working with at this moment is not what we expect due to inconsistencies in the data structure. 

```
** (generate:281561): WARNING **: 11:16:21.688: NetworkManager: passthrough key 'Utf14' format is invalid, should be 'group.key'.
processing key reprehenderit6e2

** (generate:281561): WARNING **: 11:16:21.688: NetworkManager: passthrough key 'reprehenderit6e2' format is invalid, should be 'group.key'.
processing key GParamGType

** (generate:281561): WARNING **: 11:16:21.688: NetworkManager: passthrough key 'GParamGType' format is invalid, should be 'group.key'.
/tmp/fakeroot/etc/netplan/asd.yaml: Error in network definition: 9a9A0-ZA0Z: virtual-ethernet missing 'peer' property
```

Full YAML what found the problem.
```yaml
network:
  renderer: NetworkManager
  version: 2
  virtual-ethernets:
    9a9A0-ZA0Z:
      dhcp6: false
      dhcp-identifier: mac
      ipv6-privacy: false
      networkmanager:
        uuid: et magna occaecat laborum deserunt
        name: officia et id
        passthrough:
          Utf14: sit cupidatat aliquip proident ut
          reprehenderit6e2: 16408719.105799645
          connection.type: ullamco officia
      renderer: NetworkManager
    renderer: networkd
```

## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -809,12 +809,12 @@ handle_netdef_backend_settings_str(NetplanParser* npp, yaml_node_t* node, const 
 STATIC void
 validate_kf_group_key(GQuark key_id, __unused gpointer value, gpointer user_data)
 {
-    GData** list = user_data;
+    GArray* bad_keys = user_data;
     const gchar* key = g_quark_to_string(key_id);
     gchar** group_key = g_strsplit(key, ".", -1);
     if (g_strv_length(group_key) < 2) {
         g_warning("NetworkManager: passthrough key '%s' format is invalid, should be 'group.key'.", key);
-        g_datalist_id_remove_data(list, key_id);
+        g_array_append_val(bad_keys, key_id);
     }
     g_strfreev(group_key);
 }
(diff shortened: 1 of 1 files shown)


slyon (approved):
LGTM!

PR #454, opened 2024-04-09
[WIP] sriov: accept setting the eswitch mode without VFs

daniloegea:
The embedded switch mode can be set even if the interface doesn't have any virtual functions. This is a requisite to use the PF with only scalable functions for example.

With this change, the presence of the 'embedded-switch-mode' will be enough to identify the interface as a PF and emit the systemd services required to apply the configuration.

The changes required by the CLI will be done is the next commit.

## Checklist


netplan_cli/cli/sriov.py
```diff
@@ -212,37 +211,47 @@ def unbind_vfs(vfs: typing.Iterable[PCIDevice], driver) -> typing.Iterable[PCIDe
...
+            for interface in interfaces:
+                if not _interface_matches(netdef, interface):
+                    continue
+                # we have a matching PF
+                # error out if we matched more than one
+                if len(matches) > 1:
+                    raise ConfigurationError('matched more than one interface for a PF device: %s' % netdef.id)
+                matches.add(interface)
+            if matches:
+                return list(matches)[0]
```
slyon, line 248: suggestion (non-blocking): I think some of this could be replaced by the `utils.find_matching_iface()` method. Error handling cannot be clearly translated, though.

netplan_cli/cli/sriov.py
```diff
@@ -212,37 +211,47 @@ def unbind_vfs(vfs: typing.Iterable[PCIDevice], driver) -> typing.Iterable[PCIDe
...
+            # if we had a match: stanza and set-name: this means we should
+            # assume that, if found, the interface has already been
+            # renamed - use the new name
+            return set_name
+        else:
+            matches: Set[str] = set()
+            # we walk through all the system interfaces to determine if there is
+            # more than one matched interface
+            for interface in interfaces:
+                if not _interface_matches(netdef, interface):
```
slyon, line 240: question (non-blocking): Why not just call `netdef._match_interface()` directly here?

slyon (approved):
Thank you for the excelent testing story!
Also, I like the refactoring & breakdown into smaller methods a lot!

LGTM. I left some minor inline comments, it's up to you if you want to address those.

netplan_cli/cli/sriov.py
daniloegea: I don't have a strong opinion here, but I don't know it looks cleaner with `netdef._match_interface` wrapped in this method...

netplan_cli/cli/sriov.py
daniloegea: this method returns None if it either found more than one match or no matches at all. Based on the current logic I need to know if there's multiple matches. But yeah I understand there some code duplication going on here... maybe it's a task for a future refactoring :|

PR #455, opened 2024-04-11
Create individual netplan-networkd-wait-online@.service units (LP: #2060311)

slyon:
## Description
This is in favor to RequiredForOnline=yes as the behavior of upstream (pure) systemd-networkd-wait-online.service is not mean to be used in this way.

If "RequiredForOnline=no" sd-networkd-wait-online will fully ignore the corresponding interface and it will block/delay network-online.target if no interfaces are "RequiredForOnline=yes" at all.

This is as per discussions in https://github.com/systemd/systemd/pull/32016

FR-7246

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2060311

### DIFF
--- a/data/netplan-networkd-wait-online@.service
+++ b/data/netplan-networkd-wait-online@.service
@@ -1,0 +1,26 @@
+#  SPDX-License-Identifier: LGPL-2.1-or-later
+#
+#  This file is part of systemd.
+#
+#  systemd is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+[Unit]
+Description=Wait for Network Interface %i to be Configured
+Documentation=man:systemd-networkd-wait-online.service(8)
+ConditionCapability=CAP_NET_ADMIN
+DefaultDependencies=no
+Conflicts=shutdown.target
+BindsTo=systemd-networkd.service
+After=systemd-networkd.service
+Before=network-online.target shutdown.target
+
+[Service]
+Type=oneshot
+ExecStart=/lib/systemd/systemd-networkd-wait-online -i %i
+RemainAfterExit=yes
+
+[Install]
+WantedBy=network-online.target
--- a/src/generate.c
+++ b/src/generate.c
@@ -75,21 +75,6 @@ enable_networkd(const char* generator_dir, const char* rootdir)
         // LCOV_EXCL_STOP
     }
 
-    // Mask the generic systemd-networkd-wait-online.service unit, to enforce
-    // "netplan-networkd-wait-online@IFACE.service" wait-online policy
-    g_autofree gchar* masking_link = NULL;
-    masking_link = g_build_path(G_DIR_SEPARATOR_S, rootdir ?: G_DIR_SEPARATOR_S,
-                                "run/systemd/system/systemd-networkd-wait-online.service",
-                                NULL);
-    g_debug("Masking systemd-networkd-wait-online.service via %s -> /dev/null", masking_link);
-    _netplan_safe_mkdir_p_dir(masking_link);
-    if (symlink("/dev/null", masking_link) < 0 && errno != EEXIST) {
-        // LCOV_EXCL_START
-        g_fprintf(stderr, "failed to create masking symlink: %m\n");
-        exit(1);
-        // LCOV_EXCL_STOP
-    }
-
     // Copy /usr/lib/systemd/system/systemd-networkd-wait-online@.service to
     // /run/systemd/system/netplan-networkd-wait-online@.service so we can make
     // it our own service.
--- a/src/generate.c
+++ b/src/generate.c
@@ -63,7 +63,7 @@ reload_udevd(void)
  * Create enablement symlink for systemd-networkd.service.
  */
 static void
-enable_networkd(const char* generator_dir)
+enable_networkd(const char* generator_dir, const char* rootdir)
 {
     g_autofree char* link = g_build_path(G_DIR_SEPARATOR_S, generator_dir, "multi-user.target.wants", "systemd-networkd.service", NULL);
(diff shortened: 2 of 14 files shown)


src/networkd.c
```diff
@@ -32,6 +33,68 @@
...
+
+    return g_path_get_basename(link);
+}
+
+/**
+ * Enumerate all network interfaces (/sys/clas/net/...) and check
+ * netplan_netdef_match_interface() to see if they match the current NetDef
+ */
+STATIC void
+_netplan_enumerate_interfaces(const NetplanNetDefinition* def, GStrvBuilder* builder, const char* rootdir)
```
enr0n, line 83: Is this here to work around the fact that we could have a generated file like:

```
# my-interface.network
[Match]
MacAddress=AA:BB:CC:DD:EE:FF

[Network]
...
```

but we want to enable e.g. `systemd-networkd-wait-online@ethZ.service`, if `ethZ` is the interface with the matching MAC?

enr0n (commented):
This looks good so far to me.

One other thing is that we will actually need to mask `systemd-networkd-wait-online.service` by default I think, because otherwise e.g. a unit generated by `systemd-sysv-generator` could re-enable it, which is just annoying.

Since netplan would be the one enforcing this new `systemd-networkd-wait-online@.service` policy, do you think it makes sense to ship the mask (symlink to /dev/null) in netplan?

src/networkd.c
slyon: Yes! We need to expand any Netplan `match.mac|driver|name` setting to the actual matching network interfaces. Therefore, we need to check those values on all of them.

data/netplan-networkd-wait-online@.service
```diff
@@ -0,0 +1,26 @@
+#  SPDX-License-Identifier: LGPL-2.1-or-later
```
slyon, line 1: note: This should be compatible with our GPL-3 code, according to https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

src/networkd.c
```diff
@@ -1398,6 +1452,48 @@ _netplan_netdef_write_networkd(
...
+
+        // create run/systemd/system/network-online.target.wants/ dir
+        if (g_strv_length(ifnames) > 0) {
+            g_autofree gchar* enablement_dir = NULL;
+            enablement_dir = g_build_path(G_DIR_SEPARATOR_S, rootdir ?: G_DIR_SEPARATOR_S,
+                                          "run", "systemd", "system", "network-online.target.wants", "NOFILE", NULL);
+            _netplan_safe_mkdir_p_dir(enablement_dir);
+        }
+
+        // Create an enablement link for each interface in the <ifnames> vector
```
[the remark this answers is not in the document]
slyon: ACK. See spec FO165

src/generate.c
```diff
@@ -75,11 +75,17 @@ enable_networkd(const char* generator_dir)
...
+                                NULL);
+    g_debug("Masking systemd-networkd-wait-online.service via %s -> /dev/null", masking_link);
+    _netplan_safe_mkdir_p_dir(masking_link);
+    if (symlink("/dev/null", masking_link) < 0 && errno != EEXIST) {
         // LCOV_EXCL_START
-        g_fprintf(stderr, "failed to create enablement symlink: %m\n");
+        g_fprintf(stderr, "failed to create masking symlink: %m\n");
         exit(1);
         // LCOV_EXCL_STOP
     }
```
slyon, line 91: note: I'm not sure if that change should be generally included in our upstream codebase. We usually do not want to hard-enforce policy like this, but let everybody do what they want in parallel to Netplan.

Could be an Ubuntu specific patch, though.

slyon:
I've built a test-package in a PPA. May I ask @waveform80 and @frank-heimes for testing?

src/networkd.c
```diff
@@ -1398,6 +1452,48 @@ _netplan_netdef_write_networkd(
     if (!_netplan_netdef_write_network_file(np_state, def, rootdir, path_base, has_been_written, error))
         return FALSE;
     SET_OPT_OUT_PTR(has_been_written, TRUE);
+
+    /* When activation-mode is used we default to being optional.
+     * Otherwise, systemd might wait indefinitely for the interface to
+     * come online.
+     */
+    if (!(def->optional || def->activation_mode)) {
```
slyon, line 1460: question: Should we also check if IP `addresses` or `dhcp4`/`dhcp6` are defined, e.g. to ignore definitions like this `ethernets.encc000: {}`?

We could define such scenarios as "user error", or try to be smart about it..

slyon: I'd probably opt to "user error" for now (the easy workaround is adding `optional: true` instead of `{}`) and in the future we might want to wait for different conditions, e.g. link-layer configuration, which would still involve  some waiting, but not necessarily on routable IPs.

src/generate.c
enr0n: As discussed elsewhere, I think we should try to instead generate a drop-in like:

```
[Service]
ExecStart=
ExecStart=/lib/systemd/systemd-networkd-wait-online -i eth0 -i wlan0 ... -i ethX
```

where eth0, wlan0, ..., ethx are required interfaces. This would eliminate the need for using the templated units, and hence obviate the need for disabling or masking `systemd-networkd-wait-online.service`.

src/networkd.c
daniloegea: I'm not sure if I understood the question. Are you suggesting we shouldn't create the link for that type of definition (empty)?

src/generate.c
slyon: see https://github.com/canonical/netplan/pull/456

src/networkd.c
slyon: Empty NetDefs will not reach the "online" state, due to lacking an IP address. I was wondering if we should implicitly mark them "optional: true". But I now feel like we shouldn't do it, as logic might change in the future with the implementation of "optional-addresses".

PR #456, opened 2024-04-16
Create custom systemd-networkd-wait-online.service override to wait on individual interfaces. (LP: #2060311)

slyon:
## Description
Skip s-n-wait-online if we don't have any non-optional interfaces, using a "ConditionPathIsSymbolicLink=" checking Netplan's s-n-wait-online.service enablement symlink.

This is in favor to RequiredForOnline=yes as the behavior of upstream (pure)
systemd-networkd-wait-online.service is not mean to be used in this way.
    
If "RequiredForOnline=no" sd-networkd-wait-online will fully ignore the
corresponding interface and it will block/delay network-online.target if
no interfaces are "RequiredForOnline=yes" at all.
    
FR-7246

Note: This is a replacement for https://github.com/canonical/netplan/pull/455 that keeps compatibility with cloud-init, so it can still sort `After=systemd-networkd-wait-online.service` AND `Before=network-online.target`.

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2060311

### DIFF
--- a/src/generate.c
+++ b/src/generate.c
@@ -302,7 +302,10 @@ int main(int argc, char** argv)
     if (netplan_state_get_backend(np_state) == NETPLAN_BACKEND_NM || any_nm)
         _netplan_g_string_free_to_file(g_string_new(NULL), rootdir, "/run/NetworkManager/conf.d/10-globally-managed-devices.conf", NULL);
 
-    gboolean enable_wait_online = _netplan_networkd_write_wait_online(np_state, rootdir);
+    gboolean enable_wait_online = FALSE;
+    if (any_networkd)
+        enable_wait_online = _netplan_networkd_write_wait_online(np_state, rootdir);
+
     if (called_as_generator) {
         /* Ensure networkd starts if we have any configuration for it */
         if (any_networkd)
--- a/src/generate.c
+++ b/src/generate.c
@@ -63,7 +63,7 @@ reload_udevd(void)
  * Create enablement symlink for systemd-networkd.service.
  */
 static void
-enable_networkd(const char* generator_dir)
+enable_networkd(const char* generator_dir, gboolean enable_wait_online)
 {
     g_autofree char* link = g_build_path(G_DIR_SEPARATOR_S, generator_dir, "multi-user.target.wants", "systemd-networkd.service", NULL);
     g_debug("We created networkd configuration, adding %s enablement symlink", link);
@@ -75,13 +75,15 @@ enable_networkd(const char* generator_dir)
         // LCOV_EXCL_STOP
     }
 
-    g_autofree char* link2 = g_build_path(G_DIR_SEPARATOR_S, generator_dir, "network-online.target.wants", "systemd-networkd-wait-online.service", NULL);
-    _netplan_safe_mkdir_p_dir(link2);
-    if (symlink("/lib/systemd/system/systemd-networkd-wait-online.service", link2) < 0 && errno != EEXIST) {
-        // LCOV_EXCL_START
-        g_fprintf(stderr, "failed to create enablement symlink: %m\n");
-        exit(1);
-        // LCOV_EXCL_STOP
+    if (enable_wait_online) {
+        g_autofree char* link2 = g_build_path(G_DIR_SEPARATOR_S, generator_dir, "network-online.target.wants", "systemd-networkd-wait-online.service", NULL);
+        _netplan_safe_mkdir_p_dir(link2);
+        if (symlink("/lib/systemd/system/systemd-networkd-wait-online.service", link2) < 0 && errno != EEXIST) {
+            // LCOV_EXCL_START
+            g_fprintf(stderr, "failed to create enablement symlink: %m\n");
+            exit(1);
+            // LCOV_EXCL_STOP
+        }
     }
 }
 
@@ -300,10 +302,11 @@ int main(int argc, char** argv)
     if (netplan_state_get_backend(np_state) == NETPLAN_BACKEND_NM || any_nm)
         _netplan_g_string_free_to_file(g_string_new(NULL), rootdir, "/run/NetworkManager/conf.d/10-globally-managed-devices.conf", NULL);
 
+    gboolean enable_wait_online = _netplan_networkd_write_wait_online(np_state, rootdir);
     if (called_as_generator) {
         /* Ensure networkd starts if we have any configuration for it */
         if (any_networkd)
-            enable_networkd(files[0]);
+            enable_networkd(files[0], enable_wait_online);
 
         /* Leave a stamp file so that we don't regenerate the configuration
          * multiple times and userspace can wait for it to finish */
@@ -324,7 +327,8 @@ int main(int argc, char** argv)
         /* covered via 'cloud-init' integration test */
         if (any_networkd) {
             start_unit_jit("systemd-networkd.socket");
-            start_unit_jit("systemd-networkd-wait-online.service");
+            if (enable_wait_online)
+                start_unit_jit("systemd-networkd-wait-online.service");
             start_unit_jit("systemd-networkd.service");
         }
         g_autofree char* glob_run = g_build_path(G_DIR_SEPARATOR_S,
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -1498,12 +1498,18 @@ _netplan_networkd_write_wait_online(const NetplanState* np_state, const char* ro
 
     // We have non-optional interface, so let's wait for those explicitly
     GHashTableIter idx;
-    gpointer ifname;
+    gpointer key;
     g_string_append(content, "\n[Service]\nExecStart=\n"
                                 "ExecStart=/lib/systemd/systemd-networkd-wait-online");
     g_hash_table_iter_init (&idx, non_optional_interfaces);
-    while (g_hash_table_iter_next (&idx, &ifname, NULL))
-        g_string_append_printf(content, " -i %s", (const char*)ifname);
+    while (g_hash_table_iter_next (&idx, &key, NULL)) {
+        const char* ifname = key;
+        g_string_append_printf(content, " -i %s", ifname);
+        // XXX: We should be checking IFF_LOOPBACK instead of interface name.
+        //      But don't have access to the flags here.
+        if (!g_strcmp0(ifname, "lo"))
+            g_string_append(content, ":carrier"); // "carrier" as min-oper state for loopback
+    }
     g_string_append(content, "\n");
 
     _netplan_g_string_free_to_file(content, rootdir, override, NULL);
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -75,25 +75,39 @@ _netplan_sysfs_get_driver_by_ifname(const char* ifname, const char* rootdir)
     return g_path_get_basename(link);
 }
 
-/**
- * Enumerate all network interfaces (/sys/clas/net/...) and check
- * netplan_netdef_match_interface() to see if they match the current NetDef
- */
 STATIC void
-_netplan_enumerate_interfaces(const NetplanNetDefinition* def, GHashTable* tbl, const char* rootdir)
+_netplan_query_system_interfaces(GHashTable* tbl)
 {
     g_assert(tbl);
     struct if_nameindex *if_nidxs, *intf;
     if_nidxs = if_nameindex();
     if (if_nidxs != NULL) {
-        for (intf = if_nidxs; intf->if_index != 0 || intf->if_name != NULL; intf++) {
-            if (g_hash_table_contains(tbl, intf->if_name)) continue;
-            g_autofree gchar* mac = _netplan_sysfs_get_mac_by_ifname(intf->if_name, rootdir);
-            g_autofree gchar* driver = _netplan_sysfs_get_driver_by_ifname(intf->if_name, rootdir);
-            if (netplan_netdef_match_interface(def, intf->if_name, mac, driver))
-                g_hash_table_add(tbl, g_strdup(intf->if_name));
-        }
+        for (intf = if_nidxs; intf->if_index != 0 || intf->if_name != NULL; intf++)
+            g_hash_table_add(tbl, g_strdup(intf->if_name));
         if_freenameindex(if_nidxs);
+    }
+}
+
+/**
+ * Enumerate all network interfaces (/sys/clas/net/...) and check
+ * netplan_netdef_match_interface() to see if they match the current NetDef
+ */
+STATIC void
+_netplan_enumerate_interfaces(const NetplanNetDefinition* def, GHashTable* ifaces, GHashTable* tbl, const char* carrier, const char* set_name, const char* rootdir)
+{
+    g_assert(ifaces);
+    g_assert(tbl);
+
+    GHashTableIter iter;
+    gpointer key;
+    g_hash_table_iter_init (&iter, ifaces);
+    while (g_hash_table_iter_next (&iter, &key, NULL)) {
+        const char* ifname = key;
+        if (g_hash_table_contains(tbl, ifname)|| (set_name && g_hash_table_contains(tbl, set_name))) continue;
+        g_autofree gchar* mac = _netplan_sysfs_get_mac_by_ifname(ifname, rootdir);
+        g_autofree gchar* driver = _netplan_sysfs_get_driver_by_ifname(ifname, rootdir);
+        if (netplan_netdef_match_interface(def, ifname, mac, driver))
+            g_hash_table_replace(tbl, set_name ? g_strdup(set_name) : g_strdup(ifname), g_strdup(carrier));
     }
 }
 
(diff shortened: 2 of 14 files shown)


slyon:
PPA for testing can be found in: https://launchpad.net/~slyon/+archive/ubuntu/lp2060311/+packages

src/networkd.c
```diff
@@ -1499,6 +1499,15 @@ _netplan_networkd_write_wait_online(const NetplanState* np_state, const char* ro
...
         g_hash_table_destroy(non_optional_interfaces);
         return TRUE;
+    } else {
+        // Systemd's /lib/systemd/system/systemd-networkd.service installs an
+        // enablement link for systemd-networkd-wait-online.service in its
+        // [Install] section via "Also=systemd-networkd-wait-online.service".
+        // We want to avoid it being called without --interface arguments, to
+        // avoid a boot delay, when waiting for configured, but optional interfaces.
+        GString* content_noop = g_string_new("[Service]\nExecStart=\n"
+                                             "ExecStart=/bin/true\n");
```
slyon, line 1509: note: I'm not so sure we want this behavior by default, at least not on non-Ubuntu systems.

src/networkd.c
```diff
@@ -1466,6 +1457,60 @@ _netplan_netdef_write_networkd(
...
+                // XXX: we shouldn't run this enumeration for every NetDef...
+                _netplan_enumerate_interfaces(def, non_optional_interfaces, rootdir);
+            }
+        }
+    }
+
+    // create run/systemd/system/systemd-networkd-wait-online.service.d/
+    const char* override = "/run/systemd/system/systemd-networkd-wait-online.service.d/10-netplan.conf";
+    // The "ConditionPathIsSymbolicLink" is Netplan's s-n-wait-online enablement symlink,
+    // as we want to run -wait-only only if enabled by Netplan.
```
daniloegea, line 1490: s/wait-only/wait-online/ ?

src/generate.c
```diff
@@ -300,10 +302,11 @@ int main(int argc, char** argv)
     if (netplan_state_get_backend(np_state) == NETPLAN_BACKEND_NM || any_nm)
         _netplan_g_string_free_to_file(g_string_new(NULL), rootdir, "/run/NetworkManager/conf.d/10-globally-managed-devices.conf", NULL);
 
+    gboolean enable_wait_online = _netplan_networkd_write_wait_online(np_state, rootdir);
```
slyon, line 305: Note: we should consider writing the s-n-wait-online override only if we have "any_networkd", to avoid fiddling with networkd when nothing is configured through Netplan.

src/networkd.c
```diff
@@ -1466,6 +1457,60 @@ _netplan_netdef_write_networkd(
...
+    if (g_hash_table_size(non_optional_interfaces) == 0) {
+        _netplan_g_string_free_to_file(content, rootdir, override, NULL);
+        g_hash_table_destroy(non_optional_interfaces);
+        return FALSE;
+    }
+
+    // We have non-optional interface, so let's wait for those explicitly
+    GHashTableIter idx;
+    gpointer ifname;
+    g_string_append(content, "\n[Service]\nExecStart=\n"
```
daniloegea, line 1502: Why do we need this empty `ExecStart=`?

daniloegea:
This seem to work well. The new `10-netplan.conf` file can get overcrowded if one has lots of interfaces on Netplan but I guess it shouldn't be a problem (unless you have enough interfaces to reach the maximum number of command line arguments o.O).

I did find a case though where the behavior changes: when you have the loopback interface in your yaml. Apparently `-wait-online -i lo` will always get stuck. So, people with `ethernets.lo` in their configuration will notice that wait-online will start to block after this chance. Maybe `lo` should be a special case and never included in the list?

src/networkd.c
enr0n: It is valid to have multiple `ExecStart=`. Using an empty one clears the current list so that you can completely override it.

enr0n:
> This seem to work well. The new `10-netplan.conf` file can get overcrowded if one has lots of interfaces on Netplan but I guess it shouldn't be a problem (unless you have enough interfaces to reach the maximum number of command line arguments o.O).
> 
> I did find a case though where the behavior changes: when you have the loopback interface in your yaml. Apparently `-wait-online -i lo` will always get stuck. So, people with `ethernets.lo` in their configuration will notice that wait-online will start to block after this chance. Maybe `lo` should be a special case and never included in the list?

Yeah either ignore it or would have to set a different operational state range, e.g. `-o carrier`. But the default behavior of  systemd-networkd-wait-online (i.e. when no `-i` are specified on the command line) is to ignore loopback.

enr0n (approved):
The new behavior matches what we have discussed. I did some testing with various combinations of `optional: ` for `eth0` and `wlan0` and things work as expected in all cases.

TheRealFalcon:
While I'm not seeing issues specifically related to cloud-init, this doesn't seem to be working correctly, or I'm not understanding how it is supposed to work. After install the PPA, if I reboot, regardless what I have under `/etc/netplan`,  `/run/systemd/system/systemd-networkd-wait-online.service.d/10-netplan.conf` contains:
```
[Unit]
ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait-online.service
```

and nothing else.

slyon:
> regardless what I have under `/etc/netplan`, `/run/systemd/system/systemd-networkd-wait-online.service.d/10-netplan.conf` contains:
> 
> ```
> [Unit]
> ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait-online.service
> ```
> 
> and nothing else.

As long as you have any network definition in /etc/netplan/ that uses networkd as a renderer and does NOT use "optional: true" you should see the corresponding interface listed in that file after calling "netplan apply" (or reboot).

TheRealFalcon:
Does it matter that I'm in an LXD container?

```
root@asdf:~# cat /etc/netplan/50-cloud-init.yaml 
# This file is generated from information provided by the datasource.  Changes
# to it will not persist across an instance reboot.  To disable cloud-init's
# network configuration capabilities, write a file
# /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with the following:
# network: {config: disabled}
network:
    renderer: networkd
    ethernets:
        eth0:
            dhcp4: true
            match:
                macaddress: 12:34:56:78:90:ab
            # optional: true
        eth1:
            dhcp4: true
            match:
                macaddress: 00:16:3e:38:c3:9e
            # optional: false
    version: 2
root@asdf:~# netplan apply
root@asdf:~# cat /run/systemd/system/systemd-networkd-wait-online.service.d/10-netplan.conf 
[Unit]
ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait-online.service
root@asdf:~# reboot
root@asdf:~# 
Session terminated, killing shell...
 ...killed.
(cloud-init37) cow:net $ lxc shell asdf
root@asdf:~# cat /run/systemd/system/systemd-networkd-wait-online.service.d/10-netplan.conf 
[Unit]
ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait-online.service
root@asdf:~# 
```

slyon:
> Does it matter that I'm in an LXD container?

It worked in an LXD container for me earlier today. But it could have some impact on the macaddress matching.. can you try with a different matching condition, like name or driver? Looks like it doesn't find the interface here. I need to double-check that.

TheRealFalcon:
@slyon , doh, sorry. This was after I tore down a previous container and forgot to update my network configuration. Once I corrected the MAC address, it worked for me.

TheRealFalcon:
Looks like cloud-init is still having issues first boot. During our init-local timeframe, we write out netplan config, then call `netplan generate`. After boot is finished, checking the `systemd-networkd-wait-online.service`, systemd says we need to call `daemon-reload`.

This can be simulated in a container by running `cloud-init clean --logs; rm /etc/netplan/50-cloud-init.yaml`, then reboot. 

After boot:
```
root@asdf:~# ls /run/systemd/generator/
getty.target.wants  local-fs.target.wants  netplan.stamp
root@asdf:~# systemctl status systemd-networkd-wait-online.service 
Warning: The unit file, source configuration file or drop-ins of systemd-networkd-wait-online.service changed on disk. Run 'systemctl daemon-reload' to reload units.
○ systemd-networkd-wait-online.service - Wait for Network to be Configured
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd-wait-online.service; enabled; preset: enabled)
    Drop-In: /run/systemd/system/systemd-networkd-wait-online.service.d
             └─10-netplan.conf
     Active: inactive (dead)
  Condition: start condition unmet at Tue 2024-04-16 21:40:25 UTC; 54s ago
             └─ ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait-online.service was not met
       Docs: man:systemd-networkd-wait-online.service(8)

Apr 16 21:40:25 asdf systemd[1]: systemd-networkd-wait-online.service - Wait for Network to be Configured was skipped because of an unmet condition check (ConditionPathIsSymbolicLink=/run/systemd/generator/network-online.target.wants/systemd-networkd-wait>
root@asdf:~# systemctl daemon-reload
root@asdf:~# ls /run/systemd/generator/
getty.target.wants  local-fs.target.wants  multi-user.target.wants  netplan.stamp  network-online.target.wants
root@asdf:~# 
```

daniloegea:
Another thing I've found is that `-wait-online` blocks if the interface doesn't exist. So if you have interfaces in your YAMLs that are not present in the system it will block now. The previous behavior is to not block.

slyon:
Thanks for all of your testing and comments!

I put all fixes into follow-up commits, so it can be reviewed more easily. @daniloegea PITAL at the most recent commits.
I've built the current version in my PPA as `~ppa6`, and it is fixing:

* no blocking on loopback interface
* no blocking on non-existing interfaces
* no need for manual `daemon-reload` in cloud-init "init-local" timeframe (that is now covered by the `netplan generate` CLI, which is being called anyway)
* Interfaces with IP configuration (dhcp4, dhcp6, static IP or link-local IP) will be listed as `-i IFACE:degraded`
  * Interfaces without IP configuration will be listed as `-i IFACE:carrier`, to avoid blocking on non-existing IPs
* Fixed typos
* no fiddling with `systemd-networkd-wait-online.service.d/10-netplan.conf` override when no networkd-based NetDefs are defined in Netplan's configuration.
  * We do not enable `systemd-networkd.service` in this case, so we should not modify `s-n-wait-online.service` either.
* Included integrationtest/autopkgtest, checking for MAC address matching, too, on a real system/VM
  * see `ethernets.py -> test_systemd_networkd_wait_online`

@TheRealFalcon I addressed the `daemon-reload` failure, and your reproducer seems to be fine now:
```
root@well-pug:~# cloud-init clean --logs; rm /etc/netplan/50-cloud-init.yaml
root@well-pug:~# reboot

$ lxc shell well-pug 

root@well-pug:~# ls /run/systemd/generator/
getty.target.wants  local-fs.target.wants  multi-user.target.wants  netplan.stamp  network-online.target.wants
root@well-pug:~# systemctl status systemd-networkd-wait-online.service 
● systemd-networkd-wait-online.service - Wait for Network to be Configured
     Loaded: loaded (/usr/lib/systemd/system/systemd-networkd-wait-online.service; enabled; preset: enabled)
    Drop-In: /run/systemd/system/systemd-networkd-wait-online.service.d
             └─10-netplan.conf
     Active: active (exited) since Wed 2024-04-17 13:05:40 UTC; 18s ago
       Docs: man:systemd-networkd-wait-online.service(8)
    Process: 301 ExecStart=/lib/systemd/systemd-networkd-wait-online -i eth0:degraded (code=exited, status=0/SUCCESS)
   Main PID: 301 (code=exited, status=0/SUCCESS)
        CPU: 6ms

Apr 17 13:05:39 well-pug systemd[1]: Starting systemd-networkd-wait-online.service - Wait for Network to be Configured.
Apr 17 13:05:40 well-pug systemd[1]: Finished systemd-networkd-wait-online.service - Wait for Network to be Configured.
```

src/networkd.c
```diff
@@ -1460,29 +1474,41 @@ _netplan_netdef_write_networkd(
...
          * come online.
          */
         if (!(def->optional || def->activation_mode)) {
-            if (!netplan_netdef_has_match(def)) { // no matching => single interface
-                g_hash_table_add(non_optional_interfaces, g_strdup(def->id));
-            } else if (def->set_name) { // matching on a single interface
-                g_hash_table_add(non_optional_interfaces, g_strdup(def->set_name));
+            // Check if we have any IP configuration
+            struct address_iter* addr_iter = _netplan_netdef_new_address_iter(def);
+            gboolean any_ips = _netplan_address_iter_next(addr_iter) != NULL || netplan_netdef_get_link_local_ipv4(def) || netplan_netdef_get_link_local_ipv6(def);
```
daniloegea, line 1497: question: should dhcp4 and dhcp6 be considered here? When I have the config below I get `enp3s0:carrier` but according to your last comment it should be `:degraded`?

```yaml
network:
  ethernets:
    enp3s0:
      dhcp4: true
      link-local: []
```

daniloegea (commented):
I was doing some testing based on you comments and I noticed that dhcp is not considered to set the interface as degraded. So if I disabled link-local and enable dhcp4 it will be set to carrier. Is that expected?

src/networkd.c
slyon: Good catch! Of course, it needs to be considered. Fixing...

src/networkd.c
```diff
@@ -1460,29 +1474,41 @@ _netplan_netdef_write_networkd(
 gboolean
 _netplan_networkd_write_wait_online(const NetplanState* np_state, const char* rootdir)
 {
+    // Set of all current network interfaces, potentially non yet renamed
+    GHashTable* system_interfaces = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+    _netplan_query_system_interfaces(system_interfaces);
```
blackboxsw, line 1479: I'm not certain here about the impact of ignoring devices not currently present in /sys/class/net during systemd generator timeframe.  By only considering interfaces already present in /sys/class/net, are we exposed to scenarios where the device drivers for the specified device haven't loaded yet (in initramfs-less environments without the proper drivers compiled into the kernel)?   This reminds me of this cloud-init issue https://github.com/canonical/cloud-init/issues/4451 for minimal images.   In the event that we don't have device drivers loaded (from initramfs or kernel static compiled), is it possible netplan's generator may not see the device in /sys/class/net via `if_nameindex` and systemd-networkd-wait-online will be disabled right?  Maybe in this case, we prescribe that initramfs/kernel must have the desired device drivers in order to expect systemd-networkd-wait-online to block properly.

daniloegea: A similar (the same?) issue will happen with virtual devices. By the time the generator runs they will not be created yet so they will not be in the list.

blackboxsw: James corrected me, this isn't systemd generator timeframe, but the point in time when `netplan  generate` is called.   For cloud-init init-local  boot stage, I think we may still be exposed to this race where cloud-init local gets in any invokes `netplan generate` possibly before supplemental device drivers have loaded. So this may still be a concern.

slyon: Oh! That's a very interesting case.. I just pushed more commits to not require virtual devices (bridges, bonds, dummys, ...) to be available in `/sys/class/net` just yet. Those will be created later.

For physical devices that are not yet loaded, I don't know. I don't think there's anything that can be done as part of this PR and such interfaces might be ignored by `systemd-networkd-wait-online`.

A longer-term solution might be splitting up Netplan into a sytemd-generator and a systemd-service (running `Before=network-pre.target`, similar to `systemd-network-generator.service`), which we are tracking as spec "FO165 - Netplan generator architecture".

Do you think this should be a blocker?

slyon:
Added more commits fixing:
* considering `dhcp4: true` and `dhcp6: true` for "degraded" operational state.
* do not require virtual devices to be created already

src/networkd.c
blackboxsw: I think this retains the same gap/issue cloud-init and systemd-networkd-wait-online are already exposed to, so I don't think this needs to be fixed in this PR.  The stance we've taken already [Ubuntu minimal case](https://bugs.launchpad.net/cloud-images/+bug/2036968) was to recompile the kernel to include the module built-ins that are required to support virtio-net to ensure they are loaded in time for network generation and network-online for "initrdless" boots. It feels "ok" to state that if you have a critical device driver needed in early boot for your cloud image and you are booting without initramrd, you would either need to statically compile the necessary device drivers into that kernel, or use initrd and provide any supplemental drivers needed earlier boot. This may just be something we take into consideration for future behavior though, and thinking of how to approach this type of concern later in F0165 probably makes the most sense at the moment.

daniloegea:
Thanks, Lukas. The cases I've found seem to be fixed now.

daniloegea (approved):

slyon:
Added a tiny fix to recognize that we do not configure link-local IPs on bridge/bond members.

PR #459, opened 2024-04-18
Fix wrong syntax in example

fzakfeld:
Syntax in example was wrong


slyon (approved):
LGTM, thank you very much for your contribution to Netplan!

PR #460, opened 2024-04-18
parse: fix redefinition of gateway(4|6)

daniloegea:
Options gateway4 and gateway6 were not being properly redefined when a merge happens. Redefinition was causing a crash due to an assert in the set_str_if_null macro.

While this assertion helped identifying the merge issue, I believe it's not the best place to keep it so I'm removing it from the macro and making the macro do only what it name suggests it does.

This macro is used in other places but this problem seems to not happen there as far as I can see.

Add a couple of unit test to check the merge is working.

## Description

Reproducer:

```yaml
network:
  ethernets:
    eth0:
      addresses:
        - 192.168.0.1/24
      gateway4: 192.168.0.123
    eth0:
      gateway4: 192.168.0.111
```
```
$ netplan generate --root-dir /tmp/fakeroot4

** (generate:190262): WARNING **: 18:08:41.479: `gateway4` has been deprecated, use default routes instead.
See the 'Default routes' section of the documentation for more details.
**
ERROR:../src/parse.c:1434:handle_gateway4: assertion failed (scalar(node) == npp->current.netdef->gateway4): ("192.168.0.111" == "192.168.0.123")
Bail out! ERROR:../src/parse.c:1434:handle_gateway4: assertion failed (scalar(node) == npp->current.netdef->gateway4): ("192.168.0.111" == "192.168.0.123")
```
## Checklist

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -49,9 +49,7 @@
 #define vxlan_offset(field) GUINT_TO_POINTER(offsetof(NetplanVxlan, field))
 
 /* convenience macro to avoid strdup'ing a string into a field if it's already set. */
-#define set_str_if_null(dst, src) { if (dst) {\
-    g_assert_cmpstr(src, ==, dst); \
-} else { \
+#define set_str_if_null(dst, src) { if (dst == NULL) {\
     dst = g_strdup(src); \
 } }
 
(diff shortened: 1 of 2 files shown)


slyon (approved):
Thanks, lgtm.

PR #464, opened 2024-04-25
python: elements of __all__ must be strings

daniloegea:
## Checklist

### DIFF
--- a/netplan_cli/__init__.py
+++ b/netplan_cli/__init__.py
@@ -17,4 +17,4 @@
 
 from .cli.core import Netplan
 
-__all__ = [Netplan]
+__all__ = ['Netplan']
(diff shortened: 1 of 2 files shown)


slyon (approved):
LGTM

According to [Python's reference](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement):
> If the list of identifiers is replaced by a star ('*'), all public names defined in the module are bound in the local namespace for the scope where the import statement occurs.
>
> The public names defined by a module are determined by checking the module’s namespace for a variable named __all__; if defined, it must be a sequence of strings which are names defined or imported by that module. The names given in __all__ are all considered public and are required to exist. If __all__ is not defined, the set of public names includes all names found in the module’s namespace which do not begin with an underscore character ('_'). __all__ should contain the entire public API. It is intended to avoid accidentally exporting items that are not part of the API (such as library modules which were imported and used within the module).

PR #466, opened 2024-05-03
netplan status --diff fixes and improvements

daniloegea:
This PR contains a bunch of small improvements to `netplan status --diff` and some additions to the libnetplan API:

- Add getters for the `gateway4` and `gateway6` properties. This is required during the analysis of routes.
- Routes from `gateway4` and `gateway6` are added to the list of routes being analyzed.
- Make use of the property `dynamic` from `addr_info` to help determining if the address was received dynamically.
- Make use of the `ra` flag from routes to help finding IPs that were received via RA.
- Add getter for the `accept-ra` property.
- Improve the analysis of RA and LL addresses.
- Fix a rendering issue in the `--diff` mode when `python3-rich` is not present.
- Use the ConfigSource property from networkd to reduce the guessing and be more precise when analyzing dynamic configuration.

I recommend testing it inside a LXD VM where IPV6 addresses are assigned dynamically. The currently netplan.io will show the IPv6 address and DNS as diffs. The heuristics added to the analysis help to reduce this noise.

## Checklist

### DIFF
--- a/include/netdef.h
+++ b/include/netdef.h
@@ -205,6 +205,16 @@ NETPLAN_PUBLIC gboolean
 netplan_netdef_get_link_local_ipv6(const NetplanNetDefinition* netdef);
 
 /**
+ * @brief   Query a @ref NetplanNetDefinition for the value of its `accept-ra` setting.
+ * @param[in] netdef The @ref NetplanNetDefinition to query
+ * @return           Indication if @p netdef is configured to accept Router Advertisements.
+ *                   Possible values are: 0) not set (will use kernel or back end defaults),
+ *                   1) enabled and 2) disabled.
+ */
+NETPLAN_PUBLIC int
+netplan_netdef_get_accept_ra(const NetplanNetDefinition* netdef);
+
+/**
  * @brief   Get the `macaddress` setting of a given @ref NetplanNetDefinition.
  * @details Copies a `NUL`-terminated string into a sized @p out_buffer. If the
  *          buffer is too small, its content is not `NUL`-terminated.
--- a/include/netdef.h
+++ b/include/netdef.h
@@ -227,5 +227,3 @@ netplan_netdef_get_accept_ra(const NetplanNetDefinition* netdef);
  */
 NETPLAN_PUBLIC ssize_t
 netplan_netdef_get_macaddress(const NetplanNetDefinition* netdef, char* out_buffer, size_t out_buffer_size);
-
-
--- a/include/netdef.h
+++ b/include/netdef.h
@@ -217,3 +217,5 @@ netplan_netdef_get_link_local_ipv6(const NetplanNetDefinition* netdef);
  */
 NETPLAN_PUBLIC ssize_t
 netplan_netdef_get_macaddress(const NetplanNetDefinition* netdef, char* out_buffer, size_t out_buffer_size);
+
+
--- a/netplan_cli/cli/state_diff.py
+++ b/netplan_cli/cli/state_diff.py
@@ -145,7 +145,7 @@ class NetplanDiffState():
             flags = addr_data.get('flags', [])
 
             # Select only static IPs
-            if not {'dhcp', 'dynamic', 'link'}.intersection(flags):
+            if not {'dhcp', 'dynamic', 'link', 'ra'}.intersection(flags):
                 system_ips.add(addr)
 
             # Handle the link local address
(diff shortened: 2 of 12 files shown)


netplan_cli/cli/state_diff.py
```diff
@@ -442,6 +486,62 @@ def _normalize_routes(self, routes: set) -> set:
...
+        interface = config.get('id')
+
+        ifaces = [iface for iface in system_interfaces.values()
+                  if iface.get('id') == interface]
+        routes = [route for iface in ifaces for route in iface.get('routes', [])]
+
+        if gateway4 := config.get('gateway4'):
+            default_routes = [route for route in routes
+                              if route.get('to') == 'default'
+                              and route.get('family') == 2
```
slyon, line 511: nitpick: use `socket.AF_INET` named constant, also in the `route = NetplanRoute(...)` below.

And same for `socket.AF_INET6` in `gateway6` below.

netplan_cli/cli/state_diff.py
```diff
@@ -442,6 +486,62 @@ def _normalize_routes(self, routes: set) -> set:
...
+        system_interfaces = self.system_state.get_data()
+        interface = config.get('id')
+
+        ifaces = [iface for iface in system_interfaces.values()
+                  if iface.get('id') == interface]
+        routes = [route for iface in ifaces for route in iface.get('routes', [])]
+
+        if gateway4 := config.get('gateway4'):
+            default_routes = [route for route in routes
+                              if route.get('to') == 'default'
```
slyon, line 510: question (non-blocking): Are we sure this is already normalized to `default`, or could it also be something like `0.0.0.0/0`?

netplan_cli/cli/state_diff.py
```diff
@@ -442,6 +486,62 @@ def _normalize_routes(self, routes: set) -> set:
 
         return new_routes_set
 
+    def _normalize_gateways(self, config: dict) -> AbstractSet[NetplanRoute]:
```
slyon, line 489: thought: Wow, that's lots of code/logic for those deprecated `gateway4/6` fields.. But now that we have it we should keep it ;-)

netplan_cli/cli/state_diff.py
```diff
@@ -486,7 +591,7 @@ def _filter_system_routes(self, system_routes: AbstractSet[NetplanRoute], system
                 continue
 
             # Filter IPv6 local routes
-            if route.family == 10 and (route.to in local_networks or route.to in addresses):
+            if route.family == 10 and route.protocol != 'ra' and (route.to in local_networks or route.to in addresses):
```
slyon, line 594: Please use `socket.AF_INET6` instead.

tests/cli/test_state_diff.py
```diff
@@ -628,6 +628,326 @@ def test_diff_dhcp_addresses_are_filtered_out(self):
...
+                        "from": "10.86.126.253",
+                        "metric": 100,
+                        "type": "unicast",
+                        "scope": "global",
+                        "protocol": "dhcp",
+                        "table": "main"
+                    },
+                    {
+                        "to": "10.86.126.0/24",
+                        "family": 2,
```
slyon, line 671: nitpick: Consider using proper constants here and in the other tests below.

slyon (commented):
Thank you very much for those enhancements and quality of life improvements!

Looks mostly good to me. Please consider using named constants for the ip-address family instead of plain numbers.

Also, we should avoid introducing new public API for deprecated symbols (gateay4/6), still handling them as you did via internal/private API is fine!

tests/cli/test_state_diff.py
daniloegea: hm I think I'll leave these as numbers. This is copied from state dumps and they will be represented as numbers. But I totally agree we should use the constants in non-test code.

daniloegea:
Thank you, Lukas. I addressed your comments and tried to improve a little the analysis of dynamic configuration with networkd.

include/netdef.h
```diff
@@ -217,3 +227,5 @@ netplan_netdef_get_link_local_ipv6(const NetplanNetDefinition* netdef);
  */
 NETPLAN_PUBLIC ssize_t
 netplan_netdef_get_macaddress(const NetplanNetDefinition* netdef, char* out_buffer, size_t out_buffer_size);
+
+
```
slyon, line 231: nitpick (non-blocking): Unnecessary newlines.

tests/cli/test_state_diff.py
slyon: fair enough!

slyon (approved):
Thanks a lot for addressing my remarks! I like the idea of using networkd's `ConfigSource`. Handling similar cases through NetworkManager's DBus API in the future is a good approach!

I pushed a tiny commit to clean up some unnecessary newlines/whitespace in `include/netdef.h`. Otherwise, LGTM. Let's get this merged!

include/netdef.h
slyon: I fixed this in https://github.com/canonical/netplan/pull/466/commits/692f9023432ea35cc5e36485e7c5dd12ecd61b96

PR #468, opened 2024-05-08
nm/nd: fix a couple of crashes

daniloegea:
These problems happen when the ignore_errors flag is used.

Wifi: When parsing is interrupted before access points are added to the interface, the code that generates configuration for NM and networkd will crash. They assume the list is never empty. Add a check to see if the list is empty.

Wireguard: The same problem happens on networkd when the list of peers is empty. The NM code already checks it.

Add unit tests for these cases.

## Checklist


slyon (approved):
Thanks, lgtm!

PR #470, opened 2024-05-10
cli/generate: skip daemon_reload with --mapping

daniloegea:
When --mapping is used, nothing is generated so there is no need for calling daemon-reload.

## Checklist


slyon (approved):
lgtm

PR #471, opened 2024-05-21
networkd:apply: Drop handling of legacy wpa@ instance units

slyon:
## Description
networkd:apply: Drop handling of legacy wpa@ instance units
Those haven't been genearted anymore since 0.98. We can drop that legacy codepath now.

## Checklist


daniloegea (approved):
lgtm!

PR #482, opened 2024-06-24
networkd: make s-n-wait-online wait for at least one routable interface

slyon:
## Description
wait-online: wait for 'routable' state, if corresponding IPs are defined

Generally, we should be waiting on "routable" state (instead of "degraded") when we have static IP/dhcp4/dhcp6/RA.

[Our spec](https://discourse.ubuntu.com/t/spec-definition-of-an-online-system/27838) defines "at least one interface MUST be up on the link layer and have received layer 3 (IP) configuration" (ignoring link-local).

We now start two separate `systemd-networkd-wait-online` processes. One waiting for "carrier" & "degraded" interfaces (including the "routable" ones), not using "--any". Another one waiting just for "routable" interfaces, combined with "--any". Both blocking "network-online.target".

FR-7838

## Checklist

### DIFF
--- a/src/networkd.c
+++ b/src/networkd.c
@@ -88,17 +88,22 @@ _netplan_query_system_interfaces(GHashTable* tbl)
     }
 }
 
+typedef struct wait_online_data {
+    gboolean ignore_carrier;
+    gboolean degraded;
+    gboolean routable;
+} WaitOnlineData;
+
 /**
  * Enumerate all network interfaces (/sys/clas/net/...) and check
  * netplan_netdef_match_interface() to see if they match the current NetDef
  */
-STATIC gboolean
-_netplan_enumerate_interfaces(const NetplanNetDefinition* def, GHashTable* ifaces, GHashTable* tbl, const char* min_oper_state, const char* set_name, const char* rootdir)
+STATIC void
+_netplan_enumerate_interfaces(const NetplanNetDefinition* def, GHashTable* ifaces, GHashTable* tbl, const char* set_name, WaitOnlineData* data, const char* rootdir)
 {
     g_assert(ifaces);
     g_assert(tbl);
 
-    gboolean found_match = FALSE;
     GHashTableIter iter;
     gpointer key;
     g_hash_table_iter_init (&iter, ifaces);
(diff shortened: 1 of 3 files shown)


enr0n (approved):
Nice! LGTM.

tests/generator/test_args.py
```diff
@@ -186,7 +190,8 @@ def test_systemd_generator(self):
 
 [Service]
 ExecStart=
-ExecStart=/lib/systemd/systemd-networkd-wait-online -i eth99.43:degraded -i br0:degraded -i lo:carrier -i eth99.42:carrier\n''')
+ExecStart=/lib/systemd/systemd-networkd-wait-online -i lo:carrier -i eth99.42:carrier -i eth99.44:degraded
+ExecStart=/lib/systemd/systemd-networkd-wait-online --any -o routable -i eth99.43 -i br0\n''')
```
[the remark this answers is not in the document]
slyon: Now included in https://github.com/canonical/netplan/pull/482/commits/a119c1576ee2585e7faeebb217ed74139440a870

src/networkd.c
```diff
@@ -1546,58 +1550,40 @@ _netplan_networkd_write_wait_online(const NetplanState* np_state, const char* ro
          * come online.
          */
         if (!(def->optional || def->activation_mode)) {
+            WaitOnlineData* d = g_malloc0(sizeof(WaitOnlineData));
+            d->ignore_carrier = def->ignore_carrier;
+
             // Check if we have any IP configuration
             // bond and bridge members will never ask for link-local addresses (see above)
```
daniloegea, line 1557: thought: I wonder if we really should wait for all the bond members by default. One might have a bunch of the bond members disconnected for whatever reason and the system might still be online if at least one member is connected. I'd say it would be better to use `--any` for groups of interfaces that are members of the same bond. That would complicate the implementation though. I suppose we at least need to recommend setting all the members as `optional` (but not the bond itself) to avoid having anything blocking on the wait-online target...

daniloegea (approved):
It seems to be working as intended.

I just left one thought about bonds. I'm not sure if we should wait for carrier on all the members by default. The system doesn't need to have carrier on all the bond members to be online.

But to have it waiting for any member we'd need to create groups of members for each bond and have one call to wait-online --any for each group... maybe we could consider that in the future. Maybe we should recommend that bond members should be optional.

src/networkd.c
slyon: Good thought! I'm now issuing an INFO log message about it, at least.

slyon:
> I'm not sure if we should wait for carrier on all the members by default. The system doesn't need to have carrier on all the bond members to be online.

Right, I think this edge cases was not considered in the spec. But issuing a log message about it is probably a good first step. Further improvements can then be done as follow-up PRs.

PTAL.

src/networkd.c
```diff
@@ -1566,6 +1566,12 @@ _netplan_networkd_write_wait_online(const NetplanState* np_state, const char* ro
                               && !(netplan_netdef_get_bond_link(def) || netplan_netdef_get_bridge_link(def)));
             _netplan_address_iter_free(addr_iter);
 
+            // Not all bond members need to be connected (have carrier) for the parent to be ready
+            NetplanNetDefinition* bond_parent = netplan_netdef_get_bond_link(def);
+            if (bond_parent && !d->routable && !d->degraded) {
```
daniloegea, line 1571: suggestion: we could log this message only for members that are not already optional. It would also be nice to identify the interface in the message and that the user is seeing that due to the wait-online configuration.

For example, in my test environment I see the same log message 4 times as I have a bond with 4 interfaces.

slyon: The log message actually shouldn't trigger for optional bond members, as we have that condition above:

```C
if (!(def->optional || def->activation_mode)) {
```

But mentioning the bond-member name in the log message is a good idea. I'll also add a test for not logging optional interfaces.

daniloegea (approved):
lgtm

PR #483, opened 2024-06-25
bullet point removal

shirleyherox:
There was a single bullet point that divided a sentence. I removed it because I thought it was a typo.

## Checklist

### DIFF
--- a/doc/netplan-generate.md
+++ b/doc/netplan-generate.md
@@ -64,7 +64,7 @@ There are 3 locations that **`netplan generate`** considers:
 
 If there are multiple files with exactly the same name, then only one
 will be read. A file in `/run/netplan` will shadow - completely replace
-- a file with the same name in `/etc/netplan`. A file in `/etc/netplan`
+a file with the same name in `/etc/netplan`. A file in `/etc/netplan`
 will itself shadow a file in `/lib/netplan`.
 
 Or, in other words, `/run/netplan` is top priority, then `/etc/netplan`,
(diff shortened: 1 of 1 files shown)


doc/netplan-generate.md
```diff
@@ -64,7 +64,7 @@ There are 3 locations that **`netplan generate`** considers:
 
 If there are multiple files with exactly the same name, then only one
 will be read. A file in `/run/netplan` will shadow - completely replace
```
slyon, line 66:
```suggestion
will be read. A file in `/run/netplan` will shadow (completely replace)
```

slyon (changes requested):
Oh good catch!
I guess it was never meant to be a bullet point, but rather an emphasis on the "completely replace" words, removing just one of the dashes would break up that structure, so maybe we should completely change it, e.g. by putting the "(completely replace)" into parentheses?

WDYT? And thank you very much for your contribution to Netplan!

doc/netplan-generate.md
ashrimpnamedbob: That makes sense and seems easy to understand! I tried to commit your suggestion. I don't really know how to use github and am a complete noob. But that change seems good. I'm very happy if this was at all helpful.

slyon (approved):
Thanks for addressing my remarks. LGTM!

doc/netplan-generate.md
slyon: Oh yes, it was. Thanks a lot for your contribution to Netplan. And the commit went through and is now merged into Netplan's main branch!

PR #486, opened 2024-07-02
Don't escape certain non-ascii characters

daniloegea:
## Description

g_strescape will mess up with utf-8 characters. It happens because all the range from 0x7f to 0xff are escaped by default. The bytes in this range can be found in utf-8 code points so non-ascii strings will end being changed.

This PR adds all the range in the exception list.

For example, the file below:
```
network:
  renderer: NetworkManager
  wifis:
    wlan0:
      access-points:
        "áááÁÁÁ":
          password: "áéí"
        "normal":
          password: "á\né\tí\rÁ"
```
will become this (output from `netplan get`):
```
network:
  version: 2
  renderer: NetworkManager
  wifis:
    wlan0:
      access-points:
        "\\303\\241\\303\\241\\303\\241\\303\\201\\303\\201\\303\\201":
          auth:
            key-management: "psk"
            password: "\\303\\241\\303\\251\\303\\255"
        "normal":
          auth:
            key-management: "psk"
            password: "\\303\\241\\n\\303\\251\\t\\303\\255\\r\\303\\201"
```
With the new exceptions it will become:
```
network:
  version: 2
  renderer: NetworkManager
  wifis:
    wlan0:
      access-points:
        "normal":
          auth:
            key-management: "psk"
            password: "á\\né\\tí\\rÁ"
        "áááÁÁÁ":
          auth:
            key-management: "psk"
            password: "áéí"
```

## Checklist


src/parse.c
```diff
@@ -59,8 +59,20 @@
...
  * Escaping backslashes in the parser affects "netplan set" as it will always escape \'s from
  * the input and update YAMLs with all the \'s escaped again.
+ *
+ * g_strescape will also escape all the non-ascii chars in the range 0x7f-0xff. This affects non-ascii
+ * strings and will break utf-8 strings for example. So we add the entire range to the list of exceptions.
 */
-static char* STRESCAPE_EXCEPTIONS = "\"\\";
+static char* STRESCAPE_EXCEPTIONS =
+    "\"\\"
+    "\x7f\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90"
```
slyon, line 68: todo: Can you please give some more evidence which characters this affects specifically?

E.g. according to https://www.charset.org/utf-8 the \x7f is a "Control character for DELETE", which is probably something that we actually want to have escaped, for security reasons. There might be more control characters that might need to be excluded.

slyon (changes requested):
Looks fine overall, but we might whitelist some control characters that we don't want to have un-escaped (like DELETE). Can you please clarify on this?

And maybe also add a small unit-test around it, so that we can see the actual characters being used?

src/parse.c
daniloegea: Ah yes, I added the 0x7f because it's mentioned in the non-ascii range here https://docs.gtk.org/glib/func.strescape.html but IT IS in the ascii table. So it will not be part of any multi-byte utf-8 code point I suppose.

The rest of the range affects all the multi-byte characters.

daniloegea:
I removed 0x7f from the list and rebased. I also added a unittest using the wifi password.

slyon (approved):
Thanks, LGTM!

PR #488, opened 2024-07-04
generate: avoid calling 'udevadm control --reload' (LP: #1999178)

slyon:
## Description
The udev rules directories are monitored and re-loaded automatically with modern systemd-udevd. No need to manually reload them in the generator, causing side-effects.

We can still force-reload them when issuing `netplan generate` or `netplan apply` through the CLI, just to be on the safe side.

Replaces: https://github.com/canonical/netplan/pull/304

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#1999178

### DIFF
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -262,8 +262,7 @@ class NetplanApply(utils.NetplanCommand):
                                       stdout=subprocess.DEVNULL,
                                       stderr=subprocess.DEVNULL)
 
-        # Reloading of udev rules happens during 'netplan generate' already
-        # subprocess.check_call(['udevadm', 'control', '--reload-rules'])
+        subprocess.check_call(['udevadm', 'control', '--reload'])
         subprocess.check_call(['udevadm', 'trigger', '--action=move', '--subsystem-match=net', '--settle'])
 
         # apply any SR-IOV related changes, if applicable
--- a/netplan_cli/cli/commands/generate.py
+++ b/netplan_cli/cli/commands/generate.py
@@ -82,6 +82,10 @@ class NetplanGenerate(utils.NetplanCommand):
             argv += ['--mapping', self.mapping]
         logging.debug('command generate: running %s', argv)
         res = subprocess.call(argv)
+        try:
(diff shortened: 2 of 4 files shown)


slyon:
Another option might be calling `udevadm control --reload --resolve-names=late` in the generator. This should avoid it trying to resolve names during `systemctl daemon-reload`, but instead to it when the actual event fires.

slyon:
Thinking about it. I feel like it should be fine just dropping the `udevadm control --reload` from the genrator binary. There shouldn't be any race conditions. Any `.rules` or `.link` files generated are written at systemd's generator stage. The `systemd-udevd.service` starts only afterwards and reads all available rules, before listening for kernel uevens (or triggering pending uevents). So we shouldn't miss any rule exection on first-boot/reboot.

At runtime, we can do whatever we need, through the Python CLI.

I'm pondering about adopting the `autostart` autopkgtest, which does a reboot test, adding some udev settings (like offloading flags) to the physical `eth0 en*` management interface of the test runner (ugh, we shouldn't fiddle with that, but it's the only physical interface we have).

daniloegea:
The `--resolve-names` parameter seems to not be supported by `udevadm control`?

daniloegea:
The workaround suggested in https://bugs.launchpad.net/netplan/+bug/1999178/comments/17 doesn't seem to work unfortunately.

slyon:
We've used this in the Ubuntu devel series for a while now and didn't observe any fallout.
I feel relatively confident this works as expected and think it should be merged.

@daniloegea WDYT?

daniloegea:
It seems to be working fine.

The only thing I've found so far, that might not be related to this change, is that the integration test `TestNetworkManager.test_remove_virtual_interfaces` from the "scenarios" is consistently failing in my backport of "1.0.1-1ubuntu2" to Noble, which includes this change. I'm still testing it to be sure this change is not causing problems.

daniloegea (approved):
The test I've been investigating seems to be flaky on ppc64el. After running it a bunch of times it passed once. But it's passing on Oracular where this patch is already present. It seems it's not related to it, I'll go ahead and merge this PR.

PR #493, opened 2024-07-29
meson: Add 'testing' option

slyon:
## Description
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1071599

FR-8335

## Checklist

### DIFF
--- a/.github/workflows/debci.yml
+++ b/.github/workflows/debci.yml
@@ -16,7 +16,7 @@ on:
 jobs:
   lxc-debian-testing:
     # The type of runner that the job will run on
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
@@ -29,6 +29,7 @@ jobs:
       # it's needed (will be auto-loaded) by routing.test_vrf_basic
       - name: Install dependencies
         run: |
+          sudo add-apt-repository -y -n -u -s ppa:slyon/netplan-ci
           sudo apt update
           sudo apt install debci lxc lxc-templates debian-archive-keyring autopkgtest ubuntu-dev-tools devscripts linux-modules-extra-$(uname -r) #openvswitch-switch
       # See: https://discourse.ubuntu.com/t/containers-lxc/11526 (Apparmor section)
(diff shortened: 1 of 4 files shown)


.github/workflows/debci.yml
```diff
@@ -50,6 +51,8 @@ jobs:
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
           sed -i 's|rm debian/tmp/lib/netplan/generate|# DELETED|' debian/rules
+          sed -i 's|bytecompile=-1|bytecompile=-1 -Dtesting=false|' debian/rules
```
daniloegea, line 54: question: is this still needed? In the autopkgtest we already set use `--env=DEB_BUILD_OPTIONS=nocheck` and I saw that you already updated netplan.io on Debian. Well, I suppose it's here because the package wasn't published yet...

daniloegea (changes requested):
Thanks, it looks good.

I just have two suggestions:

- In src/meson.build, the block `libnetplan_testing` should also be enabled only when `testing` is enabled as it's only used by the ctests. This should probably be addressed before merging.
- In the netplan.io package, `libcmocka-dev` should probably also be set with `<!nocheck>` now.

meson.build
```diff
@@ -70,77 +70,78 @@ test_env = [
...
-     timeout: 900,
-     env: test_env)
-#TODO: split out dbus tests into own test() instance, to run in parallel
-test('unit-tests',
-     pycoverage,
-     args: ['run', '-a', '-m', 'pytest', '-s', '-v', '--cov-append', meson.current_source_dir()],
-     timeout: 600,
-     env: test_env)
+if get_option('testing')
+    if get_option('unit_testing')
```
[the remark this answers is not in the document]
slyon: ACK. https://salsa.debian.org/debian/netplan.io/-/commit/866427b0e65d40b77531d4c6afdf0d3ada5e94f0

.github/workflows/debci.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       # it's needed (will be auto-loaded) by routing.test_vrf_basic
       - name: Install dependencies
         run: |
+          sudo add-apt-repository -y -n -u -s ppa:slyon/netplan-ci
```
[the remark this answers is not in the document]
slyon: Yes, exactly. This will be obsolete as soon as the `autopkgtest` package `5.37ubuntu3~24.04.1` migrates from `noble-proposed` to `noble-updates`. But the CI-PPA might still be useful for other stuff in the future..

.github/workflows/debci.yml
slyon: Yes. The new `autopkgtest` also uses `nocheck` by default. Now that those changes are pulled from the Debian package `1.0.1-2` we don't need the changes here, anymore.

slyon:
Thanks, I addressed all of your remarks. PTAL.

.github/workflows/debci.yml
slyon: Oh well, I guess we need to actually drop the `--env=DEB_BUILD_OPTIONS=nocheck `, using the new `autopkgtest` package. Otherwise, the `filter` in `debian/rules` will not match the double `nocheck nocheck` of `DEB_BUILD_OPTIONS`

daniloegea (approved):
lgtm!

PR #496, opened 2024-08-06
Handle implicity conversions

daniloegea:
## Description
Fix some type conversions and make some implicit conversions explicit.
It's a requirement to add the TIOBE analysis to our CI as it will build the project with `-Wconversion`.

I'd recommend to use interactive rebase to step through each commit individually starting from the first one and call `meson compile` with `-j1` so you can see what compiler errors each commit fixes. I didn't want to paste the error to the commit messages...

## Checklist

### DIFF
--- a/src/parse-nm.c
+++ b/src/parse-nm.c
@@ -138,7 +138,7 @@ keyfile_handle_generic_uint(GKeyFile* kf, const gchar* group, const gchar* key, 
 {
     g_assert(dataptr != NULL);
     if (g_key_file_has_key(kf, group, key, NULL)) {
-        guint data = g_key_file_get_uint64(kf, group, key, NULL);
+        guint data = (guint)g_key_file_get_uint64(kf, group, key, NULL);
         if (data != default_value)
             *dataptr = data;
         _kf_clear_key(kf, group, key);
@@ -262,7 +262,7 @@ parse_routes(GKeyFile* kf, const gchar* group, GArray** routes_arr)
 
         /* Append metric */
         if (split[0] && split[1] && split[2] && strtoul(split[2], NULL, 10) != NETPLAN_METRIC_UNSPEC)
-            route->metric = strtoul(split[2], NULL, 10);
+            route->metric = (guint)strtoul(split[2], NULL, 10);
         g_strfreev(split);
 
         /* Parse route options */
@@ -275,17 +275,17 @@ parse_routes(GKeyFile* kf, const gchar* group, GArray** routes_arr)
                 if (g_strcmp0(kv[0], "onlink") == 0)
                     route->onlink = (g_strcmp0(kv[1], "true") == 0);
                 else if (g_strcmp0(kv[0], "initrwnd") == 0)
-                    route->advertised_receive_window = strtoul(kv[1], NULL, 10);
+                    route->advertised_receive_window = (guint)strtoul(kv[1], NULL, 10);
                 else if (g_strcmp0(kv[0], "initcwnd") == 0)
-                    route->congestion_window = strtoul(kv[1], NULL, 10);
+                    route->congestion_window = (guint)strtoul(kv[1], NULL, 10);
                 else if (g_strcmp0(kv[0], "mtu") == 0)
-                    route->mtubytes = strtoul(kv[1], NULL, 10);
+                    route->mtubytes = (guint)strtoul(kv[1], NULL, 10);
                 else if (g_strcmp0(kv[0], "table") == 0)
-                    route->table = strtoul(kv[1], NULL, 10);
+                    route->table = (guint)strtoul(kv[1], NULL, 10);
                 else if (g_strcmp0(kv[0], "src") == 0)
                     route->from = g_strdup(kv[1]); //no need to free, will stay in netdef
                 else if (g_strcmp0(kv[0], "advmss") == 0)
-                    route->advmss = strtoul(kv[1], NULL, 10);
+                    route->advmss = (guint)strtoul(kv[1], NULL, 10);
                 else
                     unhandled_data = TRUE;
                 g_strfreev(kv);
@@ -505,10 +505,10 @@ parse_tunnels(GKeyFile* kf, NetplanNetDefinition* nd)
         _kf_clear_key(kf, "wireguard", "private-key");
 
         /* Reading the listen port */
-        nd->tunnel.port = g_key_file_get_uint64(kf, "wireguard", "listen-port", NULL);
+        nd->tunnel.port = (guint)g_key_file_get_uint64(kf, "wireguard", "listen-port", NULL);
         _kf_clear_key(kf, "wireguard", "listen-port");
 
-        nd->tunnel_private_key_flags = g_key_file_get_integer(kf, "wireguard", "private-key-flags", NULL);
+        nd->tunnel_private_key_flags = (guint)g_key_file_get_uint64(kf, "wireguard", "private-key-flags", NULL);
         _kf_clear_key(kf, "wireguard", "private-key-flags");
 
         gchar** keyfile_groups = g_key_file_get_groups(kf, NULL);
@@ -590,7 +590,7 @@ parse_tunnels(GKeyFile* kf, NetplanNetDefinition* nd)
         reset_vxlan(nd->vxlan);
 
         /* Reading the VXLAN ID*/
-        nd->vxlan->vni = g_key_file_get_integer(kf, "vxlan", "id", NULL);
+        nd->vxlan->vni = (guint)g_key_file_get_uint64(kf, "vxlan", "id", NULL);
         _kf_clear_key(kf, "vxlan", "id");
 
         nd->tunnel.local_ip = g_key_file_get_string(kf, "vxlan", "local", NULL);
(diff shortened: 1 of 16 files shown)


src/parse.c
```diff
@@ -2103,9 +2103,10 @@ handle_bridge_path_cost(NetplanParser* npp, yaml_node_t* node, const char* key_p
             if (*endptr != '\0')
                 return yaml_error(npp, node, error, "invalid unsigned int value '%s'", scalar(value));
 
-            g_debug("%s: adding path '%s' of cost: %d", npp->current.netdef->id, scalar(key), v);
+            g_debug("%s: adding path '%s' of cost: %ld", npp->current.netdef->id, scalar(key), v);
```
slyon, line 2106: question: I wonder if we should be using some marcos from `<inttypes.h>` here instead (e.g. `PRIu64`), as `%ld` is not always guaranteed to match uint64.

I guess we should also try (cross-)compiling this branch for armhf and s390x to see if there are any side-effects.

src/parse.c
```diff
@@ -3539,7 +3541,7 @@ process_document(NetplanParser* npp, GError** error)
 {
     gboolean ret;
     int previously_found;
-    int still_missing;
+    guint still_missing;
```
slyon, line 3544: note: this is due to `g_hash_table_size` returning a `guint`. For consistency's sake, we might transform `previously_found` and other related numbers to `guint`, too. WDYT?

tests/ctests/test_netplan_parser.c
```diff
@@ -418,7 +418,7 @@ test_parse_utf8_characters(__unused void** state)
 
     netplan_state_dump_yaml(np_state, fd, NULL);
 
-    size = lseek(fd, 0, SEEK_CUR) + 1;
+    size = (size_t)lseek(fd, 0, SEEK_CUR) + 1;
```
slyon, line 421: note (non-blocking): `lseek()` returns `off_t` which is a signed int, that should easily fit into a `size_t` type.

tests/ctests/test_netplan_parser.c
```diff
@@ -389,8 +389,8 @@ test_parse_utf8_characters(__unused void** state)
 {
     NetplanState *np_state = NULL;
     int fd;
-    int size;
-    int res;
+    size_t size;
+    long res;
```
slyon, line 393: todo: `res = read(...)` returns a `ssize_t`, why are we defining it as a `long` up here?

src/yaml-helpers.h
```diff
@@ -41,7 +41,7 @@
 }
 #define YAML_SCALAR_PLAIN(event_ptr, emitter_ptr, scalar) \
 { \
-    yaml_scalar_event_initialize(event_ptr, NULL, (yaml_char_t *)YAML_STR_TAG, (yaml_char_t *)scalar, strlen(scalar), 1, 0, YAML_PLAIN_SCALAR_STYLE); \
+    yaml_scalar_event_initialize(event_ptr, NULL, (yaml_char_t *)YAML_STR_TAG, (yaml_char_t *)scalar, (int)strlen(scalar), 1, 0, YAML_PLAIN_SCALAR_STYLE); \
```
slyon, line 44: question: Do we need some assert/checks to make sure long strings (`size_t`) do not overflow our `int` datatype?

src/util.c
```diff
@@ -648,15 +648,17 @@ netplan_get_id_from_nm_filepath(const char* filename, const char* ssid, char* ou
 
     /* Move pointer to start of netplan ID inside filename string */
     start = pos + strlen(nm_prefix);
-    id_len = end - start;
+    g_assert(end - start > 0);
+    id_len = (gsize)(end - start);
```
slyon, line 652: nitpick: If we do an explicit typecast anyway. Should we declare `id_len` as `ssize_t` directly above?

slyon (changes requested):
Thank you very much for working on this, this is tedious work!

I left a bunch of remarks inline that need clarification. Furthermore, I think we should try compiling this branch on some additional architectures (i386, armhf, s390x, maybe more from [Debian buildd](https://buildd.debian.org/status/package.php?p=netplan.io)?) to make sure it does not trigger any side-effects of types having different size on non-64bit architectures.

src/parse.c
daniloegea: good point, I'll create a package from this branch and test it against all the archs supported by Ubuntu

src/parse.c
```diff
@@ -2343,7 +2345,7 @@ handle_arp_ip_targets(NetplanParser* npp, yaml_node_t* node, __unused const void
 
     /* Avoid adding the same arp_ip_targets in a 2nd parsing pass by comparing
      * the array size to the YAML sequence size. Skip if they are equal. */
-    guint item_count = node->data.sequence.items.top - node->data.sequence.items.start;
+    long item_count = node->data.sequence.items.top - node->data.sequence.items.start;
```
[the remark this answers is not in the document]
daniloegea: `top` and `start` are pointers, so `top - start` produces a signed long int. A signed long can be 64bit-long. But I suppose it would be better to use `ptrdiff_t` here.

src/parse-nm.c
```diff
@@ -855,7 +855,7 @@ netplan_parser_load_keyfile(NetplanParser* npp, const char* filename, GError** e
     /* Wifis */
     if (g_key_file_has_group(kf, "wifi")) {
         if (g_key_file_get_uint64(kf, "wifi", "wake-on-wlan", NULL)) {
-            nd->wowlan = g_key_file_get_uint64(kf, "wifi", "wake-on-wlan", NULL);
+            nd->wowlan = g_key_file_get_integer(kf, "wifi", "wake-on-wlan", NULL);
```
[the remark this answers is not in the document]
daniloegea: The problem is that `nd->wowlan` is a `gint`. That's why I used `g_key_file_get_integer`. But as the property is an `uint32` it's indeed better to continue to use `_get_uint64` with a cast but add an assertion before.

src/util.c
daniloegea: Technically `end - start` is a `ptrdiff_t` (`signed long`). Changing the type of `id_len` causes conversion issues in other lines. As I added the assertion before, I think it's ok to leave the cast.

tests/ctests/test_netplan_parser.c
daniloegea: `off_t` can be a `long int`, see below:

```
../tests/ctests/test_netplan_parser.c:421:12: error: conversion to ‘size_t’ {aka ‘long unsigned int’} from ‘__off64_t’ {aka ‘long int’} may change the sign of the result [-Werror=sign-conversion]
```

In this particular case it's lseeking to zero so it's safe do to the cast.

daniloegea: Unless of course lseek returns an error (-1)

src/parse.c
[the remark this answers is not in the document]
daniloegea: it's still a problem in armhf. `ptrdiff_t` is an `int` there.

```
../src/parse.c: In function ‘handle_arp_ip_targets’:
../src/parse.c:2350:62: error: comparison of integer expressions of different signedness: ‘guint’ {aka ‘unsigned int’} and ‘ptrdiff_t’ {aka ‘int’} [-Werror=sign-compare]
 2350 |     if (npp->current.netdef->bond_params.arp_ip_targets->len == item_count) {
      |                                                              ^~
../src/parse.c: In function ‘handle_wireguard_peers’:
../src/parse.c:2696:51: error: comparison of integer expressions of different signedness: ‘guint’ {aka ‘unsigned int’} and ‘ptrdiff_t’ {aka ‘int’} [-Werror=sign-compare]
 2696 |     if (npp->current.netdef->wireguard_peers->len == item_count) {
      |                                                   ^~
cc1: all warnings being treated as errors
```

src/util.c
daniloegea: correction, `ptrdiff_t` is a `long` on 64bit archs, on 32bit (tested on armhf) it's an int.

daniloegea:
Thanks Lukas. I think I addressed your comments.

In particular, there was a problem with 32-bit. It happened on armhf. I added the compiler error to one of my replies.

I created a package from this branch and tested it with autopkgtest on all the archs supported by Ubuntu and it looks good.

src/yaml-helpers.h
```diff
@@ -41,7 +42,9 @@
 }
 #define YAML_SCALAR_PLAIN(event_ptr, emitter_ptr, scalar) \
 { \
-    yaml_scalar_event_initialize(event_ptr, NULL, (yaml_char_t *)YAML_STR_TAG, (yaml_char_t *)scalar, strlen(scalar), 1, 0, YAML_PLAIN_SCALAR_STYLE); \
+    size_t _length = strlen(scalar); \
+    g_assert(_length < G_MAXINT); \
+    yaml_scalar_event_initialize(event_ptr, NULL, (yaml_char_t *)YAML_STR_TAG, (yaml_char_t *)scalar, (int)_length, 1, 0, YAML_PLAIN_SCALAR_STYLE); \
```
slyon, line 47: nitpick (non-blocking): I guess we could also just pass `-1` as the `length` parameter and letting `libyaml` deal with it itself, avoiding any extra calls. But it's fine as-is, too.

See: https://github.com/yaml/libyaml/pull/102

src/yaml-helpers.h
slyon: nitpick (non-blocking): I guess we could also just pass `-1` as the `length` parameter and letting `libyaml` deal with it itself, avoiding any extra calls. But it's fine as-is, too.

See: https://github.com/yaml/libyaml/pull/102

slyon (approved):
Thanks a lot for addressing my remarks, the deep investigations and running those extra cross-architecture tests!

LGTM! I left one tiny inline comment, but I think it's fine to be merged as-is.

src/yaml-helpers.h
daniloegea: hm I see. That seems to be an undocumented behavior though? The doc string doesn't mention it https://github.com/yaml/libyaml/blob/0.2.5/include/yaml.h#L580

slyon: Yeah, I though about that, too. So we might be safer keeping our code as-is.

PR #497, opened 2024-08-08
Change default umask when creating dirctories (LP: #2076319)

rmalz-c:
## Description
fixes: lp2076319

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2076319

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <sys/mman.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


rmalz-c:
Added proper description in commit msg

slyon (approved):

The change LGTM and indeed resembles what we used to do in previous versions of Netplan. I agree with Danilo that (for now) it is safer to change `umask` instead of querying users/groups from inside the generator.

PR #501, opened 2024-08-13
ci: run meson tests with unbuffer

daniloegea:
Fix the job name and disable the scheduler for now so the job can be triggered manually [1].

[1] - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

## Checklist


slyon (approved):
lgtm

PR #503, opened 2024-08-13
cli: drop python-netifaces (LP: #2065870, LP: #2017585)

daniloegea:
netifaces doesn't return the permanent MAC address of physical interfaces.

This causes a problem with bonds for example where all the interfaces in the bond will share the same MAC address. When `netplan apply` tries to find a unique matching based on the MAC returned by netifaces and the MAC set in the YAML, it will not be found.

```
# netplan apply
[]
Cannot find unique matching interface for nic0
```

The `is_composite_member()` method wasn't working properly as well. It was always returning False. In the end I stopped using it.

Regarding skipping bond/bridge members during renames, that seems to be put in place as a workaround for [LP: #1802322](https://bugs.launchpad.net/netplan/+bug/1802322). The problem it fixed was happening because the code was matching interfaces by their non-permanent MAC address. When they are added to a bond, they will share the same MAC. The code was confusing one interface with another and trying to rename them.

By not renaming them the backend will "forget" about the interface. We will emit the new name to the backend config files but keep the interface with the old name in the system so it will not be found anymore. This is worse than causing a quick disconnection during the rename process as further changes to the interface will not be applied. The user would need to either rename it manually (which will cause a disconnection anyway) or reboot. I also found that networkd <= 255 was triggering a rename anyway when we called `udevadm test` from `netplan apply` so we've been unintentionally allowing renames anyway...

Fixes: [LP#2065870](https://bugs.launchpad.net/netplan/+bug/2065870)

## Checklist

- [ ] \(Optional\) Closes an open bug in Launchpad. LP#2065870, LP#2017585

### DIFF
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -331,23 +331,6 @@ class NetplanApply(utils.NetplanCommand):
                 utils.nm_bring_interface_up(loopback_connection)
 
     @staticmethod
-    def is_composite_member(composites, phy):
-        """
-        Is this physical interface a member of a 'composite' virtual
-        interface? (bond, bridge)
-        """
-        for composite in composites:
-            for _, settings in composite.items():
-                if not type(settings) is dict:
-                    continue
-                members = settings.get('interfaces', [])
-                for iface in members:
-                    if iface == phy:
-                        return True
-
-        return False
-
-    @staticmethod
     def clear_virtual_links(prev_links, curr_links, devices=[]):
         """
         Calculate the delta of virtual links. And remove the links that were
@@ -381,7 +364,6 @@ class NetplanApply(utils.NetplanCommand):
         """
 
         changes = {}
-        composite_interfaces = [config_manager.bridges, config_manager.bonds]
 
         # Find physical interfaces which need a rename
         # But do not rename virtual interfaces
@@ -391,11 +373,6 @@ class NetplanApply(utils.NetplanCommand):
                 continue  # Skip if no new name needs to be set
             if not netdef._has_match:
                 continue  # Skip if no match for current name is given
-            if NetplanApply.is_composite_member(composite_interfaces, netdef.id):
-                logging.debug('Skipping composite member {}'.format(netdef.id))
-                # do not rename members of virtual devices. MAC addresses
-                # may be the same for all interface members.
-                continue
             # Find current name of the interface, according to match conditions and globs (name, mac, driver)
             current_iface_name = utils.find_matching_iface(interfaces, netdef)
             if not current_iface_name:
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -27,7 +27,6 @@ import subprocess
 import shutil
 import tempfile
 import filecmp
-import netifaces
 import time
 
 from .. import utils
@@ -117,7 +116,7 @@ class NetplanApply(utils.NetplanCommand):
             old_ovs_glob.remove(ovs_cleanup_service)
         old_files_ovs = bool(old_ovs_glob)
         old_nm_glob = glob.glob('/run/NetworkManager/system-connections/netplan-*')
-        nm_ifaces = utils.nm_interfaces(old_nm_glob, netifaces.interfaces())
+        nm_ifaces = utils.nm_interfaces(old_nm_glob, utils.get_interfaces())
         old_files_nm = bool(old_nm_glob)
 
         restart_networkd = False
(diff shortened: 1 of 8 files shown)


slyon:
see also https://bugs.launchpad.net/netplan/+bug/2017585

netplan_cli/cli/commands/apply.py
```diff
@@ -332,21 +333,19 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
...
-                for iface in members:
-                    if iface == phy:
-                        return True
 
-        return False
+        # Should this be extended to other types of parent interfaces (vlans, vrfs)
+        bond_link = netdef.links.get('bond')
+        bridge_link = netdef.links.get('bridge')
+
+        is_composite = bool(bond_link) or bool(bridge_link)
```
slyon, line 346: question: what about VRFs? Those can also have member devices.

netplan_cli/cli/commands/apply.py
```diff
@@ -332,21 +333,19 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
                 utils.nm_bring_interface_up(loopback_connection)
 
     @staticmethod
-    def is_composite_member(composites, phy):
+    def is_composite_member(netdef: NetDefinition) -> bool:
```
slyon, line 336: todo: looks like this method is not used any more (besides in its own tests). Let's drop it then?

netplan_cli/cli/utils.py
```diff
@@ -196,10 +196,43 @@ def get_interface_driver_name(interface, only_down=False):  # pragma: nocover (c
...
 
 
-def get_interface_macaddress(interface):
-    # return an empty list (and string) if no LL data can be found
-    link = netifaces.ifaddresses(interface).get(netifaces.AF_LINK, [{}])[0]
-    return link.get('addr', '')
+def _get_permanent_macaddress(interface: str) -> str:
+    mac = ""
+    try:
+        out = subprocess.check_output(['ethtool', '-P', interface]).decode('utf-8')
```
slyon, line 202: note (non-blocking): We need a new runtime dependency on `ethtool` to make this work. (It's already in Ubuntu "main")

Could you please temporarily add it to the CI test, so we can see if this changes behaviour in any way? For now it would just be falling back to non-permanent MAC in case `ethtool` is not installed. – Which is fine, IMO, but we want to test the real thing in the integration tests.

netplan_cli/cli/utils.py
```diff
@@ -196,10 +196,43 @@ def get_interface_driver_name(interface, only_down=False):  # pragma: nocover (c
...
+        split = out.split(': ')
+        if len(split) == 2 and split[1].strip() != 'not set':
```
slyon, line 204: nitpick: are we sure the "not set" string might not change in the future, or introduce other options?

Maybe it would be safer to actually use `cli.utils.is_valid_macaddress` to validate the output.

netplan_cli/cli/utils.py
```diff
@@ -196,10 +196,43 @@ def get_interface_driver_name(interface, only_down=False):  # pragma: nocover (c
...
+
+    return mac
+
+
+def _get_macaddress(interface: str) -> str:
+    try:
+        with open(f'/sys/class/net/{interface}/address') as f:
+            return f.read().strip()
+    except Exception:
+        return ""
```
slyon, line 217: question (non-blocking): Is it essential to return the empty string here? If so, why? Otherwise, I'd prefer returning `None` in the error case, as I think that makes the intention a bit more clear.

tests/cli/test_units.py
```diff
@@ -43,19 +46,52 @@ def tearDown(self):
         shutil.rmtree(self.tmproot)
 
     def test_is_composite_member(self):
```
slyon, line 48: we might want to drop this all together. `is_composite_member` is not part of any public API and not used any more within the CLI

slyon (commented):
Thank you very much. I generally like those changes, especially as `netifaces` is not maintained upstream anymore: https://github.com/al45tair/netifaces/issues/78

I left a bunch of inline comments, some asking for some additional work (nothing big).

Also note, that after this change Netplan grows a new runtime dependency on `ethtool`.

netplan_cli/cli/utils.py
daniloegea: The only reason is because I annotated the function with `-> str`. I've been trying to avoid this typing confusion in Python (None is not a str). I will adopt `Optional[str]` here for the type hint and return None.

netplan_cli/cli/utils.py
daniloegea: ethtool is seeded on Ubuntu actually. We might need to install it for the Debian test in our CI though.

netplan_cli/cli/commands/apply.py
daniloegea: in the end this function was removed

daniloegea:
Thanks, Lukas. I addressed your comments and added some extra unit tests.

netplan_cli/cli/utils.py
slyon: https://salsa.debian.org/debian/netplan.io/-/commit/d6405858d585b833eb5d1c58cf43eff8568a16ed

PR #508, opened 2024-08-20
ci/tics: set the token before trying to download TICS

daniloegea:
The download started to fail without authentication.

I tested it locally and the script `install_tics.sh` works when the variable is exported.

## Checklist


slyon (approved):
+1

PR #510, opened 2024-08-28
CI: cleanup after v1.1 release

slyon:
## Description
* Dropping some quirks that are not needed anymore as of the v1.1 packaging.
* Re-enable testing for network devices that need additional kernel modules (like `vrf` or `wireguard`) in containers.

## Checklist

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -48,8 +48,6 @@ jobs:
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
           sed -i 's|iproute2,|iproute2, ethtool,|' debian/control  # add ethtool as a dependency of netplan.io temporarily
-          sed -i 's|  python3-gi,|  python3-gi, python3-packaging,|' debian/tests/control  # needed for the 'routing' test (nm_version)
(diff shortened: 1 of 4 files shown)


.github/workflows/autopkgtest.yml
```diff
@@ -48,8 +48,6 @@ jobs:
           cp -r netplan.io-*/debian .
           rm -r debian/patches/  # clear any distro patches
           sed -i 's|iproute2,|iproute2, ethtool,|' debian/control  # add ethtool as a dependency of netplan.io temporarily
-          sed -i 's|  python3-gi,|  python3-gi, python3-packaging,|' debian/tests/control  # needed for the 'routing' test (nm_version)
```
daniloegea, line 51: I'm not sure why python3-packaging was needed in the first place and I'm not sure why it can be removed now :D well it seems that it's not needed for tests anyway...

daniloegea (approved):
lgtm

.github/workflows/autopkgtest.yml
slyon: Sorry I should have been more clear. Context is given in https://salsa.debian.org/debian/netplan.io/-/commit/9233d668db581ae66f1739ddb35ea0a3b5e2db83

In essence, we need it as part of our autopkgtests for https://github.com/canonical/netplan/commit/b13b518 but it's integrated in the distro packages by now, so we don't need the quirk in our CI anymore.

PR #509, opened 2024-08-20
tests: ethernets: Add ipv6-address-generation integration tests

tatokis:
## Description

These tests validate the Stable Privacy and Modified EUI-64 IPv6 address generation methods under both NetworkManager and systemd-networkd.

## Checklist

- [x] Runs `make check` successfully.
- [x] Retains 100% code coverage (`make check-coverage`).
- [ ] New/changed keys in YAML format are documented.
- [ ] \(Optional\) Adds example YAML for new feature.
- [ ] \(Optional\) Closes an open bug in Launchpad.

PR #514, opened 2024-09-03
tests:ethernets: avoid flaky test_ip6_eui64 results

slyon:
## Description
Avoid flaky `test_ip6_eui64` results, when statefull DHCPv6 is quicker than stateless EUI-64 address generation.
Also use a `ra-stateless` configuration option to dnsmasq, to avoid providing stateful DHCP addresses to the client;
`ra-only` doesn't seem to work properly, when combined with `--dhcp-range` (from `integration/base.py`).

This seems to fail from time to time on systemd v256 based systems, such as Debian testing or Ubuntu oracular.

Follow-up to https://github.com/canonical/netplan/pull/509

## Checklist

### DIFF
--- a/tests/integration/ethernets.py
+++ b/tests/integration/ethernets.py
@@ -223,7 +223,7 @@ class _CommonTests():
         self.assert_iface_up(self.dev_e_client, ['inet6 2600::42/64'])
 
     def test_ip6_stable_privacy(self):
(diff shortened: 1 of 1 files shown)


slyon:
I've tested this locally multiple times in a Oracular LXD container (where it failed previously). Looking all good. The CI is green, too. Merging...

PR #515, opened 2024-09-03
[RFC] CI: Ignore errors off by default

slyon:
## Description
CI:tests: Disable NETPLAN_PARSER_IGNORE_ERRORS by default
    
It's taking lots of compute time in upstream and downstream CI systems.
We still keep it on inside our `autopkgtest.yml` CI test, but turn it off
by default otherwise.

## Checklist

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -56,5 +56,7 @@ jobs:
           dch -v "$VER" "Autopkgtest CI testing (Noble)"
       - name: Run autopkgtest (incl. build)
         run: |
-          autopkgtest . \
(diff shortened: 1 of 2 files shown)


slyon:
@daniloegea How do you feel about this? Is it necessary to have the duplicated tests run in multiple upstream CI, downstream autopkgtests and Salsa CI? Or is just a single, upstream CI test enough?

daniloegea:
I think it's enough to have it enabled only in our Github CI for autopkgtests. Arguably it should be enough to have it only in the config-fuzzing workflow but I'd feel more comfortable with the autopkgtest too.

daniloegea (approved):
lgtm

PR #516, opened 2024-09-10
Use more permissive umask for .service units (LP: #2072486)

slyon:
## Description
networkd:openvswitch:sriov: Permissive umask for .service units (LP#2072486)
    
Writing world-inaccessible config files doesn't help when the data can still be accessed by any local user through the systemd APIs, e.g.: `systemctl show UNIT_NAME.service`

Using `640` permission on such .service unit files provides a false illusion of security and leads to systemd warnings in the journal, like this:
```systemd[1]: Configuration file /run/systemd/system/netplan-ovs-cleanup.service is marked world-inaccessible. This has no effect as configuration data is accessible via APIs without restrictions. Proceeding anyway.```

Also:
* Update tests according
* Update doc/tread model accordingly

FR-8238

## Checklist

- [x] \(Optional\) Closes an open bug in Launchpad. LP#2072486

### DIFF
--- a/doc/security.md
+++ b/doc/security.md
@@ -19,6 +19,23 @@ Security advice: ensure all YAML files in `/etc/netplan`, `/run/netplan` and
 `/lib/netplan` are not readable by non-privileged users.
 :::
 
+## Systemd `.service` units
+
+Netplan generates many systemd `.service` units, which are world-accessible to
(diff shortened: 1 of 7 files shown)


daniloegea (approved):
Looks good to me.

PR #521, opened 2024-09-25
Handling encoding issues on netplan status

daniloegea:
## Description
This PR addresses some issues with `netplan status`:

1) SSID is not displayed with newer systemd
On Ubuntu Noble for example, `netplan status` will not show the SSID for wifi interfaces.
The problem was likely caused by https://github.com/systemd/systemd/commit/8fff78a1dded105e1ee87bc66e29ef2fd61bf8c9.

2) The output of external programs was broken if non-ascii characters are present.

3) Netdef IDs were not being extracted correctly from file names when non-ascii characters were escaped.

4) Retrieving the SSID via `networkctl` wasn't producing the readable SSID string.
To verify the enigmatic trick used in the commit `state/SSID: handle non-ascii SSIDs`:

```python
ssid = '\\303\\241\\303\\251\\303\\255\\303\\263\\303\\272\\302\\242\\302\\242\\302\\242\\302\\243\\302\\243\\302\\243'
ssid2 = ssid.encode('latin1').decode('unicode-escape').encode('latin1').decode('utf-8')
print(ssid2)
```
The output should be `áéíóú¢¢¢£££`.

The line 1 above is what we receive from networkctl and we are currently displaying it as the SSID in `netplan status`.

5) On some devices `networkctl` and `iw` will not retrieve the SSID and `"(null)"` will be displayed. This problem was fixed for Network Manager connections (where the SSID is retrieved via nmcli) but is still there when networkd is the backend.
Example: `●  3: wlp2s0 wifi/"(null)" UP (NetworkManager: wlp2s0)`

## Checklist


netplan_cli/cli/state.py
```diff
@@ -59,6 +60,9 @@
     'vrf': 'vrf',
     'vxlan': 'tunnel',
 
+    # Used for wifi testing
+    'ieee80211_radiotap': 'wifi',
```
slyon, line 64: question: Could you please provide some instruction of how this can be used for wifi testing? Either as part of the PR description or a code comment. I cannot see it being used in the automated tests.

netplan_cli/cli/state.py
```diff
@@ -328,13 +348,22 @@ def vendor(self) -> str:
...
+            #       Sometimes it will return "(null)".
             for line in self._networkctl.splitlines():
                 line = line.strip()
-                key = 'WiFi access point: '
-                if line.startswith(key):
-                    ssid = line[len(key):-len(' (xB:SS:ID:xx:xx:xx)')].strip()
+                key = r'^Wi-?Fi access point: (.*) \(.*\)'
+                if match := re.match(key, line):
+                    ssid = match.group(1)
+                    # TODO: Find a better way to retrieve the SSID
```
slyon, line 362: nitpick (non-blocking): We should try getting this information from the `networkctl --json=short` output instead.

IIRC, this was not implemented back when we first wanted to use it in `netplan status`, but I can see it on systems using systemd v255+ today: `networkctl --json=short status wlp0s20f3 | jq | grep SSID`

slyon (approved):
Thanks, LGTM. With a few (non-critical) inline comments.

netplan_cli/cli/state.py
daniloegea: Yes, sorry. This is the device type for the device `hwsim0` created by the `mac80211_hwsim` module.
I used it to test my changes and noticed warnings from `netplan status`.

This is how `networkctl` sees it: `7 hwsim0 ieee80211_radiotap off         unmanaged`

I added it to the list to avoid this warning from `netplan status`:

```
~# netplan status
Unknown device type: ieee80211_radiotap
Unknown device type: ieee80211_radiotap
     Online state: online
...
```

I'll improve the comment.

netplan_cli/cli/state.py
daniloegea: Sometimes it's not available. The output below is from one of my laptops:

```
danilo@danilo-Aspire-ES1-572:~$ networkctl --json=short status wlp2s0 | jq | grep SSID
  "BSSID": [
danilo@danilo-Aspire-ES1-572:~$
```

On another laptop it will give me the SSID though. I suspect it might be some limitation of the driver, I don't really know.
That's the reason why I changed it to get it from Network Manager when NM is the backend.

daniloegea: And in these cases we will show it like this on `netplan status`:

```
●  5: wlan2 wifi/"(null)" UP (networkd: wlan2)
```

PR #193, opened 2021-02-24
libnetplan: provide API for NetworkManager YAML backend

slyon:
## Description
* fix `match.original_name` vs `netdef->id` shortcut (it shall be the same value, but not same reference/pointer)
* Tracking of filename per netdef in the YAML parser
* Improve wake-on-lan handling
* Use NM aliases for `wifi`/`ethernet` keyfile groups
* Refactoring NetworkManager keyfile generator to use GKeyFile instead of GString
* Adding a NetworkManager keyfile parser
* Adding a netplan YAML generator

New APIs introduced:
```
// YAML generator
void write_netplan_conf(const NetplanNetDefinition* def, const char* rootdir);
// NM keyfile parser
gboolean netplan_parse_keyfile(const char* filename, GError** error);
// general & utils
guint netplan_clear_netdefs();
gboolean netplan_delete_connection(const char* id, const char* rootdir);
gboolean netplan_generate(const char* rootdir);
gchar* netplan_get_id_from_nm_filename(const char* filename, const char* ssid);
```

## Checklist

- [x] Runs `make check` successfully.
- [x] Retains 100% code coverage (`make check-coverage`).
- [x] New/changed keys in YAML format are documented.
- [ ] \(Optional\) Adds example YAML for new feature.
- [ ] \(Optional\) Closes an open bug in Launchpad.

PR #522, opened 2024-09-30
RFC: New "networkmanager.passthrough" structure (LP: #2080301)

daniloegea:
## Description

This PR proposes a new format for the `networkmanager.passthrough` section. The new format was a suggestion from the [LP#2080301](https://bugs.launchpad.net/netplan/+bug/2080301) bug reporter. It addresses the problem reported in the bug and is future proof against similar problems.

The main problem with the currently used format is that it's ambiguous. The keyfile `[group].key=value` format is encoded as a string separated by dots. As dots can be used in the key name, it becomes tricky to separate what is the group and what is the key name. A different character could be used but we could eventually hit the same problem.

Arguably we could do something like `[group-name-between-brackets]key-name: value`. Brackets are forbidden in group names (I think?!). But it still would be a change in the format so I opted for turning each entry into mappings.

Here is how it looks like:

```yaml
  nm-devices:
    NM-4bf69677-e326-49b4-8e37-2acf3074f6c7:
      renderer: NetworkManager
      networkmanager:
        uuid: "4bf69677-e326-49b4-8e37-2acf3074f6c7"
        name: "tuntap7"
        passthrough:
          connection:
            type: "tun"
            autoconnect: "false"
            interface-name: "tuntap7"
          proxy: {}
          tun: {}
          ipv4:
            method: "manual"
            address1: "4.3.2.1/24,1.2.3.254"
            dns: "1.2.3.254;4.3.2.254;"
          ipv6:
            addr-gen-mode: "default"
            method: "manual"
            address1: "1:2:3::4/64,1:2:3::abcd"
            dns: "1:2:3::abcd;abcd:3:2::1;"
```

With the integration of netplan and Network Manager on Ubuntu, tweaking the passthrough configuration might be necessary every now and then due to things missing in netplan. This format makes it easier to do that.

The old format is still accepted for backwards compatibility. The parser still accepts it but will emit the new format. So YAMLs generated by Network Manager will be updated if the user modifies the connection and it will still work with existing YAMLs if they are never updated.

**IMPORTANT**: the `datalist` data structure was replaced with hash tables. This will make the order in which elements are added to keyfiles unpredictable. I'm not 100% sure that ordering would cause problems with configuration from the `passthrough` section. My main motivation to drop datalists is that [GQuarks might leak memory](https://github.com/GNOME/glib/blob/main/glib/gquark.c#L295). We hit this issue in our config_fuzzer test due to the number of GQuarks needed when a big number of entries are added to the datalist.

A couple of Network Manager autopkgtests are failing due to the change in the structure and will need to be fixed as well.

I understand it's a big change and we might decide to not merge it so I made it an RFC.

## Checklist

### DIFF
--- a/src/parse-nm.c
+++ b/src/parse-nm.c
@@ -448,39 +448,44 @@ parse_bond_arp_ip_targets(GKeyFile* kf, GArray **targets_arr)
 
 /* Read the key-value pairs from the keyfile and pass them through to a map */
 STATIC void
-read_passthrough(GKeyFile* kf, GData** list)
+read_passthrough(GKeyFile* kf, GHashTable** list)
 {
     gchar **groups = NULL;
     gchar **keys = NULL;
-    gchar *group_key = NULL;
     gchar *value = NULL;
     gsize klen = 0;
     gsize glen = 0;
+    GHashTable* group;
 
-    if (!*list)
-        g_datalist_init(list);
+    if (*list == NULL) {
+        *list = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+    }
     groups = g_key_file_get_groups(kf, &glen);
-    if (groups) {
+    if (groups != NULL) {
         for (unsigned i = 0; i < glen; ++i) {
             klen = 0;
+            group = g_hash_table_lookup(*list, groups[i]);
+            if (group == NULL) {
+                group = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+            }
             keys = g_key_file_get_keys(kf, groups[i], &klen, NULL);
             if (klen == 0) {
                 /* empty group */
-                g_datalist_set_data_full(list, g_strconcat(groups[i], ".", NETPLAN_NM_EMPTY_GROUP, NULL), g_strdup(""), g_free);
+                g_hash_table_insert(*list, g_strdup(groups[i]), group);
+                g_strfreev(keys);
                 continue;
             }
             for (unsigned j = 0; j < klen; ++j) {
                 value = g_key_file_get_string(kf, groups[i], keys[j], NULL);
-                if (!value) {
+                if (value == NULL) {
                     // LCOV_EXCL_START
                     g_warning("netplan: Keyfile: cannot read value of %s.%s", groups[i], keys[j]);
                     continue;
                     // LCOV_EXCL_STOP
                 }
-                group_key = g_strconcat(groups[i], ".", keys[j], NULL);
-                g_datalist_set_data_full(list, group_key, value, g_free);
-                g_free(group_key);
+                g_hash_table_insert(group, g_strdup(keys[j]), value);
             }
+            g_hash_table_insert(*list, g_strdup(groups[i]), group);
             g_strfreev(keys);
         }
         g_strfreev(groups);
--- a/src/parse.c
+++ b/src/parse.c
@@ -527,43 +527,116 @@ handle_generic_map(NetplanParser *npp, yaml_node_t* node, const char* key_prefix
 }
 
 /*
- * Handler for setting a DataList field from a mapping node, inside a given struct
+ * Handle legacy networkmanager.passthrough format:
+ * group_name.key: value
+ *
+ * See https://github.com/canonical/netplan/pull/522
+*/
+STATIC void
+handle_passthrough_legacy_scalar_settings(gchar* key, yaml_node_t* settings, GHashTable* groups)
+{
+    g_autofree char* escaped_value = g_strescape(scalar(settings), STRESCAPE_EXCEPTIONS);
+
+    /* Group name may contain dots, but key name may not.
+     * The "tc" group is a special case, where it is the other way around, e.g.:
+     *   tc->qdisc.root
+     *   tc->tfilter.ffff:
+     */
+
+    gchar **group_key = g_strsplit(key, ".", -1);
+    guint len = g_strv_length(group_key);
+    g_autofree gchar* old_key = NULL;
+    g_autofree gchar* k = NULL;
+    g_autofree gchar* group = NULL;
+
+    if (len < 2) {
+        g_warning("NetworkManager: passthrough key '%s' format is invalid, should be 'group.key'.", key);
+        g_strfreev(group_key);
+        return;
+    }
+    if (!g_strcmp0(group_key[0], "tc") && len > 2) {
+        k = g_strconcat(group_key[1], ".", group_key[2], NULL);
+        group = g_strdup(group_key[0]);
+    } else {
+        k = group_key[len-1];
+        group_key[len-1] = NULL; //remove key from array
+        group = g_strjoinv(".", group_key); //re-combine group parts
+    }
+
+    if (!g_hash_table_contains(groups, group)) {
+        GHashTable *keys = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+        /* Only add to the hash map if it's not the empty group */
+        if (g_strcmp0(k, NETPLAN_NM_EMPTY_GROUP)) {
+            g_hash_table_insert(keys, g_strdup(k), g_strdup(escaped_value));
+        }
+        g_hash_table_insert(groups, g_strdup(group), keys);
+    } else {
+        GHashTable *keys = g_hash_table_lookup(groups, group);
+
+        /* If we find the group again and it's empty, drop all the items from the group */
+        if (!g_strcmp0(k, NETPLAN_NM_EMPTY_GROUP)) {
+            g_hash_table_remove_all(keys);
+        } else {
+            g_hash_table_insert(keys, g_strdup(k), g_strdup(escaped_value));
+        }
+    }
+    g_strfreev(group_key);
+}
+
+/*
+ * Handler for setting a Hashmap field from a mapping node, inside a given struct
  * @entryptr: pointer to the beginning of the to-be-modified data structure
  * @data: offset into entryptr struct where the boolean field to write is located
 */
 STATIC gboolean
-handle_generic_datalist(NetplanParser *npp, yaml_node_t* node, const char* key_prefix, void* entryptr, const void* data, GError** error)
+handle_passthrough(NetplanParser *npp, yaml_node_t* node, const char* key_prefix, void* entryptr, const void* data, GError** error)
 {
     guint offset = GPOINTER_TO_UINT(data);
-    GData** list = (GData**) ((void*) entryptr + offset);
-    if (!*list)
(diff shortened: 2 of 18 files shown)


tests/generator/test_passthrough.py
```diff
@@ -58,6 +58,83 @@ def test_passthrough_basic(self):
 method=ignore
 '''}, '''[device-netplan.ethernets.NM-87749f1d-334f-40b2-98d4-55db58965f5f]
 match-device=type:ethernet
+managed=1\n\n''')
+
+    def test_passthrough_basic_new_format(self):
```
slyon, line 63: nitpick: `_new_format` is very generic. Maybe better call it `test_passthrough_basic_mapping`.

(Also in other places where "new_format" is used.)

tests/generator/test_passthrough.py
```diff
@@ -58,6 +58,83 @@ def test_passthrough_basic(self):
...
+        uuid: 87749f1d-334f-40b2-98d4-55db58965f5f
+        name: some NM id
+        passthrough:
+          connection:
+            uuid: 87749f1d-334f-40b2-98d4-55db58965f5f
+            type: ethernet
+            permissions: ""
+          group2:
+            key: a
+            key: b''', skip_generated_yaml_validation=True)
```
slyon, line 115: question: should the result rather be `key=b` here. So that the value defined later wins?

also followup question: what happens if we mix and match the old format?

e.g.
```yaml
passthrough:
  connection:
    name: "another name"
  connection.name: "some name"
 ```

I'd argue the value defined in the new (mapping) format should take precendence. WDYT?

File: tests/generator/test_passthrough.py [hunk 3]
```diff
@@ -75,6 +152,61 @@ def test_passthrough_wifi(self):
...
+
+[wifi]
+ssid=OTHER-SSID
+mode=infrastructure
+hidden=true
+'''}, '''[device-netplan.wifis.NM-87749f1d-334f-40b2-98d4-55db58965f5f]
+match-device=type:wifi
+managed=1\n\n''')
+
+    def test_passthrough_wifi_new_format(self):
```
slyon, line 226: nitpick: Maybe call it ` test_passthrough_wifi_mapping` to be more precise.

src/validation.c
```diff
@@ -407,7 +407,14 @@ validate_netdef_grammar(const NetplanParser* npp, NetplanNetDefinition* nd, GErr
...
 
-    if (nd->type == NETPLAN_DEF_TYPE_NM && (!nd->backend_settings.passthrough || !g_datalist_get_data(&nd->backend_settings.passthrough, "connection.type")))
+    GHashTable* key = NULL;
+    if (nd->backend_settings.passthrough) {
+        GHashTable* group = g_hash_table_lookup(nd->backend_settings.passthrough, "connection");
+        if (group) {
+            key = g_hash_table_lookup(group, "type");
+        }
+    }
+    if (nd->type == NETPLAN_DEF_TYPE_NM && (!nd->backend_settings.passthrough || !key))
```
slyon, line 417: I don't think we need the `!nd->backend_settings.passthrough` here any more. `key` is only defined if `nd->backend_settings.passthrough` was accessible in the previous if-statement.

tests/cli/test_get_set.py
```diff
@@ -479,15 +479,17 @@ def test_set_delete_nm_passthrough(self):
...
             passthrough:
-              connection.permissions: ""
-              ipv4.dns-search: ""''')
+              connection:
+                permissions: ""
+              ipv4:
+                dns-search: ""''')
         ap_key = 'network.wifis.wlan0.access-points.SOME-SSID'
-        self._set([ap_key+'.networkmanager.passthrough.connection\\.permissions=null'])
+        self._set([ap_key+'.networkmanager.passthrough.connection.permissions=null'])
```
slyon, line 487: question: Do we need to change the escaping in existing `netplan get` calls, out in the wild? That might have the potential to break the snapd integration used in Ubuntu Core.

slyon (commented):
Thank you very much for working on this. I think the new and improved mapping format makes sense and is more sustainable than the existing flat list of group+key-values, going forward.

As can be seen from the CI NetworkManager tests, we need to adopt some cases in NetworkManager, and we should do so in a compatible way, e.g. using "netplan get", which can handle the old and new format, so we won't run into issues when backporting this change to older series.

I like the way you kept it reading the old format! But in general, we should double-check that we don't break backwards compatibility. Especially around `netplan get/set` and corresponding DBus calls, as used by the snapd integration.

Overall, getting rid of the `GData` list data-structure is a win, to avoid memory leaks in upstream GLib. And having a mapping structure is much more readable than the current flat list. I don't think there are any ordering issues introduced by using hashmaps, as we have one hashmap per keyfile group. And also the `GKeyFIle` functions put things into the correct place (group) inside the resulting file automatically.

I left a bunch of comments inline, mostly minor stuff, but also some questions, which we need to think through.

src/parse-nm.c
```diff
@@ -448,25 +448,29 @@ parse_bond_arp_ip_targets(GKeyFile* kf, GArray **targets_arr)
...
             klen = 0;
+            group = g_hash_table_lookup(*list, groups[i]);
+            if (!group) {
+                group = g_hash_table_new(g_str_hash, g_str_equal);
+            }
             keys = g_key_file_get_keys(kf, groups[i], &klen, NULL);
             if (klen == 0) {
                 /* empty group */
-                g_datalist_set_data_full(list, g_strconcat(groups[i], ".", NETPLAN_NM_EMPTY_GROUP, NULL), g_strdup(""), g_free);
+                g_hash_table_insert(*list, g_strdup(groups[i]), group);
```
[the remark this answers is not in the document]
daniloegea: In this particular case, `groups` is created by `g_key_file_get_groups` which will not add repetitions to the result.

Although, there is a different memory leak there: thanks to the `continue` in the next line it will not free the `keys` array.

src/parse.c
```diff
@@ -548,22 +549,78 @@ handle_generic_datalist(NetplanParser *npp, yaml_node_t* node, const char* key_p
         key = yaml_document_get_node(&npp->doc, entry->key);
         value = yaml_document_get_node(&npp->doc, entry->value);
 
-        assert_type(npp, key, YAML_SCALAR_NODE);
-        assert_type(npp, value, YAML_SCALAR_NODE);
-
         escaped_key = g_strescape(scalar(key), STRESCAPE_EXCEPTIONS);
         escaped_value = g_strescape(scalar(value), STRESCAPE_EXCEPTIONS);
```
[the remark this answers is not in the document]
daniloegea: oh, that's a good point. The `escaped_value` is only relevant when the value type is `scalar`. I'll move it to inside the `if` statement.

src/parse.c
```diff
@@ -548,22 +549,78 @@ handle_generic_datalist(NetplanParser *npp, yaml_node_t* node, const char* key_p
...
+            } else {
+                k = group_key[len-1];
+                group_key[len-1] = NULL; //remove key from array
+                group = g_strjoinv(".", group_key); //re-combine group parts
+            }
+
+            if (!g_hash_table_contains(*groups, group)) {
+                GHashTable *keys = g_hash_table_new(g_str_hash, g_str_equal);
+                /* Only add to the hash map if it's not the empty group */
+                if (g_strcmp0(k, NETPLAN_NM_EMPTY_GROUP)) {
```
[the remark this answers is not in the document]
daniloegea: Well, we could do something like `if(k != NULL && k[0] != NETPLAN_NM_EMPTY_GROUP && k[1] == '\0')`... 

`strcmp` will return after checking the first byte in most cases, we would save a function call though (or maybe not even that if `strcmp` gets inlined) but I think we don't need to get into this level of optimization? :)

src/parse.c
```diff
@@ -548,22 +549,78 @@ handle_generic_datalist(NetplanParser *npp, yaml_node_t* node, const char* key_p
...
+             *   key: value
+             */
+            GHashTable* group = g_hash_table_lookup(*groups, escaped_key);
+            if (!group) {
+                group = g_hash_table_new(g_str_hash, g_str_equal);
+                g_hash_table_insert(*groups, g_strdup(escaped_key), group);
+            }
+            handle_generic_map(npp, value, full_key, &group, NULL, error);
+            if (*error) {
+                g_debug("Passthrough handling error, ignoring...: %s", (*error)->message);
```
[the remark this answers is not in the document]
daniloegea: The error message will have plenty of details, I think it should be good enough:

```
** (process:54840): DEBUG: 11:24:28.321: Passthrough handling error, ignoring...: /tmp/fakeroot/etc/netplan/c.yaml:15:13: Error in network definition: duplicate map entry 'b'
            b: 1
            ^
```

tests/cli/test_get_set.py
daniloegea: Yes, that (the escaping with `\`) will not work anymore if you are using get and set with settings inside the `passthrough` section. We probably shouldn't backport this change to LTS releases (like, never), even if we don't give any guarantees about stability for `passthrough`  block.

tests/generator/test_passthrough.py
daniloegea: hmm yeaaah it probably should be `key=b` in the end. But that's actually a problem in the `handle_generic_map` function. It will error out if it finds the same key with different values (see below). Arguably this logic should be used everywhere else then...

```
$ cat /tmp/fakeroot/etc/netplan/a.yaml
network:
  version: 2
  renderer: NetworkManager
  ethernets:
    renderer: NetworkManager
    enp5s0:
      dhcp4: true
      openvswitch:
        other-config:
          one: 1
          one: 2
$ netplan  get --root-dir /tmp/fakeroot
Command failed: /tmp/fakeroot/etc/netplan/a.yaml:10:11: Error in network definition: duplicate map entry 'one'
          one: 1
          ^
```

For the second question, mixed format is supported. Although, if the key is set using the old style first, the parser will ignore it if it's found again using the new format. The problem is the same described above, but for these cases I'm ignoring the error.

I could create a new `handle_generic_map` only for the passthrough settings... but in this case I suppose the same logic should be used across all the parser and not only here...

daniloegea:
Thank you, Lukas.

I addressed a bunch of your comments and left answers for many others.

I think the main highlights are:

- `netplan set` will not work for **updating** fields inside the passthrough block. The reason is the way `handle_generic_map` works when it finds the same fields with different values. It also doesn't work for `openvswitch.other-config` for example due to the same problem. Adding and nullifying fields works fine.
- I updated the documentation too

I changed a lot of stuff, it's a good idea to go over all the code again (sorry for that).

slyon:
FTR: This is an enhancement of https://github.com/canonical/netplan/pull/193

PR #523, opened 2024-10-04
wpa_supplicant: add ExecReload to the service unit

daniloegea:
Add an ExecReload command to call "wpa_cli reconfigure" on the interface. With that we can use "systemctl reload" to reconfigure a wireless device without having to call stop/start.

## Checklist


slyon (approved):
LGTM.

The `/usr/sbin/wpa_cli` binary seems to be installed from the same `wpasupplicant` package in Debian and Ubuntu, so no dependency change needed.

PR #524, opened 2024-10-11
networkmanager: add support for stable-ssid MAC option (LP: #2084234)

daniloegea:
802-11-wireless.cloned-mac-address = stable-ssid was added in Network Manager 1.46 and is available through the GUI applet.

Fixes https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2084234

## Checklist

### DIFF
--- a/tests/integration/wifi.py
+++ b/tests/integration/wifi.py
@@ -181,5 +181,45 @@ class TestNetworkManager(IntegrationTestsWifi, _CommonTests):
         self.assertIn('ssid fake net', out)
         self.assert_iface_up(self.dev_w_ap, ['inet 10.'])
 
+    @unittest.skip("Test if flaky. NM might generate a different MAC address.")
+    def test_wifi_cloned_macaddress_stable_ssid(self):
+        self.setup_ap('''hw_mode=g
+channel=1
+ssid=fake net
+wpa=1
+wpa_key_mgmt=WPA-PSK
+wpa_pairwise=TKIP
+wpa_passphrase=12345678
+''', None)
+
+        with open(self.config, 'w') as f:
+            f.write('''network:
+  renderer: NetworkManager
+  wifis:
+    %(wc)s:
+      addresses: ["192.168.1.42/24"]
+      dhcp4: false
+      dhcp6: false
+      macaddress: stable-ssid
+      access-points:
+        "fake net":
+          password: 12345678''' % {'wc': self.dev_w_client})
+
+        subprocess.check_call(['systemctl', 'start', 'NetworkManager'])
+
+        # Make the generated MAC address predictable
+        # See nm_utils_hw_addr_gen_stable_eth() in NM for details
+        # TODO: save and restore these files to avoid any impact on the
+        # entire test suite.
+        with open('/etc/machine-id', 'w') as f:
+            f.write('ee7ac3602b6306061bd984a41eb1c045\n')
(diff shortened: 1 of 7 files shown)


tests/integration/wifi.py
```diff
@@ -181,5 +181,42 @@ def test_wifi_ap_open(self):
...
+        subprocess.check_call(['systemctl', 'start', 'NetworkManager'])
+
+        # Make the generated MAC address predictable
+        # See nm_utils_hw_addr_gen_stable_eth() in NM for details
+        with open('/etc/machine-id', 'w') as f:
+            f.write('ee7ac3602b6306061bd984a41eb1c045\n')
+        with open('/var/lib/NetworkManager/secret_key', 'w') as f:
+            f.write('nm-v2:hnIHoHp4p9kaEWU5/+dO+gFREirN1AsMoO1MPaoYxCc=')
+
+        subprocess.check_call(['systemctl', 'restart', 'NetworkManager'])
```
daniloegea, line 216: note: this test was a little flaky, bouncing NM after overwriting the files above made it stable (tested many times locally as they will be skipped here anyway)

daniloegea: meh I just tried again and it generated a different MAC address... I might just drop this test...

tests/integration/wifi.py
```diff
@@ -181,5 +181,42 @@ def test_wifi_ap_open(self):
...
+          password: 12345678''' % {'wc': self.dev_w_client})
+
+        subprocess.check_call(['systemctl', 'start', 'NetworkManager'])
+
+        # Make the generated MAC address predictable
+        # See nm_utils_hw_addr_gen_stable_eth() in NM for details
+        with open('/etc/machine-id', 'w') as f:
+            f.write('ee7ac3602b6306061bd984a41eb1c045\n')
+        with open('/var/lib/NetworkManager/secret_key', 'w') as f:
+            f.write('nm-v2:hnIHoHp4p9kaEWU5/+dO+gFREirN1AsMoO1MPaoYxCc=')
```
slyon, line 214: I guess this might influence other tests using the same testbed. So could have side-effects on the whole test suite.

tests/integration/wifi.py
slyon: Maybe mark it `@unittests.skip()` with a comment for now? So we can improve upon in the future.
Dropping the work that was already done would be sad.

slyon (approved):
Thanks, LGTM! And should be fine to be pulled into the pending 1.1.1 stable release.

I left a small remark about the autopkgtest (thanks for providing one!)

tests/integration/wifi.py
daniloegea: I did that

tests/integration/wifi.py
daniloegea: I added a TODO comment to improve this. For now the test is skipped.

slyon:
Regarding the MAC address flakyness, apparently networkd changes to a random MAC address after taking over an interface, changing the original MAC that was assigned when the (test-)interface was created, e.g. via `ip link add ...`.

We tried to tackle that in the past, but maybe we're seeing another instance of this here?? https://github.com/canonical/netplan/commit/1413f0e7b8f4d068f817a009d998656de3224370

daniloegea:
It could be something related to that. I think the previous MAC address is used as one of the inputs for the calculation of the new one...

PR #529, opened 2024-12-02
Do not execute generator during daemon-reload

slyon:
## Description
As intended when originally implemented the netplan generator.stamp is supposed to avoid re-generating all of the Netplan configuration during 'systemd daemon-reload'.

The stamp has originally been placed in /run/systemd/generator/, which is nowadays cleaned during 'daemon-reload', leading to Netplan regenerating its configuration during each such call.

Let's move the stamp file into /run/netplan/generator.stamp instead to avoid that. Triggering the Netplan generator only at reboot, or when called explicitly.

Original intend: d2a45b3066cb20f8cfeb3e778af025ecdfbc6461

## Checklist

### DIFF
--- a/src/generate.c
+++ b/src/generate.c
@@ -236,7 +236,9 @@ int main(int argc, char** argv)
             g_fprintf(stderr, "%s can not be called directly, use 'netplan generate'.", argv[0]);
             return 1;
         }
-        generator_run_stamp = g_build_path(G_DIR_SEPARATOR_S, files[0], "netplan.stamp", NULL);
+        /* Write generator.stamp in /run/netplan/generator.stamp, this is due to
+         * /run/systemd/generator/... being cleaned on every daemon-reload. */
+        generator_run_stamp = g_build_path(G_DIR_SEPARATOR_S, rootdir, "run", "netplan", "generator.stamp", NULL);
         if (g_access(generator_run_stamp, F_OK) == 0) {
             g_fprintf(stderr, "netplan generate already ran, remove %s to force re-run\n", generator_run_stamp);
             return 0;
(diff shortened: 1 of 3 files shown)


tests/generator/test_args.py
```diff
@@ -409,6 +410,7 @@ def test_systemd_generator_escaping(self):
 
         # after removing the stamp it generates again, and not trip over the
```
daniloegea, line 411: oh the test above was passing because we don't call it through a daemon-reload

slyon:
As the [lxd-network-manager](https://github.com/canonical/netplan/actions/runs/12123010981/job/33797527469?pr=529#logs) CI shows, the NetworkManager-Netplan integration seem to rely on the re-generation of Netplan config during `daemon-reload`, which probably needs to be fixed first.

daniloegea:
> As the [lxd-network-manager](https://github.com/canonical/netplan/actions/runs/12123010981/job/33797527469?pr=529#logs) CI shows, the NetworkManager-Netplan integration seem to rely on the re-generation of Netplan config during `daemon-reload`, which probably needs to be fixed first.

oohh right, that's true. Because it calls "generate" and NM capabilities are limited, it can't do things such as changing files owners... it only works because it does a daemon-reload... oofff

slyon:
Adding `CapabilityBoundingSet+=CAP_CHOWN` to NetworkManager.service might help, but would have other (security) implications.

slyon:
FTR: This caused an issue in Debian Trixie cloud images: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1100418

src/generate.c
```diff
@@ -318,6 +320,7 @@ int main(int argc, char** argv)
 
         /* Leave a stamp file so that we don't regenerate the configuration
          * multiple times and userspace can wait for it to finish */
+        _netplan_safe_mkdir_p_dir(generator_run_stamp);
         FILE* f = fopen(generator_run_stamp, "w");
         g_assert(f != NULL);
```
slyon, line 325: attention: This failed in Debian trixie.
```
| # /usr/lib/systemd/system-generators/netplan . . .
| **
| ERROR:../src/generate.c:325:main: assertion failed: (f != NULL)
| Bail out! ERROR:../src/generate.c:325:main: assertion failed: (f != NULL)
| Aborted
```

slyon:
See also https://github.com/canonical/netplan/pull/552

PR #540, opened 2025-01-31
bridges: implement vlans, port-vlans, vlan-filtering and vlan-default-pvid options

p-strusiewiczsurmacki-mobica:
## Description

This PR is a followup to #127 .

As described in #127: 
Implements the vlans & port-vlans option keys for bridges, as required by the "NetworkManager readwrite plugin" spec. Currently only supports the nm backend.

I've rebased the old PR onto the current `main` and it seems to work OK (at least the tests are passing). Additionally added the `vlan-filtering` bool that can be forced `true` even if no `vlans` are defined (as described by @chdxD1 in [this comment here](https://github.com/canonical/netplan/pull/410#issuecomment-1750315850).

~~Right now I am starting to work on `networkd` backend. However decided to create this draft PR to enable an early review and to just let others know that I am working on this.~~

I've added `vlan-default-pvid` to set default PVID for a bridge and `networkd` backend so i think the PR is ready. :)

`make check-coverage` fails for me in `codestyle` test within `netplan/_build/meson-private/pycompile.py`, which is Meson's file so I believe this is out of scope here, therefore checking `make check-coverage` in the checklist as passed.

~~I can also see that GitHub action `NetworkManager Autopkgtest / lxd-network-manager` is failing due to `libc6-dev : Depends: libc6 (= 2.39-0ubuntu8.3) but 2.39-0ubuntu8.4 is to be installed` but it also seems to be an issue with Ubuntu repositories I think.~~ I think Ubuntu's repositories were fixed already.

## Checklist

### DIFF
--- a/doc/.custom_wordlist.txt
+++ b/doc/.custom_wordlist.txt
@@ -133,6 +133,7 @@ networkd
 nm
 passthrough
 programmatically
+pvid
 renderer
 reselection
 runtime
@@ -145,6 +146,8 @@ udev
 unconfigured
 unencrypted
 untagged
+vid
+vlans
 vSwitch
 wpa
 wpasupplicant
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -1361,6 +1361,11 @@ The specific settings for bridges are defined below.
   - **`vlan-filtering`** (boolean)
 
     > Enables VLAN filtering. Will be enabled by default if *vlans* are defined.
+  
+  - **`vlan-default-pvid`** (scalar)
+
+    > Specifies the default port VLAN ID. Can be set to values between 1 and 4094,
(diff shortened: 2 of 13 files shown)


p-strusiewiczsurmacki-mobica:
Additional info: I've changed the code by @slyon a bit as it seems that for both `NetworkManager` and `networkd` only one PVID can be specified per interface. `Networkd` seems to specify this [explicity](https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#PVID=) in the docs. `NetworkManager` does not, but I've tried to configure multiple PVIDs with `nmcli` on Ubuntu 24.04 and it seems to not accept ranges. So I assume this is what is expected.

p-strusiewiczsurmacki-mobica:
Hi @Forst, I think that's a neat idea! However, I'll wait for an opinion from @slyon as this implementation is based on theirs after all.

PR #547, opened 2025-03-17
Fix function prototypes for *_new() functions of parser and state

JanSoundhouse:
The state.h and parse.h files are part of the public api of libnetplan. These function prototypes declared there should adhere the c-standard.
Function prototypes for functions that don't take any parameters should have `void` as the parameter list. Function prototypes with an empty parameter list are treated by the compiler as accepting any arbitrary parameter.

### DIFF
--- a/include/parse.h
+++ b/include/parse.h
@@ -35,7 +35,7 @@
  * @return  An empty @ref NetplanParser
  */
 NETPLAN_PUBLIC NetplanParser*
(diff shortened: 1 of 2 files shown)


slyon (approved):
Thanks for your contribution, LGTM!

slyon:
closed & re-open to re-trigger the CI.

PR #548, opened 2025-03-21
Fix `netplan try` revert (LP: #2083029)

MggMuggins:
## Description
Fixes [LP#2083029](https://bugs.launchpad.net/netplan/+bug/2083029)

When using `systemd-networkd` as a backend, `netplan try` backs up `/run/systemd/network` before applying the new configuration. If the user fails to confirm the new configuration before the timeout is up, Netplan restores the old configuration by moving the backup into `/run/systemd/network` and invokes `networkctl reload` to apply the known-good configuration.

However, `networkd` will not/cannot read config files if their ownership is not `root:systemd-network`.

In effect, `netplan try` does not revert network changes _at all_ when using `networkd`.

Additionally, following the Ubuntu patches to NetworkManager in 23.10 (see [netplan everywhere](https://netplan.readthedocs.io/en/stable/netplan-everywhere/)), `netplan generate` is called (via DBus) when NetworkManager refreshes its config. This "reverts the revert" from `netplan try`, so restoring the old config doesn't work with `NetworkManager` in Ubuntu >23.10 either.

This PR adds an autopkg test for `netplan try` so as to catch these regressions before they are released.

## Implementation Notes
I took advantage of the [stamp](https://github.com/canonical/netplan/commit/e19441c32a35cc58770753dfa8e6783438db295b) that exists when `netplan try` is running. There doesn't appear to me to be a way to prevent NetworkManager from calling generate on reload.

I considered that this may be prone to races if NetworkManager takes its time reloading the configuration; however, it looks like `netplan try` calls `command_apply` with `sync=True`, so it shouldn't delete the stamp until [after NetworkManager has finished loading its config](https://github.com/canonical/netplan/blob/main/netplan_cli/cli/commands/apply.py#L305-L320).

## Test Plan
Using the default netplan configuration in a LXD instance:
```yaml
network:
    version: 2
    ethernets:
        enp5s0:
            dhcp4: true
```

Replace the configuration with a static IP and run `netplan try --timeout 3`, and allow the timeout to elapse.
```yaml
network:
    version: 2
    ethernets:
        enp5s0:
            dhcp4: false
            addresses:
              - 10.58.215.80/24
```

Before the fix in Jammy, the static IP remains assigned to `enp5s0` after the revert. In Noble and later, the interface has the IPv4 removed, but a new dhcp lease is not acquired.

`networkctl status --all` will report the device as `unmanaged`. In Jammy:
```
● 2: enp5s0
                     Link File: /usr/lib/systemd/network/99-default.link
                  Network File: /run/systemd/network/10-netplan-enp5s0.network
                          Type: ether
                         State: routable (unmanaged)
                  Online state: online
```

## Checklist

- [ ] ~~New/changed keys in YAML format are documented.~~
- [ ] \(Optional\) ~~Adds example YAML for new feature.~~
- [x] \(Optional\) Closes an open bug in Launchpad. LP#2083029

Test failures look like Launchpad having trouble.

### DIFF
--- a/netplan_cli/configmanager.py
+++ b/netplan_cli/configmanager.py
@@ -153,12 +153,18 @@ class ConfigManager(object):
 
     def _copy_tree(self, src, dst, missing_ok=False):
         try:
-            shutil.copytree(src, dst)
+            shutil.copytree(src, dst, copy_function=copy_with_ownership)
         except FileNotFoundError:
             if missing_ok:
                 pass
             else:
                 raise
+
+
+def copy_with_ownership(src, dst, follow_symlinks=True):
+    shutil.copy2(src, dst, follow_symlinks=follow_symlinks)
+    stat = os.stat(src, follow_symlinks=follow_symlinks)
+    os.chown(dst, stat.st_uid, stat.st_gid, follow_symlinks=follow_symlinks)
 
 
 class ConfigurationError(Exception):
--- a/netplan_cli/configmanager.py
+++ b/netplan_cli/configmanager.py
@@ -89,6 +89,7 @@ class ConfigManager(object):
         # Convoluted way to dump the parsed config to the logs...
         with tempfile.TemporaryFile() as tmp:
             self.np_state._dump_yaml(output_file=tmp)
+            tmp.seek(0)
             logging.debug("Merged config:\n{}".format(tmp.read()))
 
         return self.np_state
--- a/netplan_cli/configmanager.py
+++ b/netplan_cli/configmanager.py
@@ -151,9 +151,9 @@ class ConfigManager(object):
     def _copy_file(self, src, dst):
         shutil.copy(src, dst)
 
-    def _copy_tree(self, src, dst, missing_ok=False):
+    def _copy_tree(self, src, dst, missing_ok=False, **kwargs):
         try:
-            shutil.copytree(src, dst, copy_function=copy_with_ownership)
+            shutil.copytree(src, dst, copy_function=copy_with_ownership, **kwargs)
         except FileNotFoundError:
             if missing_ok:
                 pass
--- a/netplan_cli/configmanager.py
+++ b/netplan_cli/configmanager.py
@@ -104,14 +104,14 @@ class ConfigManager(object):
 
     def backup(self, backup_config_dir=True):
         if backup_config_dir:
-            self._copy_tree(os.path.join(self.prefix, "etc/netplan"),
-                            os.path.join(self.temp_etc, "netplan"))
-        self._copy_tree(os.path.join(self.prefix, "run/NetworkManager/system-connections"),
-                        os.path.join(self.temp_run, "NetworkManager", "system-connections"),
-                        missing_ok=True)
-        self._copy_tree(os.path.join(self.prefix, "run/systemd/network"),
-                        os.path.join(self.temp_run, "systemd", "network"),
-                        missing_ok=True)
+            self.copy_tree(os.path.join(self.prefix, "etc/netplan"),
+                           os.path.join(self.temp_etc, "netplan"))
+        self.copy_tree(os.path.join(self.prefix, "run/NetworkManager/system-connections"),
+                       os.path.join(self.temp_run, "NetworkManager", "system-connections"),
+                       missing_ok=True)
+        self.copy_tree(os.path.join(self.prefix, "run/systemd/network"),
+                       os.path.join(self.temp_run, "systemd", "network"),
+                       missing_ok=True)
 
     def revert(self):
         try:
@@ -122,12 +122,12 @@ class ConfigManager(object):
             temp_networkd_path = "{}/systemd/network".format(self.temp_run)
             if os.path.exists(temp_nm_path):
                 shutil.rmtree(os.path.join(self.prefix, "run/NetworkManager/system-connections"))
-                self._copy_tree(temp_nm_path,
-                                os.path.join(self.prefix, "run/NetworkManager/system-connections"))
+                self.copy_tree(temp_nm_path,
+                               os.path.join(self.prefix, "run/NetworkManager/system-connections"))
             if os.path.exists(temp_networkd_path):
                 shutil.rmtree(os.path.join(self.prefix, "run/systemd/network"))
-                self._copy_tree(temp_networkd_path,
-                                os.path.join(self.prefix, "run/systemd/network"))
+                self.copy_tree(temp_networkd_path,
+                               os.path.join(self.prefix, "run/systemd/network"))
         except Exception as e:  # pragma: nocover (only relevant to filesystem failures)
             # If we reach here, we're in big trouble. We may have wiped out
             # file NM or networkd are using, and we most likely removed the
@@ -152,7 +152,7 @@ class ConfigManager(object):
     def _copy_file(self, src, dst):
         shutil.copy(src, dst)
 
-    def _copy_tree(self, src, dst, missing_ok=False, **kwargs):
+    def copy_tree(self, src, dst, missing_ok=False, **kwargs):
         try:
             shutil.copytree(src, dst, copy_function=copy_with_ownership, **kwargs)
         except FileNotFoundError:
--- a/src/generate.c
+++ b/src/generate.c
@@ -341,7 +341,7 @@ int main(int argc, char** argv)
         }
         g_autofree char* glob_run = g_build_path(G_DIR_SEPARATOR_S,
                                                  rootdir != NULL ? rootdir : G_DIR_SEPARATOR_S,
-                                                 "run/systemd/system/netplan-*.service",
+                                                 "run", "systemd", "system", "netplan-*.service",
                                                  NULL);
         if (!glob(glob_run, 0, NULL, &gl)) {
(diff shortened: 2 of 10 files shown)


MggMuggins:
I opened this PR prematurely; the current changes do not fix the bug. I would mark this as a draft but I don't have permission to do so. Still working on this...

MggMuggins:
This is ready for review.

Test failures:
- HTTP request failures (look transient)
- Debian testing autopkgtest

I was able to reproduce the Debian testing failure using autopkgtest locally, but it didn't reproduce when I ran the integration tests in a trixie VM. I ran the Ubuntu autopkgtests locally; they fail too but with an error message:
```
test_netplan_try (__main__.TestNetworkManager.test_netplan_try) ... eth43 eth43 Error: NetworkManager is not running.
```

I'm willing to continue chasing these but I'd appreciate a review of this PR to make sure that this is a reasonable set of fixes before I sink more time into the CI. Thanks!

slyon:
@MggMuggins Thank you very much for your work on this. I retriggered some of the test cases that failed for odd reasons. The `scenarios` autopkgtests are now failing for Ubuntu & Debian. This seems related to the `netplan try` command. Can you please investigate what's happening there?

MggMuggins:
Yeah, this is definitely a bug in my tests; fixing these is my priority for this morning. Thanks for your patience.

MggMuggins:
Tests are green.

Also added a commit to fix debug logging of the merged config.

tests/integration/base.py
```diff
@@ -373,6 +374,38 @@ def generate_and_settle(self, wait_interfaces=None, state_dir=None):
...
             print('\nWARNING: systemd units changed without reload:', out)
+
+        self.settle(wait_interfaces, state_dir)
+
+    def try_and_settle(self, try_func, wait_interfaces_try=None, wait_interfaces_after=None):
+        '''Run `netplan try`, wait for the configuration to apply, run `try_func` and allow try to time out'''
+
+        # Because `netplan try` uses select(2) to implement the timeout, stdin
+        # needs to be a stream that is never ready to read.
+        fifo_path = f"/tmp/netplan-try-test.{random.randint(10000, 99999)}"
```
slyon, line 385: though (non-blocking): This could introduce a flaky failure if the path already exists (chances are rather low). I wonder if we could use a higher level function, such as https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp, removing the created tempfile right afterward and re-using the path for our FIFO.

tests/integration/base.py
```diff
@@ -373,6 +374,38 @@ def generate_and_settle(self, wait_interfaces=None, state_dir=None):
...
+        os.mkfifo(fifo_path)
+        fifo = os.open(fifo_path, os.O_RDONLY | os.O_NONBLOCK)
+
+        proc = subprocess.Popen(["netplan", "try", "--timeout=5"], stdin=fifo)
+
+        # wait for apply to finish
+        while not pathlib.Path("/run/netplan/netplan-try.ready").exists():
+            time.sleep(0.1)
```
slyon, line 393: I wonder if timeouts are big enough here, especially for slow architectures, like riscv64 or loong64. In other tests (`regressions.py:test_try_accept/reject_*`) we're mostly working with 5 second delays, before triggering any action. So you should consider increasing that `--timeout=5` to something slightly bigger.

Furthermore, we might want to add an explicit termination condition for the `while` loop, to an avoid endless loop when `netplan try --timeout=X` does not work correctly or create the stamp in time. E.g. count the iterations and explicitly break after `TIMEOUT*10` iterations of `sleep(0.1)`.

tests/integration/base.py
```diff
@@ -373,6 +374,38 @@ def generate_and_settle(self, wait_interfaces=None, state_dir=None):
...
+
+            try_func()
+
+            proc.wait()
+            self.settle(wait_interfaces=wait_interfaces_after)
+        finally:
+            # don't step on other tests
+            proc.wait()
+            os.close(fifo)
+            os.unlink(fifo_path)
```
slyon, line 406: praise: Thanks for cleaning up. But what happens when we bail out further above? Should we extend the `try:` block to include the original  `subprocess.Popen` and `while` loop?

tests/integration/scenarios.py
```diff
@@ -134,6 +134,48 @@ def test_remove_virtual_interfaces(self):
         res = subprocess.run(['ip', 'link', 'show', 'dev', 'br54'], capture_output=True, text=True)
         self.assertIn('not exist', res.stderr)
 
+    # https://bugs.launchpad.net/netplan/+bug/2083029
+    def test_netplan_try(self):
```
slyon, line 138: nitpick: 
```suggestion
    def test_netplan_try_lp2083029(self):
```

praise: Thank you so much for adding proper integration tests!

tests/integration/scenarios.py
```diff
@@ -134,6 +134,48 @@ def test_remove_virtual_interfaces(self):
...
+        self.setup_eth(None)
+        with open(self.config, 'w') as f:
+            f.write('''network:
+  renderer: %(r)s
+  ethernets:
+    %(e2c)s:
+      match: {name: %(e2c)s}
```
slyon, line 145: suggestion: Drop the `match:` stanza, as matching on name is implicit (using the NetDefID, e.g. `%(e2c)s`) by default. Same in the YAML below.

tests/integration/scenarios.py
```diff
@@ -134,6 +134,48 @@ def test_remove_virtual_interfaces(self):
...
+                                 ['inet 192.168.6.201/24'])
+
+        self.try_and_settle(assert_static_ip,
+                            wait_interfaces_try=[
+                                self.state(self.dev_e2_client, "inet 192.168.6.201"),
+                            ],
+                            wait_interfaces_after=[
+                                # Use "dynamic" instead of state_dhcp4 since
+                                # state_dhcp4 will match our static address.
+                                self.state(self.dev_e2_client, "dynamic"),
```
slyon, line 171: suggestion (non-blocking): Should we use another IP range, e.g. `10.1.2.3/24` to avoid confusion with the DHCP4 range for eth42 (`192.168.6.10..192.168.6.200`)?

netplan_cli/configmanager.py
```diff
@@ -153,14 +153,20 @@ def _copy_file(self, src, dst):
...
-            shutil.copytree(src, dst)
+            shutil.copytree(src, dst, copy_function=copy_with_ownership)
         except FileNotFoundError:
             if missing_ok:
                 pass
             else:
                 raise
 
 
+def copy_with_ownership(src, dst, follow_symlinks=True):
```
slyon, line 165: question: I wonder if we should use this function in `try_command.py:revert()` (L142), too?

slyon (commented):
Thank you very much this contribution! This is an excellent analysis and test coverage. Much appreciated! It solves an important issue with `netplan try`.

I left some minor comments inline. Let me know what you think.

I think this might also solve https://bugs.launchpad.net/netplan/+bug/2027583

tests/integration/scenarios.py
MggMuggins: Wonder about defining a static lease for the test interfaces instead, that way `state_dhcp4` can just match the IP from the static lease

src/generate.c
```diff
@@ -243,6 +244,22 @@ int main(int argc, char** argv)
...
+    // The file at netplan_try_stamp is created while `netplan try` is waiting
+    // for user confirmation. If generate is triggered while netplan try is
+    // running, we shouldn't regenerate the configuration.
+    // We can be called by either systemd (as a generator during daemon-reload)
+    // or by NetworkManager when it is reloading configuration (Ubuntu >23.10),
+    // see https://netplan.readthedocs.io/en/stable/netplan-everywhere/.
+    // LP #2083029
+    netplan_try_stamp = g_build_path(G_DIR_SEPARATOR_S,
+                                     rootdir != NULL ? rootdir : G_DIR_SEPARATOR_S,
+                                     "run/netplan/netplan-try.ready",
```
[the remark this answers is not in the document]
MggMuggins: I based this on the other uses of `g_build_path` in the same file; added a commit that makes this consistent across the project.

netplan_cli/configmanager.py
MggMuggins: I added a commit that does it; I'm not sure that it's particularly impactful but I agree that keeping the behavior consistent is reasonable. Maybe should `s/_copy_tree/copy_tree/`.

MggMuggins:
Thanks for the review! Test failure should clear on a re-run.

tests/integration/scenarios.py
slyon: Yeah, I like the idea, but I'm not sure if it would be working for all the tests. Sometimes we add additional interfaces like bridges and try to get a DHCP lease on those, potentially multiple interfaces. So a single, static lease might not be enough.

Anyway, I considered this non-blocking. So let's keep it as-is for now!

slyon (approved):
Thank you very much for addressing my remarks. Test are now all green. I just pushed a little rename, as suggested by @MggMuggins (https://github.com/canonical/netplan/pull/548/commits/b87c3a56347f3832a4d43bbf034cc221c8f6e81a).

LGTM!

PR #550, opened 2025-04-08
CI: autopkgtest edge cases for slow architectures and without systemd-resolved

slyon:
## Description
This PR contains several autopkgtest enhancements:

When a new veth device is created via iproute2 ("ip link add ..."), the kernel
assigns a MAC address. Once udev picks up the interface, it will apply systemd
"MACAddressPolicy=permanent" default policy. But as veths cannot (by nature)
have a permanent MAC, systemd-udev will create a random (but persistent) MAC,
changing the kernel MAC.

Previously we waited for 0.1 sec to (hopefully) pick up the later system-udev
MAC, but sometimes that wasn't enough, and we still got the earlier kernel MAC.

My patch drops the "time.sleep(0.1)" and instead triggers our test interfaces
and waits for settlement of those uevents via 'udevadm', to avoid race
conditions.

... Currently testing in experimental on real DebCI infrastructure.

Also including smaller timing fixes in tests for slow architectures (e.g. riscv64 or loong64) and avoiding systemd-resolved as a test-dependency of autopkgtests.

## Checklist

- [x] \(Optional\) Closes an open bug in ~~Launchpad~~ Debian. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1100907


slyon:
It's working in Debian with systemd-resolved not enabled by default: https://salsa.debian.org/debian/netplan.io/-/merge_requests/18

slyon:
All 3 commits are passing CI & test in DebCI as part of the `1.1.2-5~exp2` upload to `experimental` (https://ci.debian.net/packages/n/netplan.io/). It's fixing a release-critical bug in Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1100907

PR #555, opened 2025-07-09
Fail cleanly when systemd-networkd is not installed

jonathanspw:
## Description
This patch makes the failure case of `netplan status` cleaner when systemd-networkd is not present.  This is at least relevant to Fedora/EPEL use-cases.

Before:
```
[root@alma-9-srpm-test 1.el9]# netplan status
Failed to start systemd-networkd.service: Unit systemd-networkd.service not found.
    netplan.main()
  File "/usr/share/netplan/netplan_cli/cli/core.py", line 58, in main
    self.run_command()
  File "/usr/share/netplan/netplan_cli/cli/utils.py", line 332, in run_command
    self.func()
  File "/usr/share/netplan/netplan_cli/cli/commands/status.py", line 77, in run
    self.run_command()
  File "/usr/share/netplan/netplan_cli/cli/utils.py", line 332, in run_command
    self.func()
  File "/usr/share/netplan/netplan_cli/cli/commands/status.py", line 829, in command
    system_state = SystemConfigState(self.ifname, self.all)
  File "/usr/share/netplan/netplan_cli/cli/state.py", line 437, in __init__
    utils.systemctl('start', ['systemd-networkd.service'], True)
  File "/usr/share/netplan/netplan_cli/cli/utils.py", line 118, in systemctl
    subprocess.check_call(command)
  File "/usr/lib64/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['systemctl', 'start', 'systemd-networkd.service']' returned non-zero exit status 5.
```

After:
```
[root@alma-9-srpm-test 1.el9]# netplan status
systemd-networkd is required for 'netplan status' functionality
```

It also fixes a bug in the `systemctl_is_installed` def.  `systemctl is-enabled` only returns 1/0 so it needs to use `systemctl status` instead to be able to use the `4` exit code to determine if a given service is installed.

## Checklist

### DIFF
--- a/netplan_cli/cli/state.py
+++ b/netplan_cli/cli/state.py
@@ -433,6 +433,9 @@ class SystemConfigState():
                               'but systemd-networkd.service is masked. '
                               'Please start it.')
                 sys.exit(1)
+            if not utils.systemctl_is_installed('systemd-networkd.service'):
+                logging.error('systemd-networkd is required for \'netplan status\' functionality')
+                sys.exit(1)
             logging.debug('systemd-networkd.service is not active. Starting...')
             utils.systemctl('start', ['systemd-networkd.service'], True)
 
--- a/netplan_cli/cli/utils.py
+++ b/netplan_cli/cli/utils.py
@@ -156,7 +156,7 @@ def systemctl_is_masked(unit_pattern):
 
 def systemctl_is_installed(unit_pattern):
     '''Return True if returncode is other than "not-found" (4)'''
-    res = subprocess.run(['systemctl', 'is-enabled', unit_pattern],
+    res = subprocess.run(['systemctl', 'status', unit_pattern],
                          stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                          text=True)
     if res.returncode != 4:
--- a/tests/cli/test_status.py
+++ b/tests/cli/test_status.py
@@ -531,6 +531,18 @@ netplan-global-state:
         self.assertEqual(1, e.exception.code)
         self.assertIn('systemd-networkd.service is masked', cm.output[0])
 
+    @patch('netplan_cli.cli.utils.systemctl_is_installed')
+    @patch('netplan_cli.cli.utils.systemctl_is_active')
+    @patch('netplan_cli.cli.utils.systemctl_is_masked')
+    def test_call_cli_networkd_installed_false(self, is_installed_mock, is_active_mock, is_masked_mock):
+        is_active_mock.return_value = False
+        is_masked_mock.return_value = False
+        is_installed_mock.return_value = False
+        with self.assertLogs() as cm, self.assertRaises(SystemExit) as e:
+            self._call([])
+        self.assertEqual(1, e.exception.code)
+        self.assertIn('systemd-networkd is required for \'netplan status\' functionality', cm.output[0])
+
     @patch('netplan_cli.cli.state.NetplanConfigState.__init__')
     @patch('netplan_cli.cli.state_diff.NetplanDiffState.__init__')
     @patch('netplan_cli.cli.state_diff.NetplanDiffState.get_diff')
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -402,7 +402,7 @@ class TestUtils(unittest.TestCase):
         os.environ['PATH'] = os.path.dirname(self.mock_cmd.path) + os.pathsep + path_env
         self.assertTrue(utils.systemctl_is_installed('some.service'))
         self.assertEqual(self.mock_cmd.calls(), [
-            ['systemctl', 'is-enabled', 'some.service']
+            ['systemctl', 'status', 'some.service']
         ])
 
     def test_systemctl_is_installed_false(self):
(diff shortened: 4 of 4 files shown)


netplan_cli/cli/state.py
```diff
@@ -434,6 +434,9 @@ def __init__(self, ifname=None, all=False):
                               'Please start it.')
                 sys.exit(1)
             logging.debug('systemd-networkd.service is not active. Starting...')
+            if not utils.systemctl_is_installed('systemd-networkd.service'):
+                logging.error('systemd-networkd is required for \'netplan status\' functionality')
+                sys.exit(1)
```
slyon, line 439: nitpick: We might want to move this above the `systemd-networkd.service is not active. Starting...` debug message, as the `systemctl start` happens only after.

netplan_cli/cli/utils.py
```diff
@@ -156,7 +156,7 @@ def systemctl_is_masked(unit_pattern):
 
 def systemctl_is_installed(unit_pattern):
     '''Return True if returncode is other than "not-found" (4)'''
-    res = subprocess.run(['systemctl', 'is-enabled', unit_pattern],
+    res = subprocess.run(['systemctl', 'status', unit_pattern],
```
slyon, line 159: question: Can you please provide additional reference of why this change is needed?

According to `man systemctl(1)` the `is-enabled` command should return `4` in case of "not-found": https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#is-enabled%20UNIT%E2%80%A6

```
       is-enabled UNIT...
           Checks whether any of the specified unit files are enabled (as with enable). Returns an exit code of 0 if
           at least one is enabled, non-zero otherwise. Prints the current enable status (see table). To suppress
           this output, use --quiet. To show installation targets, use --full.

           Table 1.  is-enabled output
           ┌───────────────────┬───────────────────────────────┬───────────┐
           │ Name              │ Description                   │ Exit Code │
           ├───────────────────┼───────────────────────────────┼───────────┤
           │ "enabled"         │ Enabled via .wants/,          │           │
           ├───────────────────┤ .requires/ or Alias= symlinks │           │
           │ "enabled-runtime" │ (permanently in               │ 0         │
           │                   │ /etc/systemd/system/, or      │           │
           │                   │ transiently in                │           │
           │                   │ /run/systemd/system/).        │           │
           ├───────────────────┼───────────────────────────────┼───────────┤
[...]
           │ "not-found"       │ The unit file doesn't exist.  │ 4         │
           └───────────────────┴───────────────────────────────┴───────────┘

           Added in version 238.
```

So IMO it should be working as intended and we might not need this change.

slyon (commented):
Thanks for this improvement!
I left two smaller inline comments below.

@Conan-Kudo IIRC systemd-networkd is not needed in principle for `netplan status` to function. But it's an implementation detail that we rely on systemd-networkd data heavily (in `netplan_cli/cli/state.py:__init__()`), besides iproute2 and NetworkManager data.

In theory, it could be changed to use only iproute2 data as ground truth and amend that with systemd-networkd and/or NetworkManager data - whatever is available. And I would be open about that when somebody wants to drive such change.

netplan_cli/cli/utils.py
jonathanspw: So, in practice, this doesn't seem to be the case for EL8 and EL9.  In EL10 it works as expected, but so does `status`.  `status` seems to be portable across all EL/systemd versions while `is-active` doesn't.

EL8:
```
[root@server ~]# systemctl is-enabled netplan
Failed to get unit file state for netplan.service: No such file or directory
[root@server ~]# echo $?
1
[root@server ~]# systemctl status netplan
Unit netplan.service could not be found.
[root@server ~]# echo $?
4
```

EL9:
```
[root@atl ~]# echo $?
3
[root@atl ~]# systemctl status netplan
Unit netplan.service could not be found.
[root@atl ~]# echo $?
4
```

EL10:
```
[root@netplan ~]# systemctl is-active netplan
inactive
[root@netplan ~]# echo $?
4
[root@netplan ~]# systemctl status netplan
Unit netplan.service could not be found.
[root@netplan ~]# echo $?
4
```

Docs cite `Added in version 238.` while even EL8 has systemd 239...so I'm not sure what's up with it not working as intended on 8/9, but `status` is consistent.

netplan_cli/cli/state.py
jonathanspw: Good nitpick.  Fixed.

jonathanspw:
Rebased PR against `main`.

netplan_cli/cli/utils.py
slyon: OK, fair enough. The `systemctl` manpage also describes this as "no such unit" (https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#Exit%20status)

And checking it against Ubuntu LTS version confirms your EL8/9/10 findings from above (`systemctl status` being consistent, while `systemctl is-active` works on newer releases only).

Ubuntu 20.04 LTS (systemd 245.4-4ubuntu3.24)
```
root@humane-chamois:~# systemctl is-enabled netplan
Failed to get unit file state for netplan.service: No such file or directory
root@humane-chamois:~# echo $?
1
root@humane-chamois:~# systemctl status netplan
Unit netplan.service could not be found.
root@humane-chamois:~# echo $?
4
```

Ubuntu 22.04 LTS (systemd 249.11-0ubuntu3.16):
```
root@worthy-sheep:~# systemctl is-enabled netplan
Failed to get unit file state for netplan.service: No such file or directory
root@worthy-sheep:~# echo $?
1
root@worthy-sheep:~# systemctl status netplan
Unit netplan.service could not be found.
root@worthy-sheep:~# echo $?
4
```

Ubuntu 24.04 LTS (systemd 255.4-1ubuntu8.10):
```
root@distinct-pika:~# systemctl is-enabled netplan
not-found
root@distinct-pika:~# echo $?
4
root@distinct-pika:~# systemctl status netplan
Unit netplan.service could not be found.
root@distinct-pika:~# echo $?
4
```

PR #571, opened 2026-02-19
Support multiple IPv6 address tokens (LP: #1950130)

lykinsbd:
This PR adds support for specifying multiple IPv6 address tokens as a sequence in addition to the existing scalar format. This allows configuring multiple IPv6 addresses with static interface identifiers.

## Changes

- **ABI Compatibility**: Added `ip6_addr_gen_tokens` array field at end of struct while keeping deprecated `ip6_addr_gen_token` field
- **Parser**: Updated to handle both scalar (single token) and sequence (multiple tokens) YAML nodes
- **networkd Generator**: Outputs multiple `IPv6Token=` lines for systemd-networkd
- **NetworkManager**: Uses first token only (NM limitation documented)
- **Validation**: Maintains mutual exclusivity with `ipv6-address-generation`
- **Tests**: Added 13 comprehensive tests covering scalar, array, validation, errors, and integration
- **Documentation**: Updated with examples and NetworkManager limitation note

## Test Results

- 13/14 meson tests pass (all functional tests)
- 1123/1126 pytest tests pass
- 3 failures are pre-existing snap/dbus environment issues unrelated to this change
- All 13 new IPv6 token tests pass

## Fixes

[LP: #1950130](https://bugs.launchpad.net/netplan/+bug/1950130)

### DIFF
--- a/doc/netplan-yaml.md
+++ b/doc/netplan-yaml.md
@@ -454,11 +454,21 @@ Match devices by MAC when setting options like: `wakeonlan` or `*-offload`.
   > Stateless Address Auto-configuration.
   > Possible values are `eui64` or `stable-privacy`.
 
-- **`ipv6-address-token`** (scalar) – since 0.100
+- **`ipv6-address-token`** (scalar or sequence) – since 0.100
 
-  > Define an IPv6 address token for creating a static interface identifier for
-  > IPv6 Stateless Address Auto-configuration. This is mutually exclusive with
-  > `ipv6-address-generation`.
+  > Define one or more IPv6 address tokens for creating static interface 
+  > identifiers for IPv6 Stateless Address Auto-configuration. This is mutually 
+  > exclusive with `ipv6-address-generation`.
+  >
+  > Multiple tokens (as a sequence) are supported since 1.3. When using 
+  > systemd-networkd as the back end renderer, multiple tokens can be specified 
+  > to configure multiple IPv6 addresses. NetworkManager only supports a single 
+  > token; if multiple tokens are specified, only the first will be used.
+
+  Examples:
+
+  - Single token: `ipv6-address-token: ::42`
+  - Multiple tokens: `ipv6-address-token: [::31, ::32, ::33]`
 
 - **`gateway4`**, **`gateway6`** (scalar)
 
--- a/src/abi.h
+++ b/src/abi.h
@@ -244,7 +244,7 @@ struct netplan_net_definition {
     GArray* address_options;
     gboolean ip6_privacy;
     guint ip6_addr_gen_mode;
-    char* ip6_addr_gen_token;
+    char* ip6_addr_gen_token; /* deprecated: use ip6_addr_gen_tokens for multiple tokens */
     char* gateway4;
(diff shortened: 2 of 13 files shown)


lykinsbd:
Fixed integration test failure - the test was expecting all 3 IPv6 addresses for NetworkManager, but NM only supports a single token (as documented). Updated the test to:
- Expect 1 address when using NetworkManager backend
- Expect 3 addresses when using networkd backend

This matches the actual implementation behavior.

benhoyt:
Hi @lykinsbd, thanks for the contribution. It looks reasonable at a high level. However, there's been a bit of flux on the netplan team here recently (and my team is getting ready for an internal conference next week), so this might take us a bit of time to get up to speed with and respond to properly. I just want to set expectations about timeline for proper review. A couple of questions in the meantime to help us build context:

- Can you please give me a bit more context or your use case for this change?
- It looks like with this change netplan will silently succeed when rendering to NetworkManager, even though it's only using the first address token and throwing away the others. It seems to me it'd be better if it hard-failed instead, similar (for example) to how activation-mode is [only allowed to be `manual`](https://github.com/canonical/netplan/blob/44f2803ee3be5389f3f86828f83cc2499fb0534e/src/nm.c#L739-L748) -- when rendering under NetworkManager, that's a hard fail. Or at the least a warning; I'm not yet sure which is best.

lykinsbd:
Hi @benhoyt, thanks for taking the time to review! No worries on the timeline - I completely understand.

**Use Case:**
My primary use case is in lab environments where we need multiple predictable IPv6 addresses on single interfaces. Being able to define them as a sequence allows us to quickly configure ranges like `::20-29` on one interface and `::30-39` on another, making it easy to organize and manage test infrastructure with consistent, memorable addresses. (To be clear configured as sequential tokens `::20`, `::21`, `::22`, and so on.)

**NetworkManager Behavior:**
You make an excellent point about the silent failure. I actually considered implementing full multiple token support in NetworkManager as well, but that seemed like a bigger problem for another time (would need to investigate NM's capabilities and potentially work around limitations). I completely agree that the current behavior of silently using only the first token is confusing and could lead to unexpected results.

I think adding validation that fails when multiple tokens are specified with NetworkManager would be the right approach, matching the pattern you referenced with `activation-mode`. I'm happy to implement either a hard error or a warning - both have merit. Let me know what you think makes the most sense and I'll update the PR accordingly.

benhoyt:
I've talked to a couple of folks here, and we'd prefer validate-and-warn with NetworkManager, rather than hard fail. So if you could update the PR with that change, we'll review and get it merged. Thanks!

lykinsbd:
Hi @benhoyt, thanks for the feedback! I've updated the PR to add a validate-and-warn when multiple IPv6 address tokens are specified with NetworkManager.

The warning is emitted during NM config generation (following the same pattern as the existing tunnel checksums/extensions warning on line 620 of `nm.c`):

```
engreen: NetworkManager does not support multiple IPv6 address tokens, only the first one will be used
```

The first token is still applied successfully — no hard failure. Test updated to verify the warning is emitted.

PR #575, opened 2026-04-08
state: label DHCPv4 using networkd ConfigSource

guilherme-puida:
## Description

`netplan status` did not label addresses as (dhcp) when systemd-networkd installs no dhcp protocol routes.

DHCPv6 already uses networkd's ConfigSource to apply the (dhcp) label. This patch extends this behavior to DHCPv4 as well, keeping the already existing route-based detection for DHCPv4 as a fallback.

LP: #2147446

## Checklist


guilherme-puida:

JegadishD (approved):

PR #576, opened 2026-04-08
tests: only consider netplan-generated files in integration tests

guilherme-puida:
## Description

Some files not managed by netplan can exist in those locations. Netplan tests should only consider files actually generated by netplan itself.

This fixes autopkgtest failures due to dracut-network installing /run/systemd/network/zzzz-dracut-default.network.

## Checklist

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -478,7 +478,7 @@ class IntegrationTestsBase(unittest.TestCase):
 
         # Check systemd-networkd files
         base_path = '/run/systemd/network'
-        files = glob.glob(f'{base_path}/*.network') + glob.glob(f'{base_path}/*.netdev')
+        files = glob.glob(f'{base_path}/*netplan*.network') + glob.glob(f'{base_path}/*netplan*.netdev')
         for file in files:
             res = os.stat(file)
             user = pwd.getpwuid(res.st_uid)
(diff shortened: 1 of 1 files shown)


guilherme-puida:

tests/integration/base.py
```diff
@@ -478,7 +478,7 @@ def assert_file_permissions(self):
 
         # Check systemd-networkd files
         base_path = '/run/systemd/network'
-        files = glob.glob(f'{base_path}/*.network') + glob.glob(f'{base_path}/*.netdev')
+        files = glob.glob(f'{base_path}/*netplan*.network') + glob.glob(f'{base_path}/*netplan*.netdev')
```
JegadishD: Making use of the fixed prefix as used in rest of Netplan code can make it consistent and provide a tighter match.
`glob.glob(f'{base_path}/10-netplan*.network') + glob.glob(f'{base_path}/10-netplan*.netdev')`

tests/integration/base.py
```diff
@@ -489,7 +489,7 @@ def assert_file_permissions(self):
 
         # Check Network Manager files
         base_path = '/run/NetworkManager/system-connections'
-        files = glob.glob(f'{base_path}/*.nmconnection')
+        files = glob.glob(f'{base_path}/*netplan*.nmconnection')
```
JegadishD: Similarly, the wildcard prefix can be removed.
`glob.glob(f'{base_path}/netplan*.nmconnection')`

JegadishD (commented):
@guilherme-puida Thanks for the PR!

tests/integration/base.py
guilherme-puida: Ack, applied.

tests/integration/base.py
guilherme-puida: Also applied.

JegadishD (approved):

PR #577, opened 2026-04-08
gen-networkd: wpa supplicant unit requires netplan-configure.service

guilherme-puida:
## Description

wpa_supplicant reads its configuration from
/run/netplan/wpa-<iface>.conf, which is written by netplan-configure.service. Without an explicit Requires=/After= dependency, wpa_supplicant could start before the config file is available, causing it to fail.

LP: #2145061

## Checklist


guilherme-puida:

JegadishD (approved):
Thanks for the fix!

PR #579, opened 2026-04-10
chore: add testing example for local build

Tony-WLB:
This pull request updates the `README.md` to provide instructions for testing a local Netplan, build with the `networkd` backend. The new section explains how to set up environment variables to run the CLI from the build tree and includes a safe example configuration for testing.

Testing and usage documentation:

* Added a new section explaining how to test a local build by setting `NETPLAN_GENERATE_PATH`, `NETPLAN_CONFIGURE_PATH`, `LD_LIBRARY_PATH`, and `PYTHONPATH`, allowing the Netplan CLI to run from the build tree without affecting the system install.
* Provided an example YAML configuration using `route-metric` for safe, non-disruptive testing, including instructions for verifying the results.
* Clarified the use of environment variables with `sudo`, noting the effect of `env_reset` in `/etc/sudoers`. 

A drive-by fix:

* Rename file `CONTRIBUTING` to `CONTRIBUTING.md`.


README.md
```diff
@@ -48,6 +48,49 @@ Steps to build Netplan using the [Meson](https://mesonbuild.com) build system in
 * meson test -C build --verbose [TEST_NAME]
 * meson install -C build --destdir ../tmproot
 
+# Test a local build (backend: networkd)
```
benhoyt: Seeing there's a nice Makefile that wraps Meson with the correct options, I think we should probably lean into that and update the section above to use the Makefile commands. It looks like the default DESTDIR is `../tmproot`, too.

For example, the first two (meson setup and meson compile) are covered by the default make target, so just typing `make`. Then test is `make check` and install is `make install`.

benhoyt: There's also a lot of build dependencies that need to be installed. I had to install these (I don't think I missed any) in addition to whatever was already on my system:

```
sudo apt install \
	meson \
	libyaml-dev \
	libsystemd-dev \
	pyflakes3 \
	pycodestyle \
	python3-coverage \
	pandoc \
	python3-cffi \
	libcmocka-dev \
	python3-dev
```

I think we should mention these steps too, from checkout to build.

Not for this PR, but soon I think we should switch to using `pyproject.toml` and `uv` to make development a lot easier. And/or using devcontainers to avoid messing so much with the local environment.

README.md
```diff
@@ -48,6 +48,49 @@ Steps to build Netplan using the [Meson](https://mesonbuild.com) build system in
...
+
+After `meson install -C _build --destdir ../tmproot`, test the local build without touching the system install by passing environment variables that redirect the CLI to your build tree (`NETPLAN_GENERATE_PATH`, `NETPLAN_CONFIGURE_PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH`). These are needed because the Python CLI resolves binary and library paths at runtime.
+
+Prepare a test config (replace `eth0` with an interface present on your system):
```
benhoyt: I think instead of using a modifying command like `netplan try`, we should just drop this test config and show `netplan info` or something simpler as the example.

We could add a `make run` target which allowed ARGS to be passed, like `make run ARGS='info'` would be like `netplan info`. That could set all the environment variables.

Or if that's too ugly we could add a `netplan` script at the top level which sets the env vars and then calls `tmproot/usr/sbin/netplan`.

benhoyt (commented):
Thanks. I think we can also clean up the "Build using Meson" section, and simplify the environment variables (putting them in the Makefile or a script). Let's lean heavily on the Makefile where possible, so devs don't have to remember or copy-n-paste long command lines.

Probably roll it into a single section "Local development". If you want to chat further, let's chat tomorrow.

Tony-WLB:
A section on the Makefile has been added, while the Meson section remains for now. We might consider removing it later. 

Additionally, the example used in testing has been changed to a non-modifying command for simplicity.

README.md
```diff
@@ -41,13 +41,101 @@ If you face issues, refer to our [comprehensive contribution guide](https://netp
 
 # Build using Meson
 
+A Makefile wrapper is also provided for simplified usage. For that approach, please refer to the [Build using Makefile][makefile-section] section below.
```
benhoyt: Should this be:
```suggestion
A Makefile wrapper is also provided for simplified usage. For that approach, please refer to the [Build using Makefile](#build-using-makefile) section below.
```

README.md
```diff
@@ -41,54 +41,99 @@ If you face issues, refer to our [comprehensive contribution guide](https://netp
...
+- `make` or `make default`  
+  Set up the build directory (`_build`) and build the project
+
+- `make check`  
+  Build and run all tests in `_build`
+
+- `make linting`  
+  Run Meson `linting` and `codestyle` test targets
+
+- `make check-coverage`
```
benhoyt: I think we should drop this one (check-coverage), as it doesn't seem generally useful enough to include here.

README.md
```diff
@@ -41,54 +41,99 @@ If you face issues, refer to our [comprehensive contribution guide](https://netp
...
+
+- `make install [DESTDIR=../tmproot]`  
+  Build and install into a staging root (defaults to `../tmproot`)
+
+- `make clean`  
+  Remove generated build and test artifacts
 
-After `meson install -C _build --destdir ../tmproot`, test the local build without touching the system install by passing environment variables that redirect the CLI to your build tree (`NETPLAN_GENERATE_PATH`, `NETPLAN_CONFIGURE_PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH`). These are needed because the Python CLI resolves binary and library paths at runtime.
+- `make run ARGS='<command>'`  
+  Run the locally built CLI with the appropriate environment, e.g.:
```
benhoyt: Nit, let's change to:

```suggestion
  Run the locally built netplan CLI with the appropriate environment, for example, to run `netplan help`:
```

README.md
```diff
@@ -41,13 +41,101 @@ If you face issues, refer to our [comprehensive contribution guide](https://netp
...
+```
+
+This wrapper sets the required environment variables automatically:
+
+`NETPLAN_GENERATE_PATH`
+`NETPLAN_CONFIGURE_PATH`
+`LD_LIBRARY_PATH`
+`PYTHONPATH`
+
+These are needed because the Python CLI resolves binary and library paths at runtime.
```
benhoyt: Let's not list them all, otherwise we have to update this list if they change. How about shortening to:
```suggestion
This wrapper sets the required environment variables (such as `NETPLAN_GENERATE_PATH`) automatically. These are needed because the Python CLI resolves binary and library paths at runtime.
```

README.md
```diff
@@ -41,13 +41,101 @@ If you face issues, refer to our [comprehensive contribution guide](https://netp
...
+  - eswitch-mode
+  - infiniband
+  - regdom
+  - vrf
+  - vxlan
+  - virtual-ethernet
+  - dbus-config
+  - generated-suppl
+```
+
```
benhoyt: Let's tighten this up. We've already mentioned `make` and `make install` above, so we can drop that. Also, let's not include the full output, just the first few lines and `...`. Let's do something like this:
```suggestion
As an example, let's use `make run` to run `netplan info`:

```sh
$ make run ARGS='info'
netplan.io:
  website: "https://netplan.io/"
  features:
  - dhcp-use-domains
  - auth-phase2
  ...
```

```

File: README.md [hunk 8]
```diff
@@ -60,3 +148,4 @@ Our mailing list is [here](https://lists.launchpad.net/netplan-developers/).
 
 Email the list at [netplan-developers@lists.launchpad.net](mailto:netplan-developers@lists.launchpad.net).
 
+[makefile-section]: #build-using-makefile
```
benhoyt: Oh, I see what you've done here. Let's just put the link inline.

benhoyt (commented):
Thanks for the updates. Yeah, this is a lot closer.

benhoyt:
I added a few minor comments, and then you should also address https://github.com/canonical/netplan/pull/579#discussion_r3070938054 -- adding a node about installing the development dependencies.

Makefile
```diff
@@ -51,3 +52,10 @@ check-coverage: pre-coverage
 install: default
 	meson install -C _build --destdir $(DESTDIR)
 
+run:
+	sudo \
```
benhoyt: Let's put an `@` in front here to avoid the noisy echoed output:
```suggestion
	@sudo \
```

benhoyt (commented):
Thanks! One minor nit, and let's also add a `/tmproot` line to our `.gitignore` to avoid git picking up all those files.

Makefile
Tony-WLB: Yeah, that would be clean.

benhoyt (approved):

PR #580, opened 2026-04-12
nm: fix filename overflow for WiFi SSIDs with multi-byte characters

rone-oliver:
## Description

SSIDs containing multi-byte UTF-8 characters (e.g. emojis) are stored
by NetworkManager in a semicolon-delimited decimal-byte format
(e.g. `240;159;154;128`). When `g_uri_escape_string()` percent-encodes
this string, each semicolon becomes `%3B`, expanding a 20-emoji SSID
from ~160 bytes to ~507 bytes — well beyond the 255-byte NAME_MAX
limit. NetworkManager then crashes with "File name too long".

**Fix:** In `netplan_netdef_get_output_filename()` and in
`write_nm_conf_access_point()`, check the candidate basename length
against `NAME_MAX` before writing. If it would overflow, replace the
percent-encoded SSID with its SHA-256 hex digest (always 64 chars →
91-byte basename). The reverse lookup in
`netplan_get_id_from_nm_filepath()` is updated to fall back to the
hashed suffix when the escaped form isn't found.

Two new ctests cover the normal path and the long-SSID hashing path.

Fixes LP: #2147259

## Checklist

- [x] New/changed keys in YAML format are documented. *(N/A — no new YAML keys)*
- [x] (Optional) Closes an open bug in Launchpad. LP: #2147259


rone-oliver:
CI notes for reviewers:

Coverage (Unit tests & Coverage): The overall C coverage shows ~96%, but this is a pre-existing baseline on main before this PR (the large gap is in gen-openvswitch.c, parse.c, etc., which are unrelated to this fix). The two files this PR touches — src/nm.c and src/util.c — are both at 100% coverage after the test commits.

Autopkgtest / RPM build / Spread failures: These three checks are failing with the same errors on other recent merged PRs (infra/environment issues, not code regressions). They are not caused by this patch.

rone-oliver:
I've documented the full reproduction steps and the technical impact of this bug on the end-user experience here: https://medium.com/@roneootan611/ubuntu-wi-fi-icon-disappearing-how-a-simple-emoji-can-crash-your-entire-network-stack-5dac8e8c5025.

rone-oliver:
Thanks for checking it out! Let me know if you need any context when you get a chance to review. @benhoyt

PR #584, opened 2026-04-23
ci: replace freedesktop links with man7

Tony-WLB:
Freedesktop links have enabled "AI scraper block / go-away page" protection, which can occasionally cause CI failures.

Documentation link updates:

* Updated the `busctl` reference link in `doc/dbus-config.md` to use the man7.org man page
* Updated the `systemd-generator` reference link in `doc/generator.md` to use the man7.org man page


benhoyt (approved):
Great, thanks for this.

PR #585, opened 2026-04-27
ci: Retry `apt update` on failure during LXD setup

Tony-WLB:
Transient network issues can cause `apt update` to fail sporadically. Adding a retry mechanism makes the CI step more resilient.

* Upgraded the `actions/checkout` step from version 2 to version 4 for improved performance and security.

* Added configuration to set apt retries (`Acquire::Retries "3";`) and HTTP timeout (`Acquire::http::Timeout "20";`) during the dependency installation step, making package installations more resilient to transient network issues.
* Ensured the same apt retry and timeout settings are applied inside the LXD test environment by adding them as `--setup-commands` to the `autopkgtest` invocation.

### DIFF
--- a/.github/workflows/autopkgtest.yml
+++ b/.github/workflows/autopkgtest.yml
@@ -21,7 +21,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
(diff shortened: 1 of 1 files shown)


JegadishD:
@Tony-WLB The autopkgtest is passing now. I am closing the PR but I shall keep this PR in context and use in future if required. Thanks!

PR #589, opened 2026-05-25
Setup asan run's prefix to ensure openvswitch binaries are in the path

JegadishD:
## Description

Setup asan run's prefix to ensure openvswitch binaries are in the path

## Checklist

- [ &check;] Runs `make check` successfully.
- [ &check;] Retains code coverage (`make check-coverage`).

### DIFF
--- a/tools/run_asan.sh
+++ b/tools/run_asan.sh
@@ -7,10 +7,10 @@ BUILDDIR="_leakcheckbuild"
 CLEANBUILDDIR="_cleanbuild"
 CC=gcc
 
-meson setup ${BUILDDIR} -Db_sanitize=address,undefined
+meson setup ${BUILDDIR} -Db_sanitize=address,undefined -Dprefix=/usr
 meson compile -C ${BUILDDIR} --verbose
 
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Thanks.

Separately, we should probably settle on a style for the commit message (which should come from the PR description title). We use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) in some of our other projects (and Charm Tech's projects). So this one would probably have a `ci: ...` prefix:

`ci: set run_asan's prefix so openvswitch binaries are in path`

JegadishD:
> Thanks.
> 
> Separately, we should probably settle on a style for the commit message (which should come from the PR description title). We use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) in some of our other projects (and Charm Tech's projects). So this one would probably have a `ci: ...` prefix:
> 
> `ci: set run_asan's prefix so openvswitch binaries are in path`

Thanks @benhoyt for the review. The commit-message suggestion makes sense. And going forward, I shall make use of this style.

PR #591, opened 2026-05-29
fix(ci): fix autopkgtest CI failures

JegadishD:
- Fixed ethtool sub-feature name mismatch. Launchpad ticket LP#: 2126938
- Skipped vxlan tunnel test due to iproute2 vxlan JSON bug. Launchpad ticket  LP#: 2124257
- Removed temporary workaround to copy debian cloud-init.sh that is no longer required

## Checklist

- [X] Retains code coverage (`make check-coverage`).
- [X] \(Optional\) Closes an open bug in Launchpad. LP#: 2126938


benhoyt (approved):
Looks reasonable to me, thanks. Two minor things in the PR title / description to tweak before we merge:

- Let's standardize on starting the commit title with an active/imperative verb, as this is pretty standard Git these days. Let's also standardize on lowercase first letter ([example from Linux kernel](https://github.com/torvalds/linux/commit/9c7da87c2dc860bb17ca1ece942495d28b1ce3b9)). So in this case: `fix(ci): fix autopkgtest ...`
- In the PR description, we should remove the big heading "Description", as that's obvious from context, and avoid the repeated title.
- In the checklist, there's something funky going on with the checkboxes. We should `[x]` rather than `[&check;]` as that will render correctly.

PR #597, opened 2026-06-23
fix: added fix to handle exception on file permission issue

JegadishD:
With NetworkManager as backend, the netplan apply command was leading to uncaught exception and triggering Apport crash report. The crash was due to uncaught permission exception thrown while accessing network configuration files. The added fix catches exception and exits with the corresponding error code.

Tests: The issue was recreated in a VM with NetworkManager as backend. Later the fixed repository code was installed in the same VM and the exception backtrace was not seen.

Closes  LP#2154081


benhoyt (approved):
Looks good to me, thanks. Given there are no integration tests, can you please write in the PR description what manual testing you've done?

JegadishD:
> Looks good to me, thanks. Given there are no integration tests, can you please write in the PR description what manual testing you've done?

I have added the testing details. Thanks for the review!

PR #598, opened 2026-06-30
fix(ci): Added setuptool package to the rpm build file

JegadishD:
The build depends on distutils package. In python <3.12, this was available as part of standard library. In the python version >=3.12, setuptools package provides distutils. As the fedora releases moved to newer python version, the build started failing.


benhoyt (approved):
"All checks have passed" -- love seeing that!

PR #599, opened 2026-07-03
Extend 'critical' to accept KeepConfiguration scalar values

zakthan:
## Description

Allow the `critical` key in netplan YAML to accept not just boolean values but also
systemd-networkd `KeepConfiguration=` scalar values: `static`, `dhcp`, and `dhcp-on-stop`.

This is backwards compatible: boolean values (`true`/`false`/`yes`/`no`) continue to work
as before, mapping to `KeepConfiguration=true`.

### New scalar values

| netplan YAML | Generated networkd config |
|---|---|
| `critical: true` | `KeepConfiguration=true` |
| `critical: static` | `KeepConfiguration=static` |
| `critical: dhcp` | `KeepConfiguration=dhcp` |
| `critical: dhcp-on-stop` | `KeepConfiguration=dhcp-on-stop` |
| `critical: false` | *(no KeepConfiguration line)* |

### Motivation

`KeepConfiguration=static` is critical for environments running BIRD/BGP alongside
systemd-networkd. When `networkctl reload` is triggered:

- `KeepConfiguration=true` causes networkd to attempt reconciliation of foreign (BIRD) routes,
  triggering a full reconfigure cycle that deletes and re-adds all routes (~2.2s disruption)
- `KeepConfiguration=static` precisely preserves only `proto static` routes and ignores
  foreign routes, enabling clean delta-only reloads with zero route disruption

Currently there is no way to set `KeepConfiguration=static` via netplan — the only option is
`critical: true` which maps to `KeepConfiguration=true`.

### Changes

- `src/abi.h`: New `NetplanKeepConfiguration` enum
- `src/parse.c`: New `handle_critical()` parser accepting booleans + scalar values
- `src/networkd.c`: Emit correct `KeepConfiguration=` value
- `src/netplan.c`: Serialize enum back to YAML
- `src/types.c`: Backwards-compatible bool getter + new enum getter
- `python-cffi/`: New `KeepConfiguration` Python enum and `.keep_configuration` property
- `doc/netplan-yaml.md`: Updated documentation
- Tests: 4 new generator tests + 1 new libnetplan test

### Supersedes

This is a clean re-implementation of #409 (which has been stale since Sept 2024 with
unresolved merge conflicts) on top of current main.

## Checklist

- [x] Retains backwards compatibility (boolean values still work)
- [x] New generator tests added
- [x] New libnetplan Python API tests added
- [x] Documentation updated
- [ ] Runs `make check` successfully (no build environment available locally)

### DIFF
--- a/tests/generator/test_common.py
+++ b/tests/generator/test_common.py
@@ -478,6 +478,95 @@ LinkLocalAddressing=ipv6
 KeepConfiguration=true
 '''})
 
+    def test_critical_static(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: static
+''')
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+KeepConfiguration=static
+'''})
+
+    def test_critical_dynamic(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: dynamic
+''')
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+KeepConfiguration=dynamic
+'''})
+
+    def test_critical_dynamic_on_stop(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: dynamic-on-stop
+''')
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+KeepConfiguration=dynamic-on-stop
+'''})
+
+    def test_critical_dhcp_alias(self):
+        """dhcp/dhcp-on-stop are accepted as legacy aliases for dynamic/dynamic-on-stop"""
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: dhcp
+''', skip_generated_yaml_validation=True)
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+KeepConfiguration=dynamic
+'''})
+
+    def test_critical_false(self):
+        self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: false
+''')
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+'''})
+
+    def test_critical_invalid(self):
+        err = self.generate('''network:
+  version: 2
+  ethernets:
+    engreen:
+      critical: invalid-value
+''', expect_fail=True)
+        self.assertIn("invalid value 'invalid-value' for critical", err)
+
     def test_dhcp_identifier_mac(self):
         self.generate('''network:
   version: 2
--- a/tests/test_libnetplan.py
+++ b/tests/test_libnetplan.py
@@ -618,6 +618,28 @@ class TestNetDefinition(TestBase):
         self.assertTrue(state['eth0'].critical)
         self.assertFalse(state['eth1'].critical)
 
+    def test_keep_configuration(self):
+        state = state_from_yaml(self.confdir, '''network:
+  ethernets:
+    eth0:
+      critical: static
+    eth1:
+      critical: true
+    eth2:
+      critical: dynamic
+    eth3: {}''')
+
+        from netplan.netdef import KeepConfiguration
+        self.assertEqual(state['eth0'].keep_configuration, KeepConfiguration.STATIC)
+        self.assertEqual(state['eth1'].keep_configuration, KeepConfiguration.TRUE)
+        self.assertEqual(state['eth2'].keep_configuration, KeepConfiguration.DYNAMIC)
+        self.assertEqual(state['eth3'].keep_configuration, KeepConfiguration.FALSE)
+        # Backwards-compatible bool check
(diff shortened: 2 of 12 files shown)


JegadishD:
@zakthan Thanks for the contribution!

From the comments and links to the previous PR, I get a view of the issue and your need for having this support in Netplan. I shall plan to have this PR reviewed in this cycle.
Due to staffing transitions and a following gap, there is an accumulated backlog being worked on by the new staff. I am mentioning this to set up reasonable expectations on the review timeline.

Meanwhile, from an initial analysis, the following are points that needs your look in.
- There are two CI/CD failures
- The newer systemd uses `dynamic` and `dynamic-on-stop` instead of `dhcp` and `dhcp-on-stop`

zakthan:
hi @JegadishD /all. could you provide any estimation/feedback for this? Thanks!

tests/generator/test_common.py
```diff
@@ -478,6 +478,95 @@ def test_dhcp_critical_true(self):
...
+''', skip_generated_yaml_validation=True)
+
+        self.assert_networkd({'engreen.network': '''[Match]
+Name=engreen
+
+[Network]
+LinkLocalAddressing=ipv6
+KeepConfiguration=dynamic
+'''})
+
```
JegadishD, line 562: Please add test for dhcp-on-stop

tests/test_libnetplan.py
```diff
@@ -618,6 +618,28 @@ def test_critical(self):
...
+    def test_keep_configuration(self):
+        state = state_from_yaml(self.confdir, '''network:
+  ethernets:
+    eth0:
+      critical: static
+    eth1:
+      critical: true
+    eth2:
+      critical: dynamic
+    eth3: {}''')
```
JegadishD: Please add checks for dynamic-on-stop, dhcp and dhcp-on-stop

JegadishD (changes requested):
Hi @zakthan The changes look good. I have added a couple of minor comments related to tests. I shall merge the PR after seeing the updated version. 

Also, Please push after rebasing on top of main. The CI workflows have recently been moved to use Ubuntu-Resolute. Though your changes should not fail with the recent CI changes, it is better to confirm before merging.

tests/generator/test_common.py
zakthan: Added `test_critical_dhcp_on_stop_alias` covering `critical: dhcp-on-stop` → `KeepConfiguration=dynamic-on-stop`.

It needs `skip_generated_yaml_validation=True` (same as the existing `dhcp` alias test) because the serializer normalises the legacy alias to the canonical `dynamic-on-stop`, so a YAML roundtrip comparison can never match by design.

tests/test_libnetplan.py
zakthan: Extended `test_keep_configuration` to cover all remaining values:

- `eth4: dynamic-on-stop` → `KeepConfiguration.DYNAMIC_ON_STOP`
- `eth5: dhcp` (legacy alias) → `KeepConfiguration.DYNAMIC`
- `eth6: dhcp-on-stop` (legacy alias) → `KeepConfiguration.DYNAMIC_ON_STOP`
- `eth7: false` → `KeepConfiguration.FALSE`

Each also asserts the backwards-compatible `.critical` boolean.

zakthan:
Thanks for the review @JegadishD! All three points are addressed in `17bb1fe`:

1. **`dhcp-on-stop` test** — added `test_critical_dhcp_on_stop_alias` in `tests/generator/test_common.py`.
2. **Additional libnetplan checks** — `test_keep_configuration` now covers `dynamic-on-stop`, `dhcp`, `dhcp-on-stop` and explicit `false`, including the backwards-compatible `.critical` boolean.
3. **Rebased on top of main** — replayed onto `ba58953`, which includes the Ubuntu Resolute CI migration (#601). No conflicts; the branch remains a single commit.

Verified locally against the Resolute-era build:

- full unit + legacy test suites pass
- `src/networkd.c` and `src/types.c` at 100% coverage; the `critical` parsing block in `src/parse.c` is fully covered
- ABI check passes against the new reference: `abidiff abi-compat/resolute_1.2.xml ... --no-added-syms` reports 0 removed / 0 changed

Let me know if you would like anything else adjusted.

zakthan:
Status update: **17 of 18 checks are green**, including `test-and-coverage`, `build` / ABI compatibility, `memory-sanitizer`, `spread`, CodeQL, all RPM builds and the autopkgtest jobs.

The single remaining failure is `Run documentation checks`, and it is **not caused by this PR**. It is the Sphinx `linkcheck` stage failing on two external URLs:

| Link | Location | Error |
|---|---|---|
| `https://askubuntu.com/questions/tagged/netplan` | `doc/index.md:70` | `403 Client Error: Forbidden` (bot protection) |
| `https://documentation.ubuntu.com/lxd/en/latest/installing/` | `doc/netplan-tutorial.md:19` | read timeout after 30s |

Evidence that it is pre-existing and unrelated:

- This PR only touches `doc/netplan-yaml.md`. Neither `doc/index.md` nor `doc/netplan-tutorial.md` is modified.
- The same workflow currently fails on `main`:
  - [run 33044097139](https://github.com/canonical/netplan/actions/runs/33044097139) (`d059585`)
  - [run 32452781991](https://github.com/canonical/netplan/actions/runs/32452781991) (`4e3bae3`)
- This check **passed on this very PR on 2026-07-06** with the same documentation content, so the failures are intermittent and depend on external rate limiting rather than on the diff.

I have deliberately not added `linkcheck_ignore` entries for these URLs, since that would be an unrelated change inside a feature PR. Happy to open a separate PR against `main` for that if you would find it useful, or to add it here if you prefer.

JegadishD (approved):
@zakthan 
Thank you for your time and effort. Your contribution is much appreciated.

PR #601, opened 2026-07-16
ci: migrate workflows to Ubuntu Resolute 26.04 and fix test regressions

JegadishD:
The base OS upgrade introduces a newer Meson version that no longer passes stdin to test scripts, surfacing a test coverage drop and exposing an inconsistency in terminal stdin handling.

- Fix terminal.py to consistently use the object's fd context for stdin access
- Update test script to create and pass a pty as stdin to the terminal object
- Remove obsolete packages and private repository dependencies that are no longer needed.
- Apply required configuration adjustments to support the Resolute baseline


.github/workflows/debci.yml
```diff
@@ -29,9 +29,8 @@ jobs:
       # it's needed (will be auto-loaded) by routing.test_vrf_basic
       - name: Install dependencies
         run: |
-          sudo add-apt-repository -y -n -s ppa:slyon/netplan-ci
           sudo apt update
-          sudo apt install debci lxc lxc-templates debian-archive-keyring autopkgtest ubuntu-dev-tools devscripts linux-modules-extra-$(uname -r) #openvswitch-switch
+          sudo apt install debci lxc lxc-templates debian-archive-keyring autopkgtest ubuntu-dev-tools devscripts #openvswitch-switch
```
benhoyt, line 31: Not your change, but do you know why `openvswitch-switch` is commented out here? Should we just remove it if it's not needed in this workflow?

.github/workflows/network-manager.yml
```diff
@@ -56,19 +54,19 @@ jobs:
...
         run: |
           sudo patch /usr/share/autopkgtest/lib/build-lxd.sh .github/workflows/snapd.patch
-          MIRROR=http://archive.ubuntu.com/ubuntu autopkgtest-build-lxd ubuntu-daily:noble # LP: #2052639
+          MIRROR=http://archive.ubuntu.com/ubuntu autopkgtest-build-lxd ubuntu-daily:resolute # LP: #2052639
       - name: Run autopkgtest
         run: |
           # When we introduce ABI breaking changes, we need to rebuild NM in DEP-8
-          #pull-lp-source network-manager noble
-          sudo autopkgtest -U \
+          #pull-lp-source network-manager resolute
```
benhoyt, line 66: Do you know why this line is commented out, or what the point of it is?

benhoyt (approved):
Thanks for this -- looks good!

I notice there are still a couple of references to "noble" in abi-compat/README.md and workflows/build-abi.yml -- do those need to be updated too?

Added a couple of minor comments -- not blocking.

.github/workflows/debci.yml
JegadishD: The OVS tests are allowed to skip in this workflow. The autopkgtest command used later in this file mentions this. Probably the author's intention is to optionally enable it during specific tests - serving as a reminder when the situation arises.

.github/workflows/network-manager.yml
JegadishD: The comment mentioned above gives some context. In my understanding, when there is an ABI change, it needs to be enabled and tested with the pulled network-manager source.

JegadishD:
> Thanks for this -- looks good!
> 
> I notice there are still a couple of references to "noble" in abi-compat/README.md and workflows/build-abi.yml -- do those need to be updated too?
> 
> Added a couple of minor comments -- not blocking.

As there was no ABI changes (diff below), my initial plan was to leave this file as is.
```
ubuntu@dev-general:~/repo/netplan$ abidiff abi-compat/noble_1.1.xml _build/src/libnetplan.so.1 \
  --headers-dir2 include/ --header-file2 src/abi.h \
  --no-added-syms
Functions changes summary: 0 Removed, 0 Changed (63 filtered out), 0 Added (7 filtered out) functions
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable
```
But on looking at git logs, I see that this file was updated and renamed as noble_1.1.xml from jammy*.xml. Following the earlier convention to update it on LTS release, I have updated and renamed the file as resolute_1.2.xml

I have made corresponding changes and updated the PR. Please review.

meson.build
```diff
@@ -1,5 +1,5 @@
 project('netplan', 'c',
-        version: '1.1',
+        version: '1.2',
```
benhoyt: For visibility, it might be better to put this version change in a separate PR, rather than stuffing it in this PR designated "ci".

benhoyt (approved):
Excellent, thanks for this. One comment about pulling out the version bump into another PR, but other than that, looks good to me.

meson.build
JegadishD: I have updated the PR to change it back as 1.1. I shall raise a separate PR for the same. Thanks for the review!

PR #604, opened 2026-07-30
parse-nm: Mark dhcp4/dhcp6 as dirty when method=manual (LP: #2160170)

JegadishD:
When parsing NM keyfiles with method=manual, explicitly mark dhcp4/dhcp6 as dirty so the YAML serializer outputs "dhcp4: false" / "dhcp6: false".

This fixes an issue in multi-file setups where an earlier YAML config with dhcp4:true would not be overridden by a NM keyfile with method=manual, because the false value was not being written to the netplan file.

Also update test_keyfile_method_manual to verify the explicit false output.

### DIFF
--- a/.github/workflows/lp2160170-nm-netplan.patch
+++ b/.github/workflows/lp2160170-nm-netplan.patch
@@ -1,0 +1,13 @@
+--- a/debian/tests/nm-netplan.py
++++ b/debian/tests/nm-netplan.py
+@@ -266,8 +266,8 @@ class TestNetplan(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


JegadishD:
There is a NetworkManager specific patch included in this PR. The patch is not being proposed for merge to Network Manager at this point in time. My line of thought is, When a new Netplan release is tagged, the patch can also be raised with the Network Manager repository.

benhoyt (approved):
We talked through this and I'm happy with the rationale/context as well as the fix.

PR #606, opened 2026-08-19
cli: handle insufficient user privilege error gracefully

JegadishD:
Insufficient privileges is leading to unhandled exception and these trigger apport crash reports. The fix is to check the return code and exit cleanly with an error message.

On similar lines, the netplan try's revert() is modified to catch exceptions rather than letting it propagate as an unhandled exception.

[LP: #2161924](https://bugs.launchpad.net/netplan/+bug/2161924)

### DIFF
--- a/netplan_cli/cli/commands/apply.py
+++ b/netplan_cli/cli/commands/apply.py
@@ -134,7 +134,12 @@ class NetplanApply(utils.NetplanCommand):
                     raise ConfigurationError("the configuration could not be generated")
             # Running 'systemctl daemon-reload' will re-run the netplan systemd generator.
             logging.debug('executing Netplan systemd-generator via daemon-reload')
-            utils.systemctl_daemon_reload()
+            res = utils.systemctl_daemon_reload()
+            if res != 0:
+                if exit_on_error:
+                    sys.exit(res)
+                else:
+                    raise RuntimeError("systemctl daemon-reload call failed: error %d" % res)
 
         devices = utils.get_interfaces()
 
--- a/netplan_cli/cli/utils.py
+++ b/netplan_cli/cli/utils.py
@@ -170,7 +170,7 @@ def systemctl_is_installed(unit_pattern):
 
 def systemctl_daemon_reload():
     '''Reload systemd unit files from disk and re-calculate its dependencies'''
-    subprocess.check_call(['systemctl', 'daemon-reload', '--no-ask-password'])
+    return subprocess.call(['systemctl', 'daemon-reload', '--no-ask-password'])
 
 
 def ip_addr_flush(iface):
--- a/tests/integration/regressions.py
+++ b/tests/integration/regressions.py
@@ -38,6 +38,36 @@ class _CommonTests():
             f.write('''''')
         self.generate_and_settle([])
 
+    def test_unprivileged_user_no_traceback(self):
+        """LP#2161924: netplan commands should not trigger sys.excepthook
+        (which would cause apport crash reports) when run without privileges."""
+
+        # Create a temporary user for testing the scenario of a regular user (non-root)
+        # running netplan commands without sudo privileges.
+        test_user = 'netplantest'
+        subprocess.run(['userdel', '-rf', test_user],
+                       stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+        subprocess.run(['useradd', '-m', '-s', '/bin/sh', test_user], check=True)
+        self.addCleanup(subprocess.run, ['userdel', '-rf', test_user],
+                        stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+
+        for command in ['generate', 'apply', 'try']:
+            with self.subTest(command=command):
+                result = subprocess.run(
+                    ['runuser', '-u', test_user, '--',
+                     'netplan', command],
+                    capture_output=True,
+                    text=True,
+                    timeout=30
+                )
+                # Check that no Python traceback was printed to stderr
+                # (tracebacks indicate unhandled exceptions that trigger apport)
+                self.assertNotIn(
+                    'Traceback (most recent call last):',
+                    result.stderr,
+                    f"netplan {command} produced a traceback (unhandled exception):\n"
+                    f"STDERR: {result.stderr}")
+
 
 @unittest.skipIf("networkd" not in test_backends,
                  "skipping as networkd backend tests are disabled")
--- a/netplan_cli/cli/commands/generate.py
+++ b/netplan_cli/cli/commands/generate.py
@@ -124,7 +124,10 @@ class NetplanGenerate(utils.NetplanCommand):
                 # automatically reloads systemd, as we might have changed
                 # service units, such as
                 # /run/systemd/generator.late/systemd-networkd-wait-online.service.d/10-netplan.conf
-                utils.systemctl_daemon_reload()
+                res = utils.systemctl_daemon_reload()
+                if res != 0:
+                    logging.error("systemctl daemon-reload call failed: error %d", res)
+                    sys.exit(res)
 
             logging.debug('command configure: running %s', argv)
             res = subprocess.call(argv)


netplan_cli/cli/commands/apply.py
```diff
@@ -134,7 +134,12 @@ def command_apply(self, run_generate=True, sync=False, exit_on_error=True, state
                     raise ConfigurationError("the configuration could not be generated")
             # Running 'systemctl daemon-reload' will re-run the netplan systemd generator.
             logging.debug('executing Netplan systemd-generator via daemon-reload')
-            utils.systemctl_daemon_reload()
+            res = utils.systemctl_daemon_reload()
+            if (res != 0):
```
benhoyt: Python style nit:
```suggestion
            if res != 0:
```

netplan_cli/cli/commands/try_command.py
```diff
@@ -135,16 +135,22 @@ def setup(self):  # pragma: nocover (requires user input)
...
+            # backup the state we just tried to apply
+            tempdir = tempfile.mkdtemp()
+            confdir = os.path.join(tempdir, 'etc', 'netplan')
+            os.makedirs(confdir)
+            self.config_manager.copy_tree('/etc/netplan', confdir, dirs_exist_ok=True)
+            # restore previous state
+            self.config_manager.revert()
+            NetplanApply().command_apply(run_generate=False, sync=True, exit_on_error=False, state_dir=tempdir)
+        except Exception as e:
+            logging.error("Revert failed with exception: %s", e)
```
benhoyt: Per discussion, let's move the `try/except Exception` handling up to the caller, so all the exception handling and printing of errors is in one place.

netplan_cli/cli/commands/try_command.py
```diff
@@ -135,16 +135,22 @@ def setup(self):  # pragma: nocover (requires user input)
...
-        tempdir = tempfile.mkdtemp()
-        confdir = os.path.join(tempdir, 'etc', 'netplan')
-        os.makedirs(confdir)
-        self.config_manager.copy_tree('/etc/netplan', confdir, dirs_exist_ok=True)
-        # restore previous state
-        self.config_manager.revert()
-        NetplanApply().command_apply(run_generate=False, sync=True, exit_on_error=False, state_dir=tempdir)
-        # clear the backup
-        shutil.rmtree(tempdir)
+        tempdir = None
```
benhoyt: Let's move this change to another isolated PR with its own tests.

netplan_cli/cli/utils.py
```diff
@@ -170,7 +170,7 @@ def systemctl_is_installed(unit_pattern):
 
 def systemctl_daemon_reload():
     '''Reload systemd unit files from disk and re-calculate its dependencies'''
-    subprocess.check_call(['systemctl', 'daemon-reload', '--no-ask-password'])
+    return subprocess.call(['systemctl', 'daemon-reload', '--no-ask-password'])
```
benhoyt: All of these changes should have associated tests, so we're testing the new `res!=0` code paths in the caller.

netplan_cli/cli/commands/try_command.py
```diff
@@ -135,16 +135,22 @@ def setup(self):  # pragma: nocover (requires user input)
...
+        try:
```
benhoyt: (In other the PR.) This is a good addition, but let's simplify to this pattern:

```python
tempdir = tempfile.mkdtemp()
try:
    ...
finally:
    shutil.rmtree(tempdir, ignore_errors=True)
```

benhoyt (changes requested):

netplan_cli/cli/commands/try_command.py
JegadishD: I have moved the changes related to 'try command' to a new PR - https://github.com/canonical/netplan/pull/611

netplan_cli/cli/commands/try_command.py
JegadishD: The new PR for 'try command' has addressed this issue - https://github.com/canonical/netplan/pull/611

netplan_cli/cli/commands/apply.py
JegadishD: Modified as suggested.

netplan_cli/cli/commands/try_command.py
JegadishD: The new PR for 'try command' has addressed this issue - https://github.com/canonical/netplan/pull/611

netplan_cli/cli/utils.py
JegadishD: I looked at unit-testing the res != 0 branch in `command_generate` and it turned out to be not fitting here. The function's early code has multiple paths/functions , each of which has to be mocked before we can reach systemctl_daemon_reload().

Instead I added an integration test (test_commands_insufficient_privilege_lp2161924 in regressions.py) that runs netplan command as an unprivileged user under a sys.excepthook tracker. It fails, if any command emits an unhandled exception — which is the primary issue in the LP#2161924. This exercises the real `netplan commands` failure path end-to-end.

This also acts as a broader safety net. Any future change that introduces an unhandled exception on the insufficient-privileges path — whether in generate, apply, or try — will be caught.

JegadishD:
@benhoyt Thanks for the review. I have updated the PR and also split a part of the changes to a new PR - https://github.com/canonical/netplan/pull/611
Please review.

tests/integration/regressions.py
```diff
@@ -38,6 +38,53 @@ def test_empty_yaml_lp1795343(self):
...
+except SystemExit:
+    pass
+sys.exit(1 if excepthook_called else 0)
+'''
+
+        env = os.environ.copy()
+        env['PYTHONPATH'] = '/usr/share/netplan'
+        for command in ['generate', 'apply', 'try']:
+            with self.subTest(command=command):
+                test_script = test_script_template.replace('__COMMAND__', command)
```
benhoyt: Seems odd to use a string replace with a magic string `__COMMAND__` here. Python has `str.format` templates. How about using that? You can use `{command}` in the format string, and then `test_script_template.format(command=command)` to insert it.

tests/integration/regressions.py
```diff
@@ -38,6 +38,53 @@ def test_empty_yaml_lp1795343(self):
...
+        self.addCleanup(subprocess.run, ['userdel', '-rf', test_user],
+                        stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
+
+        test_script_template = '''
+import sys
+excepthook_called = False
+def tracking_excepthook(exc_type, exc_value, exc_tb):
+    global excepthook_called
+    excepthook_called = True
+sys.excepthook = tracking_excepthook
```
benhoyt: To be even closer to the "real thing", can we just run it directly without all the excepthook stuff and let the exception bubble up and check that stderr contains "Traceback"?

tests/integration/regressions.py
```diff
@@ -38,6 +38,53 @@ def test_empty_yaml_lp1795343(self):
...
+sys.argv = ['netplan', '__COMMAND__']
+try:
+    from netplan_cli.cli.core import Netplan
+    Netplan().main()
+except SystemExit:
```
benhoyt: With the change above, I wouldn't think we need this try/except either. We can probably just change the subprocess.run to call `python3 -c "from netplan_cli import Netplan; Netplan().main()"` directly.

tests/integration/regressions.py
```diff
@@ -38,6 +38,53 @@ def test_empty_yaml_lp1795343(self):
...
+    pass
+sys.exit(1 if excepthook_called else 0)
+'''
+
+        env = os.environ.copy()
+        env['PYTHONPATH'] = '/usr/share/netplan'
```
benhoyt: Won't this mean it'll call the system netplan rather than the one under test?

benhoyt (commented):
The overall approach looks good to me. And it does seem reasonable to add an integration test here instead of unit tests with tons of patches -- thanks for pushing that.

I have a few comments on the test.

Does this integration test catch the issue we're fixing (before the fix is applied)?

tests/integration/regressions.py
JegadishD: Initially I was coding with str.format but later moved to using str.replace to avoid brace-interpretation issues, like if the template grew to contain literal {...} (e.g. dict literals). 
But I get that the str.format is sufficient and preferred in terms of code-style - I see it used more in the codebase.

I shall make changes to update to str.format.

tests/integration/regressions.py
JegadishD: With the approach to check if a specific-message exists in the traceback, my concern is, if there is any small change in the traceback string formats, it can break the tests. With the approach of excepthook, it can hold good and directly tests the user scenario. I understand that the traceback-string check in stderr sounds simpler. If you think, it is more code for the robustness that we are aiming for, I am open for changing to string-check based approach.

tests/integration/regressions.py
JegadishD: Integration tests are typically run after installing the new netplan code in the device under test - netplan-code under test becomes the system netplan before the integration test.

tests/integration/regressions.py
benhoyt: Yeah, let's not check too many specifics, just contains the word "Traceback" would be fine (and that's been there [since Python 1.0](https://www.bitecode.dev/p/lets-compile-python-10), 30+ years ago). So I don't have any concerns about robustness, and would prefer to avoid messing with excepthook etc.

tests/integration/regressions.py
benhoyt: Ah, that makes sense, thanks.

JegadishD:
@benhoyt Thanks for the review. I have updated the PR based on the review comments. Please review.

benhoyt (approved):
Looks good, thanks!

PR #608, opened 2026-08-21
sriov: support DOCA-OFED driver lifecycle and fix switchdev transition

zarkdav:
## Summary

This PR adds compatibility for the NVIDIA DOCA-OFED / MLNX_OFED driver in Netplan, fixing an issue where Netplan is currently unable to transition Mellanox/NVIDIA NICs to switchdev mode under the OFED driver stack.

### Background & Driver Requirements
According to NVIDIA's [Hardware Offload & Flow Steering Documentation](https://docs.nvidia.com/networking/display/mlnxofedv583070/hardware+offload#HardwareOffload-FlowSteeringMode) and the [Linux Kernel mlx5 Devlink Docs](https://www.kernel.org/doc/html/latest/networking/devlink/mlx5.html):
- DOCA-OFED defaults to Device-Managed Flow Steering (`dmfs`) in standard/legacy modes, whereas switchdev eSwitch offloading requires Software-Managed Flow Steering (`smfs`).
- NVIDIA driver semantics require `flow_steering_mode` to be configured to `smfs` **before** transitioning to `switchdev` mode. Changing steering mode after switchdev is activated or while VFs are instantiated is rejected by the driver (`Operation not supported` / `Device or resource busy`).

### Key Changes
1. **Systemd Service Ordering (`src/gen-sriov.c`)**:
   - Added `After=openibd.service` to `netplan-sriov-apply.service` and `netplan-sriov-rebind.service`.
   - Ensures SR-IOV configuration is applied only after the OFED kernel modules and system initialization are ready.
2. **Automated Flow Steering Configuration (`netplan_cli/cli/sriov.py`)**:
   - Automatically sets `flow_steering_mode` to `smfs` via devlink before switching `mlx5_core` devices to `switchdev`.
3. **Execution Order of Operations (`netplan_cli/cli/sriov.py`)**:
   - Reordered `apply_sriov_config()` so that eSwitch mode and prerequisite flow steering are configured on physical functions before virtual functions are allocated (`set_numvfs_for_pf`).
4. **Dynamic Devlink Binary Lookup**:
   - Uses `shutil.which('devlink')` with fallback to `/sbin/devlink` to support modern merged `/usr` distributions.

### Backward Compatibility
- 100% backward compatible with standard Linux kernel inbox `mlx5_core` drivers and non-mlx5 devices (Intel, Broadcom, etc.).
- Systemd safely ignores non-existent services (`openibd.service`) on installations without OFED.

### Test Plan
- Unit tests added in `tests/test_sriov.py` covering:
  - Prerequisite `flow_steering_mode` configuration.
  - Fallback mechanisms for older driver interfaces.
  - Non-mlx5 driver passthrough.
  - Systemd unit generation assertions.
- Executed and verified:
  - `pytest-3 tests/test_sriov.py::TestSRIOV` -> 41 passed
  - `meson test -C _build codestyle` -> OK (0 style errors)

### DIFF
--- a/netplan_cli/cli/sriov.py
+++ b/netplan_cli/cli/sriov.py
@@ -186,11 +186,8 @@ class PCIDevice(object):
         if self.driver in ['mlx5_core']:
             try:
                 self.devlink_param_set('flow_steering_mode', 'smfs')
-            except Exception:
-                try:
-                    self.devlink_param_set('steering_mode', 'smfs')
-                except Exception as e:
-                    logging.debug(f'Could not set flow steering mode for {self.pci_addr}: {e}')
+            except Exception as e:
+                logging.debug(f'Could not set flow steering mode for {self.pci_addr}: {e}')
 
     def devlink_eswitch_mode(self) -> str:
         """Query eswitch mode via devlink for the PCI device
--- a/netplan_cli/cli/sriov.py
+++ b/netplan_cli/cli/sriov.py
@@ -19,6 +19,7 @@
 import json
 import logging
 import os
+import shutil
 import subprocess
 import typing
 from typing import Dict, List, Optional, Set
@@ -138,9 +139,10 @@ class PCIDevice(object):
         :param value: value to set for property
         :type: str
         """
+        devlink_bin = shutil.which('devlink') or '/sbin/devlink'
         subprocess.check_call(
             [
-                "/sbin/devlink",
+                devlink_bin,
                 "dev",
                 obj_name,
                 "set",
@@ -150,16 +152,57 @@ class PCIDevice(object):
             ]
         )
 
+    def devlink_param_set(self, param: str, value: str, cmode: str = 'runtime'):
+        """Set devlink parameter for the PCI device
+        :param param: devlink param name
+        :type: str
+        :param value: value to set
+        :type: str
+        :param cmode: configuration mode (default: runtime)
+        :type: str
+        """
+        devlink_bin = shutil.which('devlink') or '/sbin/devlink'
+        subprocess.check_call(
+            [
+                devlink_bin,
+                "dev",
+                "param",
+                "set",
+                f"pci/{self.pci_addr}",
+                "name",
+                param,
+                "value",
+                value,
+                "cmode",
+                cmode,
+            ]
+        )
+
+    def ensure_mlx5_switchdev_prerequisites(self):
+        """
+        In DOCA-OFED / MLNX_OFED, transitioning to switchdev mode requires
+        the flow steering mode (smfs) to be configured prior to the transition.
+        """
+        if self.driver in ['mlx5_core']:
+            try:
+                self.devlink_param_set('flow_steering_mode', 'smfs')
+            except Exception:
+                try:
+                    self.devlink_param_set('steering_mode', 'smfs')
+                except Exception as e:
+                    logging.debug(f'Could not set flow steering mode for {self.pci_addr}: {e}')
+
     def devlink_eswitch_mode(self) -> str:
         """Query eswitch mode via devlink for the PCI device
         :return: the eswitch mode or '__undetermined' if it can't be retrieved
         :rtype: str
         """
         pci = f"pci/{self.pci_addr}"
+        devlink_bin = shutil.which('devlink') or '/sbin/devlink'
         try:
             output = subprocess.check_output(
                 [
-                    "/sbin/devlink",
+                    devlink_bin,
                     "-j",
                     "dev",
                     "eswitch",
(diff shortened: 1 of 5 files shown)


trentindav:
I did a quick manual test on a machine equipped with ConnectX-7 NICs and DOCA-OFED installed and I can confirm that using netplan to change devlink mode to switchdev changes `flow_steering_mode`  to `smfs`. 

```
# manually change flow_steering_mode
$ sudo devlink dev param set pci/0000:c1:00.0 name flow_steering_mode value dmfs cmode runtime

# netplan configuration for the CX-7 interface
#    ens16f0np0:
#    ...
#      virtual-function-count: 8
#      embedded-switch-mode: switchdev
#      delay-virtual-functions-rebind: true
$ sudo netplan apply
$ sudo devlink dev eswitch show pci/0000:c1:00.0
pci/0000:c1:00.0: mode switchdev inline-mode none encap-mode basic
$ sudo devlink dev param show pci/0000:c1:00.0 name flow_steering_mode
pci/0000:c1:00.0:
  name flow_steering_mode type driver-specific
    values:
      cmode runtime value smfs
```

netplan_cli/cli/sriov.py
```diff
@@ -150,16 +152,57 @@ def devlink_set(self, obj_name: str, prop: str, value: str):
...
+        """
+        In DOCA-OFED / MLNX_OFED, transitioning to switchdev mode requires
+        the flow steering mode (smfs) to be configured prior to the transition.
+        """
+        if self.driver in ['mlx5_core']:
+            try:
+                self.devlink_param_set('flow_steering_mode', 'smfs')
+            except Exception:
+                try:
+                    self.devlink_param_set('steering_mode', 'smfs')
```
trentindav: the fallback parameter `steering_mode` is not accepted by `mlx5_core` in tree driver nor from DOCA-OFED, and I don't see it documented in https://docs.kernel.org/networking/devlink/mlx5.html

netplan_cli/cli/sriov.py
```diff
@@ -150,16 +152,57 @@ def devlink_set(self, obj_name: str, prop: str, value: str):
...
+                except Exception as e:
+                    logging.debug(f'Could not set flow steering mode for {self.pci_addr}: {e}')
```
trentindav: `debug` level logging may let this Exception remain unnoticed preventing the use of DOCA acceleration

netplan_cli/cli/sriov.py
```diff
@@ -138,9 +139,10 @@ def devlink_set(self, obj_name: str, prop: str, value: str):
         :param value: value to set for property
         :type: str
         """
+        devlink_bin = shutil.which('devlink') or '/sbin/devlink'
```
JegadishD, line 142: devlink_bin assignment occurs multiple times, Please consider to have this as a module level cached util function. It can save multiple calls to shutil and make the maintenance easier.
```
@functools.lru_cache(maxsize=1)
def get_devlink_bin():
  return shutil.which('devlink') or '/sbin/devlink'
```

And at all places, it can be referred as get_devlink_bin()

src/gen-sriov.c
```diff
@@ -45,6 +45,7 @@ write_sriov_rebind_systemd_unit(GHashTable* pfs, const char* generator_dir, gboo
     g_string_append(s, "Description=(Re-)bind SR-IOV Virtual Functions to their driver\n");
     g_string_append_printf(s, "After=network.target\n");
     g_string_append_printf(s, "After=netplan-sriov-apply.service\n");
+    g_string_append_printf(s, "After=openibd.service\n");
```
JegadishD, line 48: `openibd.service` is a specific vendor service, please consider having it as a configurable addition. There can be systems that do not have this service, in such cases, the systemd files referring to it by default does not sound well aligned.

netplan_cli/cli/sriov.py
```diff
@@ -150,16 +152,54 @@ def devlink_set(self, obj_name: str, prop: str, value: str):
...
+        )
+
+    def ensure_mlx5_switchdev_prerequisites(self):
+        """
+        In DOCA-OFED / MLNX_OFED, transitioning to switchdev mode requires
+        the flow steering mode (smfs) to be configured prior to the transition.
+        """
+        if self.driver in ['mlx5_core']:
+            try:
+                self.devlink_param_set('flow_steering_mode', 'smfs')
```
JegadishD, line 188: Overriding the default to `smfs` may not be the user expectation in all scenarios and can potentially break some deployments.

JegadishD (commented):
@zarkdav Thanks for the PR. The context that you had provided in the earlier call was very helpful. It took me some effort to make a reasonable understanding of the context.

Considering the urgency, as agreed, the plan is to review this for applicability to prepare ppa from a private branch.

In this context, I see the changes to be good. There are other comments that I have added, which are to be considered when modifying the fix as main-merge candidate.

PR #611, opened 2026-08-31
cli: try command - handle insufficient user privilege error gracefully

JegadishD:
Insufficient privileges is leading to exception and the following revert as well fails with exception which is not handled and triggers apport crash reports. The fix is to catch on revert and exit cleanly with error messages.

### DIFF
--- a/netplan_cli/cli/commands/try_command.py
+++ b/netplan_cli/cli/commands/try_command.py
@@ -82,10 +82,25 @@ class NetplanTry(utils.NetplanCommand):
         self.parse_args()
         self.run_command()
 
+    def _safe_revert(self, reason: str | None = None) -> int:
+        """Attempt to revert configuration. Returns 0 on successful revert, 1 on revert failure."""
+        if reason:
+            print(f"\n{reason}")
+        print("\nReverting.")
+        try:
+            self.revert()
+            return 0
+        except Exception as e:
+            print(f"\nAn error occurred while reverting: {e}")
+            print("\nPlease check the configuration state")
+            return 1
+
     def command_try(self):  # pragma: nocover (requires user input)
         if not self.is_revertable():
             sys.exit(os.EX_CONFIG)
 
+        # 0 - user action success, 1 - failed
+        exit_code = 0
         try:
             fd = sys.stdin.fileno()
             self.t = terminal.Terminal(fd)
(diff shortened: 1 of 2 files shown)


netplan_cli/cli/commands/try_command.py
```diff
@@ -135,16 +147,20 @@ def setup(self):  # pragma: nocover (requires user input)
...
+            os.makedirs(confdir)
+            self.config_manager.copy_tree('/etc/netplan', confdir, dirs_exist_ok=True)
+            # restore previous state
+            self.config_manager.revert()
+            NetplanApply().command_apply(run_generate=False, sync=True, exit_on_error=False, state_dir=tempdir)
+        finally:
+            try:
+                shutil.rmtree(tempdir)
+            except Exception as e:
+                logging.warning("Failed to remove temporary directory %s: %s", tempdir, e)
```
benhoyt, line 164: Is this essentially the same as using `ignore_errors=True`?
```suggestion
            shutil.rmtree(tempdir, ignore_errors=True)
```

netplan_cli/cli/commands/try_command.py
```diff
@@ -105,19 +119,17 @@ def command_try(self):  # pragma: nocover (requires user input)
             self.touch_ready_stamp()
             self.t.get_confirmation_input(timeout=self.timeout)
         except terminal.InputRejected:
-            print("\nReverting.")
-            self.revert()
+            exit_code = self._safe_revert("Configuration rejected by user.")
```
benhoyt: "Configuration rejected by user" looks like a new error message. I'd say we shouldn't touch the UI or output for existing cases. So I suggest removing the *reason* argument and print nothing here, but keep the printing of "An error occurred..." below.

benhoyt (changes requested):
The overall approach looks reasonable, but a couple of minor comments for consideration.

Also, I'd retitle the PR: the fix here is to handle all errors during revert, not just insufficient user privilege. You should of course mention the specific case in the PR description, though.

netplan_cli/cli/commands/try_command.py
JegadishD: In terms of working, there shall be a difference. With `ignore_errors=True`, even on failure, the recursive removal continues and with `ignore_errors=False`, it raises exception on first failure and returns.
With `ignore_errors=True`. there is no log and no diagnosis information is recorded.

One other approach to consider is. This can provide both.
```
def _on_rmtree_error(func, path, exc):
    logging.warning("rmtree Failed %s %s: %s", func.__name__, path, exc)

shutil.rmtree(tempdir, onexc=_on_rmtree_error)
```

 Please let me know your thoughts.

netplan_cli/cli/commands/try_command.py
JegadishD: I get your concern. I shall remove the new log.

netplan_cli/cli/commands/try_command.py
benhoyt: Ah, I see. Yeah, I think your original approach is reasonable then, with the nested try/except to catch and log a warning.

JegadishD:
@benhoyt Thanks for the review. I have updated the PR based on the review comments. Please review.

benhoyt (approved):

benhoyt:
Thanks, looks good to me. I'm assuming you've done some manual testing locally as well.

PR #613, opened 2026-09-08
tests/integration: fix ovsdb-server test on OVS 4.0.0+

matperin:
OVS 4.0.0 added ovsdb-server.socket for socket activation. Now, stopping only ovsdb-server.service no longer keeps the DB down, since the next ovs-vsctl call triggers systemd to restart the service.

Stop the socket too so the "database is not running" check actually holds.

This change is needed short term to let OVS pass the proposed-migration autopkgtest but it also worth checking if the new service affects the current netplan implementation.

## Checklist

### DIFF
--- a/tests/integration/ovs.py
+++ b/tests/integration/ovs.py
@@ -463,8 +463,17 @@ class _CommonTests():
     def test_ovsdb_server_is_not_running(self):
         self.setup_eth(None, False)
         self.addCleanup(subprocess.call, ['ip', 'link', 'delete', 'br0'], stderr=subprocess.DEVNULL)
+        # ovsdb-server.socket (added in OVS 4.0.0) must be stopped too, otherwise
+        # systemd socket activation will restart ovsdb-server.service on the next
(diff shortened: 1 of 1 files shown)


JegadishD:
@matperin Thanks for the PR.

I am trying to understand the context. Is OVS-4.0 expected to be default version in 26.10 or 27.04?

fnordahl:
> @matperin Thanks for the PR.
> 
> I am trying to understand the context. Is OVS-4.0 expected to be default version in 26.10 or 27.04?

4.0.0 is expected to be default version in 26.10, the netplan.io autopkgtests is currently preventing its promotion from -proposed to -updates.

JegadishD:
> > @matperin Thanks for the PR.
> > I am trying to understand the context. Is OVS-4.0 expected to be default version in 26.10 or 27.04?
> 
> 4.0.0 is expected to be default version in 26.10, the netplan.io autopkgtests is currently preventing its promotion from -proposed to -updates.

@fnordahl Thanks for the info.

JegadishD (approved):

PR #614, opened 2026-09-14
util: explicitly chmod generated files to enforce permissions

vpashaiev:
When `netplan generate` runs under a restrictive umask (such as `0077`), files created by `g_file_set_contents_full()` end up with permissions `0600` instead of `0640`. As a result, `systemd-networkd` (running unprivileged as `systemd-network:systemd-network`) cannot read `/run/systemd/network/10-netplan-*.network`, causing network configuration to fail on boot.

This PR explicitly calls `chmod()` after `chown()` in `_netplan_g_string_free_to_file_with_permissions()` so that the intended file mode is enforced regardless of the caller's umask. It also adds a unit test in `test_netplan_misc` to verify file permissions under umask `0077`.

Addresses feedback from @puida on Launchpad #2164636.
Fixes: https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/2164636


JegadishD (approved):
@vpashaiev Thanks for the PR

